Issue #436, opened 2014-12-12
Updating column type "integer" with a value of 0


ellor1138:
Good day!

I'm creating a workflow to approve creative work. I have an "approvals" table with a "decision" column (integer). When the studio upload an artwork file, a row is created in the "approvals" table and the value of "decision" is set to NULL (no decision yet). When the client go to approve/reject the artwork, the row is updated with his decision (1=approved or 0=rejected) along with the userid and datetime. Everything work great when he approve the artwork (1). The problem happens only when he reject the artwork (0). The "decision" column doesn't change to 0 but stay NULL. All the other info (userid, datetime) are updating successfully.

What I've found!

The update function check if the value passed is different from the value in the table via the hasChanged function. (https://github.com/cfwheels/cfwheels/blob/master/wheels/model/properties.cfm#L319)

To compare the 2 values wheels use the function "$convertToString" to convert each value to a string (https://github.com/cfwheels/cfwheels/blob/master/wheels/global/internal.cfm#L26)

Then wheels use the CF function Compare() to compre the 2 values, wich set the response for the hasChanged function. https://github.com/cfwheels/cfwheels/blob/master/wheels/model/properties.cfm#L369-L375

The problem is when the value passed (to the function "$convertToString") is NULL, the function return 0 (not NULL or [empty string]). Then when wheels compare the real 0 value ("decision") to the false 0 value (column set to NULL) and assume the value has not changed, since they are both equal to 0.

I will submit a pull request with a modification to the "$convertToString" function which resolves the problem.

PR #437, opened 2014-12-12
Patch for $convertToString function bug

ellor1138:
This patch resolve the bug described in the issue : https://github.com/cfwheels/cfwheels/issues/436


ellor1138:
I'm closing this pull request as Per Djurner came up with a better solution for this bug.

PR #551, opened 2015-08-27
Travis parallel

scahyono:
Travis tests now run in parallel, cutting the total test time to half (from 5 minutes to 2,5 minutes).
The longest test is the ACF11/MSQL + ACF10/Oracle, because we cannot run them in parallel.
In the future, if we can upload the wheels package on to separate FTP directories, then these tests can run in parallel too.


perdjurner:
This is great!

PR #621, opened 2016-03-31
changes == to eq for ACF

chapmandu:


chapmandu:
Hmm.. this doesn't look like a failing test, but some sort of maven build error?!

perdjurner:
I restarted the build and it passed.

That server is weird...

Issue #645, opened 2016-07-07
wheels callbacks.cfm $callbacks funtion error


adrox37:
There is an loc is not defined in the function and occasionally causes error.  Adding var loc = {};
fixes the issue.

version: 	CFWheels 1.4.5

Error:
You have attempted to dereference a scalar variable of type class java.lang.String as a structure with members.

 
The error occurred in /Projects/wheels/model/callbacks.cfm: line 148
Called from /Projects/wheels/model/callbacks.cfm: line 172
Called from /Projects/wheels/model/create.cfm: line 66
Called from /Projects/wheels/model/serialize.cfm: line 59
Called from /Projects/wheels/model/serialize.cfm: line 13
Called from /Projects/wheels/model/read.cfm: line 223
Called from /Projects/wheels/model/read.cfm: line 290
Called from <generated>: line 1
Called from /Projects/wheels/model/onmissingmethod.cfm: line 110

chrisdpeters:
@adrox37 Thanks for reporting. Can you submit a pull request with the fix?

adrox37:
@chrisdpeters i don't have permissions to create a pull request.  Otherwise I would have created the request.

chapmandu:
@adrox37 All you need to do is fork the cfwheels repo, make the change in your fork, then submit a pull request.. Happy contributing!

adrox37:
Thanks @chapmandu worked great.

chapmandu:
#646

PR #646, opened 2016-07-08
added scoped variable

adrox37:
fixes error “You have attempted to dereference a scalar variable of
type class java.lang.String as a structure with members.”


chrisdpeters:
Ref #645

PR #661, opened 2016-08-10
Only expand the fullPath if the link is relative

dbelanger:
When I first imported this repo and tried "Run all tests", the controller.miscellaneous.sendfile > "Specifying a directory" failed.

I'm running CF2016 on Ubuntu 14.04.  The error occurred because on line 178, we're trying to expandPath a path that has already been expanded.  The path being expanded was

/var/www/CFWheels/wheels/tests/_assets/files

and the result is

/var/www/var/www/CFWheels/wheels/tests/_assets/files

Notice the double /var/www/ in the local.fullPath after "expandPath".  The expand path is only necessary if the fullPath is relative which is these test cases can be determined by the relative of a parent directory in the path ("..").

I chose this way of dealing with this problem as the previous tests deliver local.fullPath like this

/CFWheels/files/../wheels/tests/_assets/files/cfwheels-logo.png 

Naturally I'm open to other solutions.


perdjurner:
Looks like it's failing on Lucee but works on ACF 10.

dbelanger:
Per, I wrote this for CF2016. It looks like this may be a CF bug.  ExpandPath doesn't realize that the path is already expanded.

perdjurner:
We might be able to do something though, I'll keep this open as a reminder to look closer at it.

perdjurner:
I think a better fix here is to correct the code related to the RemoveChars call.
Can you tell me what your local.folder and local.root variables are before that if statement?

dbelanger:
Here are the two variables you asked for.  Local folder is:

/files

and local.root is:

/var/www/cf/cfwheels/

perdjurner:
Been a long time since this was reported...
Any idea if it has been fixed in a later ColdFusion update?

neokoenig:
see https://github.com/cfwheels/cfwheels/pull/875

neokoenig:
Closing, as I think https://github.com/cfwheels/cfwheels/pull/875 fixes this

PR #671, opened 2016-09-01
Updated the oracle adapter to account for the table's owner

shutterassault:
Previously the same table name could be pulled from multiple schema's in oracle and you wouldn't know which table you are getting.

I added a new application variable called schemaOwner.  I tried to find all the places dataSource was setup and just follow that path.  I didn't make schemaOwner mandatory either.  Figure if it's working for them don't make someone explicitly set it.  But if you do set it under config/settings.cfm it will carry through and filter by it in the oracle.cfc file.

Not sure if I missed anything obvious, or messed up other db adapter, but this update was working for me in Oracle.


perdjurner:
@neokoenig I'm going to leave this in your hands. I don't know enough about Oracle to make a call on this.

neokoenig:
@perdjurner you confuse me for someone who knows *anything* about Oracle. :(
but yes, I shall have a look!

neokoenig:
Closing due to dropping Oracle support in 2.x

PR #673, opened 2016-09-28
Working on the dbmigrate issues for Oracle

shutterassault:
When running the core tests using Oracle as the DB I'm down to 1 error and 244 passed.  I didn't work on this error because @neokoenig  thought the group might want to talk about it and come to a consensus.  Or see what other frameworks have done.

The first thing that failed for me was the populate.cfm file.  I added a couple of semi colons and that was working again.  That eliminated a lot of the previous errors.

The next set of errors was related to dropping sequences and tables that weren't there.  Oracle isn't as easy as some other db's where you can just write IF Exists, then delete it.  I wasn't totally sure how to handle it, but I found http://stackoverflow.com/questions/1799128/oracle-if-table-exists and it seemed to make sense.  I updated the oracle.cfm with this updated way of deleting sequences and tables.

The only issue I had was trying to get the pl/sql to run in cfquery.  I kept getting an end of line error.  All I had to do to fix it was add an additional semi colon  (;;) at the end of the pl/sql.  I'm not 100% sure why, but it's working fine on my end.  I looked around and couldn't find anyone with a similar issue in Coldfusion.  Hoping it'll be valid on Lucee as well.

Regarding the 1 error left, it's the **dbmigrate.addrecord**.  This is a little bit tricky because again Oracle doesn't make it easy with a auto-incrementing datatype.  You have to use a sequence. In the oracle.cfc file I see this note on top

> note that for oracle primary keys we add sequences [tablename_seq] but we leave it up to the model to implement a beforeCreate() callback that sets the id value based on the next value in the sequence

So I wasn't really sure what direction to go.  To me, if we are relying on the user to implement the nextval in the sequence in the model, then should the addrecord test do this work then?  It currently doesn't.  And if we want to stay consistent across databases we probably don't want to have tests specific to Oracle.

The one option is to use a trigger.  I've already seen this used in the populate.cfm file.  You basically would create a sequence, then create the table, and then attach the sequence to the id field of the table.  If we rename the table, then we'd have to rename the sequence, and the table, and then redo the trigger.  But it would probably function more like mysql, etc. and be more consistent.


neokoenig:
Hey mike - any chance you can merge the changes? Oracle.cfc was converted to cfscript between you starting work on it and these commits

PR #681, opened 2016-11-25
Fixes handling of column defaults on insert

andybellenie:


perdjurner:
Thanks, I'll merge this as soon as I've fixed the test suite in the master branch.

perdjurner:
Can you include a test for this before I merge?

I feel like this is something that could easily be broken by later commits. Would be good to have a test to prevent that from happening.

perdjurner:
I think you need to pull in the latest populate.cfm to your repo (we just changed it today).

perdjurner:
Oops, looks like something failed after all.

@andybellenie, can you have a look and see if you know what it could be?

Here is a link to the server it fails on:
http://zebrinny.viviotech.net:8510/index.cfm?controller=wheels&action=wheels&view=tests&type=core

Thanks.

andybellenie:
It looks like some Oracle issues. I reverted my copy of Wheels back to the last commit before I made any changes (814e3c3) and ran the tests on Oracle on my local box and got the same broken tests so it doesn't seem to be anything to do with the new code.

perdjurner:
@neokoenig Any thoughts on this?

neokoenig:
I kicked zebrinny hard up the backside. Tests appear to pass now. I hate that server.

perdjurner:
Great, thanks!

PR #683, opened 2016-11-28
Upgrade plugins to allow multiple override methods

### LINKED ISSUE #682: Upgrade plugins to allow multiple override methods
As discussed at the google group: https://groups.google.com/forum/#!topic/cfwheels/7JxEb5rb5IA

A current limitation in the wheels plugin system is that the system is restrictive in how many times a method can be overridden which currently stands at 1. As I see it, there are 4 things the plugins system is trying to accommodate including:

1. add processing logic on top of a framework method later calling the original method
2. completely override a method
3. use a framework method of the same name by getting it's return value then doing processing
4. add new functionality now currently in wheels

While the current plugin system accomplishes these goals, there are instances where **multiple plugins** should be able to add logic and keep passing the arguments down the chain. One glaring instance of this is `onMissingMethod` where multiple plugins should have the ability to add dynamic methods to models.

liferealized:
Fixes #682


liferealized:
I've merged the conflict with populate.cfm

perdjurner:
Why did you close this?

perdjurner:
Oh, never mind, just saw your post on the Google group :)

liferealized:
My latest commits should fix all outstanding issues. There is one more edge case I want to test out in the morning but it should already be handled with the code provided.

liferealized:
Sorry, I made a new pull request.

https://github.com/cfwheels/cfwheels/pull/684

PR #692, opened 2016-12-05
ColdRoute Integration

### LINKED ISSUE #609: Integrate ColdRoute plugin into core
It is time. :tada: 

Here is the plugin: [dhumphreys/cfwheels-coldroute](https://github.com/dhumphreys/cfwheels-coldroute)

I propose that only these public methods be ported into the core:

-  `mapper` (obviously)
-  `resources`
-  `resource`
-  `get`
-  `post`
-  `patch`/`put`
-  `delete`
-  `scope`
-  `namespace`
-  `root`

The other methods like `controller` I believe encourage bad practice and controller bloat, and we really need to be forcing developers to consider which HTTP verb they are using for each endpoint in their applications. This becomes even more important with #613.

To "link" to deletes, the default implementation that we'd encourage should be to create a form with a submit button that posts the deletion request. We'd then suggest that developers style the submit button to appear and behave like an `<a>` if that's what they want in the UI.

I'm not sure if it's still there, but I've also witnessed some issues with the RegEx patterns used to lookup routes in certain builds of Lucee 5. I believe Travis will pick those issues up, however.

Note: We may also want to consider adding a deprecation for use of the `addRoute` helper, or adding a required `method` argument that forces the developer to mark it `get` vs. `post` (and perhaps use `patch`, `delete` as options ala ColdRoute as well). It may be time to force the developer to use the new routing helpers though, keeping the community unified.

- [x] Integrate plugin
- [x] Change starter `config/routes.cfm` to have common examples
- [x] Write tests
- [x] Write documentation
  - [x] [Routing](http://docs.cfwheels.org/v2.0/docs/routing) chapter in guides
  - [x] Update [Linking Pages](http://docs.cfwheels.org/v2.0/docs/linking-pages) chapter to demonstrate how to link to new routing structure (encourage only doing `DELETE`s via forms/`buttonTo`)
  - [x] Update [Form Helpers and Showing Errors](http://docs.cfwheels.org/v2.0/docs/form-helpers-and-showing-errors) chapter with new routing structure
  - [x] API docs
    - [x] mapper
    - [x] resources
    - [x] resource
    - [x] get
    - [x] post
    - [x] put
    - [x] patch
    - [x] delete
    - [x] root
    - [x] wildcard
    - [x] namespace
    - [x] scope
    - [x] end
- [x] Remove `addRoute` and add to upgrade guide
- [ ] Separate CFWheels UJS plugin for allowing `DELETE`s via links and for grabbing CSRF token for AJAX requests
- [x] Rewrite `wildcard()` mapper to limit methods allowed
- [x] Remove `Path` and `Url` helpers
- [x] Change `module` terminology to `package`
- [x] `package` scoping method
- [x] Rename `drawRoutes()` to `mapper()`

liferealized:
Fixes #609


liferealized:
@chrisdpeters I've merged in the latest master into my changes and it seems some of the csrf stuff is breaking one of the builds

`/home/travis/build/cfwheels/cfwheels/wheels/controller/csrf.cfm: line 143 `

Some type of invalid construct?

chrisdpeters:
I re-ran the failed build in Travis, and it passed. For some reason, the Lucee 4 builds seem to have some sort of inconsistent issue with the new CSRF stuff that I introduced.

Not sure why it would pass sometimes but not, unless there is something getting encoded into that JSON that's invalid?

liferealized:
To me it looks like the decryption failed then the server tried to run `deserializejson()` on the encrypted string. Maybe an if statement above `deserializejson()` with a `isJson()` check on the string?

http://help.adobe.com/en_US/ColdFusion/9.0/CFMLRef/WSc3ff6d0ea77859461172e0811cbec22c24-79e3.html

perdjurner:
Looking good...

@liferealized, can you explain what the following attributes are for?

data-confirm
data-method
data-remote

If they do nothing (like I suspect the "confirm" one at least does) I think we can get rid of them completely.

liferealized:
@perdjurner They are all unobtrusive ways of getting data to javascript. Now that the plugin system can override a wheels core method multiple times, we could easily have a plugin to handle this functionality going forward.

perdjurner:
So, let's get rid of those "data-" attributes then and let plugins add them later as needed.

Makes sense?

liferealized:
Makes complete sense. This will mean that the confirm argument would no longer be present as well.

liferealized:
Done.

chrisdpeters:
Thank you, @liferealized! 🎉

perdjurner:
"The value returned from the init function is not of type Mapper"

Anyone know how that can be fixed?

liferealized:
I've gotten into the habbit of strict typing on lucee and every method in the mapper object returns the object itself. If that is not working on CFxx, we'll need to change the return type from Mapper to struct.

perdjurner:
Still failing, don't have time to look at it closely right now.

liferealized:
What server are you running?

liferealized:
Error message would help as well.

perdjurner:
I'm running against our test server right now (CF 10).

perdjurner:
Just follow this:
https://travis-ci.org/cfwheels/cfwheels/builds/181748842

Or maybe you don't have access?

liferealized:
There are a lot of errors going on with that build. Should the server be restarted to see if that helps?

PR #694, opened 2016-12-06
fix to make sure $invoke is called from $simplelock

liferealized:
$simplelock needs to call $invoke to make sure the plugin system can see methods called


liferealized:
@neokoenig this should be ready now. I've added tests to show that plugin methods work with $simpleLock and $doubleCheckedLock

Issue #695, opened 2016-12-07
Timestamp settings


andybellenie:
Timestamps are currently set as datetime in local time. I'm always having to convert them to UTC. It would be useful to be able to specify a timestamp mode, local, utc, or epoch

chapmandu:
YES!

neokoenig:
Merged + fixed a typo for CF10. Leaving open to document.

chapmandu:
Is this breaking change necessary?

andybellenie:
We have to choose a default and UTC is a more appropriate choice. For a major release it's not a big deal to have to add the setting so long as it's documented in the change notes.

andybellenie:
For reference, use `set("timeStampMode"="local")` to get the current behaviour

chapmandu:
https://github.com/cfwheels/cfwheels/blob/master/CHANGELOG.md#breaking-changes-1

"Timestamping (createdAt, updatedAt) is now in UTC by default (set the global timeStampMode setting to local to reinstate the old behaviour)."

perdjurner:
@chapmandu Is it documented somewhere else too?

perdjurner:
Documented here:
https://guides.cfwheels.org/v2.0/docs/automatic-time-stamps#section-time-zones

PR #696, opened 2016-12-07
Adds timestamp modes

andybellenie:
See isssue #695


perdjurner:
Nice, will merge as soon as the tests for previous commits on master are passing.

andybellenie:
Note that this is a breaking change as I've set the default to UTC. This makes sense to me, I've never understood non-utc timestamps but it's not current behaviour.

perdjurner:
I think that's the default in Rails too, makes sense in that case.

PR #697, opened 2016-12-07
Coldroute update to fix cf10 issue

liferealized:


neokoenig:
hmm, that's throwing this locally for me on CF10

```
Malformed regular expression "^/1(/w+)//?$".

Reason: Invalid backreference: /1.
 
The error occurred in D:/cfwheels/wheels/dispatch/request.cfm: line 91
Called from D:/cfwheels/wheels/dispatch/request.cfm: line 168
Called from D:/cfwheels/wheels/dispatch/request.cfm: line 134
Called from D:/cfwheels/wheels/index.cfm: line 1
Called from D:/cfwheels/index.cfm: line 2
Called from D:/cfwheels/wheels/events/onrequest.cfm: line 8
89 : 
90 : 			// if route matches regular expression, set it for return
91 : 			if (REFindNoCase(local.route.regex, arguments.path)
92 : 					OR (arguments.path == "" && local.route.pattern == "/")) {
93 : 				local.rv = Duplicate(local.route);
```

perdjurner:
I'm not completely against dropping CF 10 support...

Adobe core support ends May next year.

Could always coordinate releasing Wheels 2.0 around that time (with RC's coming out earlier).

Just a thought :)

liferealized:
Nope. We'll get it working.

liferealized:
@neokoenig did you reload your routes after updating? Can you send me the pattern that regex was created from? Should be in `application.wheels.routes`.

liferealized:
Anyone have any recommendations for running all of these versions of CF on single machine?

neokoenig:
Commandbox - essential for this stuff. I'm running CF10/11/2016 + Lucee 4/5

neokoenig:
application.wheels.routes:

image

neokoenig:
Ps, with commandbox, once you've got it installed, navigate to the root from within the shell and do `start cf10` which will launch CF10 (once it's downloaded the War files etc) - you'll need to add the datasource to CFIDE/admin, but otherwise should just work

PR #701, opened 2016-12-07
extra checks to make sure we have everything we need to grab the csrf…

liferealized:
… auth token

### DIFF
--- a/wheels/controller/csrf.cfm
+++ b/wheels/controller/csrf.cfm
@@ -122,31 +122,43 @@ public string function $generateCookieAuthenticityToken() {
 }
 
 public string function $readAuthenticityTokenFromCookie() {
-		local.cookieName = application.wheels.csrfCookieName;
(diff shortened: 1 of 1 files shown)


liferealized:
@chrisdpeters lets see if it fails the build now

liferealized:
Unfortunately I'm still at a loss for why this was happening. My only thinking is that something changed for csrf while the testing server is running like the secret key. Then, when it goes to decrypt the cookie, it fails and decrypt returns the encrypted string instead of throwing an error of some sort.

neokoenig:
Well, it just passed!

chrisdpeters:
My sense has been that it's some sort of race condition unique to testing. Glad that the fix appears to have fixed it. I'll have to pull that back into the plugin and see how it affects our production environments. (You know, if I want to have a _really fun_ week sometime. ;))

liferealized:
Always test in production... isn't that what they say?

PR #759, opened 2017-04-06
Adding support for JSON type in Mysql and also seperating the long types so that length validation isn't an issue

joelstobart:
Add JSON mysql type and also set the rv for text to cf_sql_longvarchar
Add the cf_SQL_LONGVARCHAR and CF_SQL_LONGNVARCHAR to return "text" rather than string

### DIFF
--- a/wheels/model/adapters/MySQL.cfc
+++ b/wheels/model/adapters/MySQL.cfc
@@ -54,7 +54,7 @@ component extends="Base" output=false {
 				local.rv = "cf_sql_varchar";
 				break;
 			case "json": case "text": case "mediumtext": case "longtext": case "tinytext": 
-                loc.rv = "cf_sql_longvarchar";
+                local.rv = "cf_sql_longvarchar";
                 break;
 		}
 		return local.rv;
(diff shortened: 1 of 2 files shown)


wheels/model/adapters/MySQL.cfc
```diff
@@ -50,9 +50,12 @@ component extends="Base" output=false {
 			case "varbinary":
 				local.rv = "cf_sql_varbinary";
 				break;
-			case "varchar": case "text": case "mediumtext": case "longtext": case "tinytext": case "enum": case "set":
+			case "varchar": case "enum": case "set":
 				local.rv = "cf_sql_varchar";
 				break;
+			case "json": case "text": case "mediumtext": case "longtext": case "tinytext": 
+                loc.rv = "cf_sql_longvarchar";
```
chrisdpeters: For CFWheels 2.0, this should be using the `local` scope, not the now-defunct `var`ed `loc` struct.

chrisdpeters:
@joelstobart Thank you for this. Can you fix the one issue that I pointed out?

wheels/model/adapters/MySQL.cfc
joelstobart: Sorry - I'm running a version older than this. So the syntax was differrent...

joelstobart:
FYI:  Test results

Lucee 5.1.0.34

Status: 
 Passed

Duration
00:00:11

Packages
286

Tests
1361

Failures
0

Errors
0

perdjurner:
Thank you.

Note: I moved "tinytext" back to map to "cf_sql_varchar".

PR #772, opened 2017-05-03
Modify JSON content-type search to allow for multiple content types

dbelanger:
First of all I'm not quite sure why the whole file was rewritten when I only changed one line.  I removed two comment lines, 208 and 209, and modified line 210 so that:

`// Allow multiple JSON content types by checking the start and end of the string.`
`// This way we allow both "application/json" and "application/vnd.api+json" (JSON API) for example.`
`if (Left(local.type, 12) == "application/" && Right(local.type, 4) == "json") {`

became

`if (findNoCase("json",local.type)) {`

The premise behind this change is that we've created an API for exclusive use with a mobile app that runs Vue.js .  The "Put" requests from this (and apparently from Laravel as well for our PHP friends) add more information to "Content-Type" than just application/json.  In our case, this is the content-type provided with each "Put" request:

`application/json; charset=UTF-8`

And that's a valid Content-Type.  The charset is especially important for any country that doesn't use UTF-8.  Anyhow, because the original wheels/dispatch/function/$parseJsonBody test used only **right** and **left** functions to test for the presence of JSON content, that presupposes only the 1 option being listed. In my opinion, this test should test for the mere presence of the word "json" in the content-string, and then we'd know to add that object to PARAMS.

### DIFF
--- a/wheels/dispatch/functions.cfm
+++ b/wheels/dispatch/functions.cfm
@@ -205,9 +205,7 @@ public struct function $parseJsonBody(required struct params) {
 		local.type = local.headers["Content-Type"];
 
 		// Only proceed if the content type is JSON.
-		// Allow multiple JSON content types by checking the start and end of the string.
-		// This way we allow both "application/json" and "application/vnd.api+json" (JSON API) for example.
-		if (Left(local.type, 12) == "application/" && Right(local.type, 4) == "json") {
+		if (findNoCase("json",local.type)) {
 
 			// On ACF we need to convert from binary to a string before we can work with it.
 			if (IsBinary(local.content)) {


dbelanger:
Ironically, if you look at [this post](https://www.bennadel.com/blog/2325-coldfusion-supports-http-verbs-put-and-delete-as-well-as-get-and-post.htm) by the master Ben Nadel in 2012, you'll see the Content-Type in the cf output of his examples is exactly what I'm talking about.

neokoenig:
Seems reasonable to me; I'll let @perdjurner comment though as he's the one who wrote the original code

perdjurner:
Can we do a `SpanExcluding` on `local.type` instead (using `;`) maybe?
That way the left / right functionality can be kept as is.
Also, would be great if you can redo the PR so it shows only the changes.

dbelanger:
I can indeed redo the pull request.  
Now, are we sure that
 1) charset will always follow the mime-type and
 2) there will always be a semi-colon even if there is only one option?

That would be my hesitation to using SpanExcluding and left/right

PR #776, opened 2017-05-05
The tagContent argument needs to be passed through to $mail.

perdjurner:
Without this the body is not being included for emails.


perdjurner:
Related commits:
https://github.com/cfwheels/cfwheels/commit/ab83399b70a4ac20352464128e677ea02db3ec32
https://github.com/cfwheels/cfwheels/commit/b0e4e373d4029f634437ee639003a17438b92429

Issue #623, opened 2016-03-31
Answer OPTIONS requests properly


chrisdpeters:
I'll be honest and admit that I wasn't 100% sure what the `OPTIONS` HTTP verb did, so I found [this post](http://zacstewart.com/2012/04/14/http-options-method.html) about the subject.

After ColdRoute is integrated into core (#609), I think it would be relatively trivial for CFWheels to do this automatically.

I see this more as a v2.1 feature, so I'm not going to schedule for the 2.0.0 milestone. I'd rather focus on just getting the existing plugin functionality working for the v2 release. That'll be tricky enough on its own. :)

dbelanger:
Hey Chris, I'm using the 2.0 Beta right now and we will need to make a small change here in order to get the current work-around to be at 100%.  For the moment, we're satisfying CORS by setting the following 3 headers in **.htaccess** or **web.config**:

```
Header set Access-Control-Allow-Origin "*"
Header set Access-Control-Allow-Methods "GET, POST, PATCH, PUT, DELETE, OPTIONS"
Header set Access-Control-Allow-Headers "Origin, Content-Type, X-Auth-Token, X-Requested-By, X-Requested-With"
```

This will work for the time being and a 200 is returned for the OPTIONS request, However, when you inspect the packet returned, you'll see a "Route not found error".

May I suggest we add OPTIONS to the wheels/mapper/matching controller and return a blank DATA field to the browser? (for the Beta release)

(2.1) If you did want to take this all the way, we could also simply return the methods for the routes that have been set.  They should be ready to go in the application scope at the time this is called.  You'd just have to loop through and return the ones allowed by the controller call in options.

chrisdpeters:
I don't know that it's as simple as echoing what's in the `application` scope. Let's say I have a resource with authorization logic, and I don't want certain types of users to know about the existence of certain endpoints. (For example, maybe they can access the `show` endpoint but not the `update` or `delete` endpoints.)

We'd probably need to define some sort of controller logic that would instruct what to expose via `OPTIONS` that runs on each request.

neokoenig:
As a side note, me and @perdjurner were discussing `catchall` methods for the new routing system, and the idea of verb specific catchall routes came up.

The idea was to add a `catchAll()` method to `mapper()` which could then respond appropriately to an incorrect route instead of throwing a "Route not found" error. 

[...]

PR #803, opened 2017-06-20
Workaround for #623: Inject CORS headers, stop request on OPTIONS

dbelanger:
As a temporary workaround for [#623](https://github.com/cfwheels/cfwheels/issues/623), we're setting the Allow-Method headers on every request header to comply with CORS.

Until we add the OPTIONS method into the mapper properly, we need to check for it and abort the request before wheels has a chance to process it.  This will effectively return a 200 and the appropriate headers in order to comply with CORS.

WARNING - This will enable CORS requests for everyone even if the developer doesn't want to allow these requests.

### DIFF
--- a/wheels/events/onapplicationstart.cfm
+++ b/wheels/events/onapplicationstart.cfm
@@ -185,7 +185,10 @@ public void function onApplicationStart() {
 	application.$wheels.csrfCookiePath = "/";
 	application.$wheels.csrfCookiePreserveCase = "";
 	application.$wheels.csrfCookieSecure = "";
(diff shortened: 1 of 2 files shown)


neokoenig:
This should probably use the `$header` wrapper for consistencies sake perhaps?

perdjurner:
Let's also wrap it in a global setting (or two?).

perdjurner:
This looks good, was it closed by mistake?

perdjurner:
Never mind, saw your other PR now.

PR #804, opened 2017-06-21
#623: Add allowCorsRequests & CORS headers, abort processing on OPTIONS

dbelanger:
I deleted the other pull request because once again Git decided to rewrite the whole file on my second commit.

I've added the global flag as requested.

@neokoenig , I've also suggested an update to the 2.0 docs (Configuration and Defaults) so the user is aware of this change.  By default, I've opted to turn this feature off to maintain a secure framework.


perdjurner:
Thanks!

neokoenig:
Docs merged too - ta!

PR #827, opened 2017-11-24
Allow multiple selection for checkbox form element

dyakhnov:
Simple change to allow multiple checkbox selection

### DIFF
--- a/wheels/view/formsobject.cfm
+++ b/wheels/view/formsobject.cfm
@@ -357,7 +357,7 @@ public string function checkBox(
 	arguments.name = $tagName(arguments.objectName, arguments.property);
 	arguments.value = arguments.checkedValue;
 	local.value = $formValue(argumentCollection=arguments);
-	if (local.value == arguments.value || IsNumeric(local.value) && local.value == 1 || !IsNumeric(local.value) && IsBoolean(local.value) && local.value) {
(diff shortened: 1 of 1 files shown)


perdjurner:
What's the real world use case for this?

dyakhnov:
```
for(i in ["red", "green", "blue"]) {
  writeOutput(checkBox(objectName="obj", property="test", checkedValue=i, label=i));
}
```
will render list
```
[ ] red
[ ] green
[ ] blue
```
so if you select "red" and "green", your obj[test] will be "red,green" string, but without change selected values won't stay checked (selected) on submit.

neokoenig:
I kinda see where you're going with this but I'm worried this would actually cause breaking issues with people who may (legitimately) be passing lists in as values.

With `checkboxTag()` there is a workaround:
```
params.checkedValues="red,green";
for(i in ["red", "green", "blue"]) {
  writeOutput(checkBoxTag(name="example", checked=listFindNoCase(params.checkedValues, i), checkedValue=i, label=i));
}
``` 
but I kinda see your point that with `checkBox()` Wheels decides what should be checked depending on what the value of the object is automatically. I wonder if it would be better served with an additional attribute? i.e, a Flag, or a way of passing in a list or array.

perdjurner:
I think the idea for the checkBox() function is that it maps directly to a boolean property in your database (true / false, 1 / 0 etc). Given that I don't think we should change this.

perdjurner:
Closing for the reason mentioned above.

PR #834, opened 2018-01-10
Enable more attributes for stylesheets

michaeldiederich:
Enable the rel attribute for stylesheets.
Update the documentation as sources is the main attribute and source looks deprecated.
As I see, I can use data_theme and it will be modified to data-theme, but I need to overwrite the rel-attribute, too.


perdjurner:
Thanks!

Issue #837, opened 2018-01-24
startFormTag only applies the method attribute if a route is being used


dbelanger:
I just converted all my startFormTag tags to use the method attribute while removing the hidden "_method" tags but then discovered that it only works if a route is specified, not if you use a controller and action.

I'm suggesting a fix and I'll generate a pull request to offer a solution.

neokoenig:
I think (and I'm happy to be corrected) but specifying a route is "the" way in 2.x; the move to resource based RESTful routing, as opposed to just controller / action routing is definitely one of the bigger changes. If, for instance, you removed the `wildcard()` route (which only works for `GET` requests, by design), very little would work.

I think my point is that if anything, we should go the opposite direction - demand the use of a route, and prevent just controller + action based forms/linking; we only really included the wildcard routing  to help migration from 1.x to 2.x - my point being, without a route it *shouldn't* work.

So I'd argue that a "fix" in this direction is actually against the "wheels way". Happy to discuss!

dbelanger:
I agree that 2.0 should always use RESTful routing but we can do that and respect the Wheels way all at the same time.

If I were to use
`startFormTag(controller="users", action="create", method="post")`
 it would be the same as
`startFormTag(route="user", method="post")`
and because it's a POST, our form was already set up to do that so it doesn't change the way we used it in 1.4.  It quite rightly says in the documentation that post is the default method for this tag.

Now, if I use:
`startFormTag(controller="users", action="update", method="patch")`
it should also be the same as
`startFormTag(route="user", method="patch")`
because they're both actual routes that I've set up and they would both still be RESTful.

I would even argue that if we're using these helpers in 2.0 that 
`startFormTag(controller="users", action="update")`
should **assume** that the method is a patch because our
`.resources(name="users")`
assumes the very same thing where it creates both the put and patch routes for us.  But I realize that's a lot more involved.

I'm also aware that using both the controller and the action is more verbose but we've been doing it for quite a while now.  I'm not sure it's the Wheels way but it certainly was the helper way.

The solution I've offered is really only to complete the code in place that looks for a route before deciding whether or not to use the method attribute.

neokoenig:
Thanks for the extra explanation - I think I misinterpreted your original post! So yes, I agree: basically, if there's no route specified, do a controller + method + action lookup and work it out. 

[...]

PR #838, opened 2018-01-24
Convert controller and action to route to enable the method helper

dbelanger:
I'm offering a solution to enable the method helper by looking up the route name that I mentioned in  #837.


dbelanger:
One more thing I should mention is that this didn't pop any flags in the test and I did see places where actions and controllers were used in lieu of routes.  But when you look at the code, you can see that we're only looking for routes to apply the method helper.

PR #845, opened 2018-03-04
Fix #764 Calculated properties should be able to override an existing…

### LINKED ISSUE #764: Calculated properties should be able to override an existing property
It appears that you can't use a calculated property to override an existing column. It appears that wheels ignores the `firstname` property below and just selects the column from the database unmolested.

```
component extends="Model" {

  function config() {
    property(
      name="firstname", 
      sql="IFNULL(firstname, 'These are not the properties you are looking for')"
    );
  }

}
```

andybellenie:
… property

### DIFF
--- a/wheels/model/initialization.cfm
+++ b/wheels/model/initialization.cfm
@@ -59,6 +59,17 @@ public any function $initModelClass(required string name, required string path) 
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


neokoenig:
Nice one, thanks Andy

PR #855, opened 2018-05-03
Adds `flashAppend` as a configurable flash behaviour option

neokoenig:
Add new configuration option: `set(flashAppend=true)` which then allows appending a flash value by default instead of replacing. Set to `false` by default; could make this true in future version maybe.

Stored in an array of strings, `flashMessages()` now checks to see if it's an array of strings or just a string and outputs appropriately.

Also means `flashInsert()` can be handed a simple one dimensional array. 

I don't think this is a major breaking behaviour, as the functionality is hidden behind a flag and can be turned off; the only major difference is that you can now pass an array in as you couldn't before.

So useage if `set(flashAppend=true)`  now can be:

```
flashInsert(success="One");
flashInsert(success="Two");
```
Or 
```
msg=[
   "One", "Two"
];
flashInsert(success=msg);
```
Both of which would give you:
```
<div class=""flash-messages""><p class=""success-message"">One</p><p class=""success-message"">Two</p></div>
```

Note, doesn't currently support directly passing in an array when a key already exists.

### DIFF
--- a/wheels/controller/flash.cfm
+++ b/wheels/controller/flash.cfm
@@ -69,8 +69,26 @@ public any function flashDelete(required string key) {
  */
 public void function flashInsert() {
 	local.flash = $readFlash();
+
 	for (local.key in arguments) {
-		StructInsert(local.flash, local.key, arguments[local.key], true);
+
+		// Does the user want us to append new messages or just replace?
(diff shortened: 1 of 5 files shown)


andybellenie:
👍

Issue #176, opened 2013-06-26
Using the same function names in different test packages causes debug() output duplication


chapmandu:
Edge case!
- create 2 tests in different packages with the same name. eg: name="test_something"
- in those 2 tests, create a variable with the same names. eg: loc.response
- use the debug function in one of the tests to display loc.response

The "loc.response" debug will be displayed in both tests.

Workaround: Ensure test names are unique across all packages

UPDATE:
running this package via URL "package=controllers.AdministratorsTest"

Will run any nested package called "AdministratorsTest" underneath "controllers" folder.

Eg:

controllers.AdministratorsTest
controllers.security.AdministratorsTest
controllers.subdomains.AdministratorsTest

neokoenig:
Is this still an issue in 2.x? I've not hit it myself.

PR #857, opened 2018-05-06
Fixes debugging duplicate output in the test suite

neokoenig:
A smidge hacky, but basically prepends the package name to the debug key (replacing `.` with `_`) to make sure it's unique. See https://github.com/cfwheels/cfwheels/issues/176


chapmandu:
Precisely what I would have done.

Issue #879, opened 2018-07-27
Variable APPLICATION is undefined


dbelanger:
On every request, a call is made to include plugins/standalone/injection.cfm.  That file checks for the presence of things in the application scope.  However, on a session's end, that scope isn't available.

I believe this hasn't been reported because the user isn't around to see the session ending.  However, in an application with a lot of short sessions (such as an API), the errors start to accumulate in the logs.

The following is well describe in the google groups post: https://groups.google.com/forum/#!topic/cfwheels/P4SqL5h9Zkk

The following is the trace that leads to the error:

**coldfusion.runtime.UndefinedVariableException: Variable APPLICATION is undefined.**
 at coldfusion.runtime.CfJspPage._get(CfJspPage.java:351)
 at coldfusion.runtime.CfJspPage._get(CfJspPage.java:330)
 at coldfusion.runtime.CfJspPage._get(CfJspPage.java:317)
 at coldfusion.runtime.CfJspPage._autoscalarize(CfJspPage.java:1798)
 at coldfusion.runtime.CfJspPage._autoscalarize(CfJspPage.java:1760)
 **at cfinjection2ecfm2053640414.runPage(Z:\MYPROJECTDIRECTORY\wheels\plugins\standalone\injection.cfm:5)**
 at coldfusion.runtime.CfJspPage.invoke(CfJspPage.java:253)
 at coldfusion.tagext.lang.IncludeTag.handlePageInvoke(IncludeTag.java:737)
 at coldfusion.tagext.lang.IncludeTag.doStartTag(IncludeTag.java:573)
 at coldfusion.runtime.CfJspPage._emptyTcfTag(CfJspPage.java:3696)
 **at cfModel2ecfc394272461.runPage(Z:\MYPROJECTDIRECTORY\wheels\Model.cfc:4)**
 at coldfusion.runtime.CfJspPage.invoke(CfJspPage.java:253)
 at coldfusion.filter.SilentFilter.invoke(SilentFilter.java:47)
 at coldfusion.runtime.TemplateProxyFactory.resolveComponentHelper(TemplateProxyFactory.java:542)
 at coldfusion.runtime.TemplateProxyFactory.resolveName(TemplateProxyFactory.java:251)
 at coldfusion.runtime.TemplateProxyFactory.resolveComponentHelper(TemplateProxyFactory.java:377)
 at coldfusion.runtime.TemplateProxyFactory.resolveName(TemplateProxyFactory.java:251)
[...]

PR #881, opened 2018-07-30
#879 - Test for the application variable directly in the calling cfm

dbelanger:
It looks like the injection.cfm file is still being called (this time by the wheels\Controller.cfc) and that's causing the same errors (Application not defined).  This still occurs  onSessionEnd, so it's the same error as before.

In this hotfix, I've moved isDefined to the place where the variable is actually being used (which I should have done in the first place).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ server.json
 cfwheels.*.zip
 settings.json
 .vscode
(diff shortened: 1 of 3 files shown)


perdjurner:
Looks good, thanks!

PR #896, opened 2018-10-03
fix to change text field to NVARCHAR(MAX)

reubenbrown13:
This is recommended for all versions of MS SQL that support NVARCHAR(max) especially since TEXT is being deprecated in future versions.  I also found this change to be required to get the Settings table migration to work that was included with the example app.


neokoenig:
Great stuff, thank you

Issue #891, opened 2018-09-18
Using useLayout twice in config doesn't work


dbelanger:
When trying to use "usesLayout" twice in a config, something goes wrong and it defaults to the system default layout.

Example
`usersLayout(template="/templates/layout1", except="show);`
`usersLayout(template="/templates/layout2", only="show);`

neokoenig:
Worth exploring. Also note to self, there's some weird behaviour with `onlyProvides()` which needs exploring which may be related.

PR #897, opened 2018-10-04
Feature/891

dbelanger:
This is the fix to #891.

The _usesLayout()_ function simply created/overwrote a key each time it was called from **anywhere** in the controller.  Confusion arises because on any subsequent call of _usesLayout()_, the "only" or "except" keys would get overwriiten.  Because of that the rendering function _useLayout()_ wouldn't be able to use those very important keys to decide while layout needed to be used.

The solution was to convert that struct into an array of structs and respect the order in which the _usesLayout()_ functions were declared.  

For example:

```
function config(){
	usesLayout(template="/layouts/default1",except="show");
	usesLayout(template="/layouts/default2",only="show");	
}

function index(){
	usesLayout("/layouts/default3");
}	

function show(){
}	
```

If we were to call the show template, it should use the default2 layout.  If we were to call index, it should use the default3 layout because it was declared last.

Becase _config()_ is always processed first and only once while the actions can be called repeatedly, it was necessary to create some logic to compare the structs within the array to the struct to be inserted to avoid duplication.

It was also necessary to break apart the tests to simulate individual requests.

### DIFF
--- a/wheels/controller/layouts.cfm
+++ b/wheels/controller/layouts.cfm
@@ -37,7 +37,7 @@ public void function usesLayout(
 	if (StructKeyExists(arguments, "only")) {
 		arguments.only = $listClean(arguments.only);
 	}
-	variables.$class.layout = arguments;
+	variables.$class.layouts.append(arguments);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


dbelanger:
I should also point out that I ran the old tests as they were before I split them all out.  The old tests passed these changes as well.

wheels/controller/layouts.cfm
```diff
@@ -37,7 +37,33 @@ public void function usesLayout(
 	if (StructKeyExists(arguments, "only")) {
 		arguments.only = $listClean(arguments.only);
 	}
-	variables.$class.layout = arguments;
+
+	// Check to see if the layout struct is already present in the array
+	local.layoutInArray = false;
+	if (arrayLen(variables.$class.layouts)){
+		for (i=1; i<=arrayLen(variables.$class.layouts); i++){
```
neokoenig: Does this need to be scoped?

neokoenig:
Looks good to me - merge it! (and add details to changelog)

wheels/controller/layouts.cfm
dbelanger: To which variable are you referring? i?

dbelanger: I've made those changes

neokoenig:
PS, does this also fix the similar issue with `onlyProvides()`?

dbelanger:
No, this is usesLayout specific but I have a note to recreate and look into onlyProvides.

dbelanger:
Yes, I thought about that.  My main difficulty was that the usesLayout() function has no knowledge of where it came from (which is usually the way we want it).

And in this case, we needed to allow multiple settings in config() and the same logic made it safe to also add this in the function itself.  You can even call it from the view if you like but by that point it would have no effect whatsoever.

Issue #901, opened 2018-10-24
ACF2018 won't output error/dumps without an abort


neokoenig:
See: https://groups.google.com/forum/?fromgroups=#!topic/cfwheels/5nRs6RHS1wg
Possibly related to https://github.com/cfwheels/cfwheels/issues/848

Needs some ACF2018 investigation?

andybellenie:
Yes this is the same issue as 848

zainforbjs:
@neokoenig I have gone through the described scenario on ACF 2018 and when there is an error, it shows the dump normally without an abort and if we just dump anything without any error or abort it works correctly. 
There are no blank pages as written in the issue.

Please visit the following screenshots.
![image](https://github.com/cfwheels/cfwheels/assets/115699497/b5201571-8e96-4f8a-8d7f-f182347b924b)
![image](https://github.com/cfwheels/cfwheels/assets/115699497/e12908ee-7348-4329-ad34-33b72392687e)

zainforbjs:
@neokoenig Closing this issue as the behavior could not be reproduced on `ACF 2018`, and everything appears to be functioning as expected with v3.0, please have your tests and Let us know if you find anything. Thank you as always for your support.

Issue #848, opened 2018-03-23
Cannot use <cfabort> in Lucee or no output generated


andybellenie:
See https://dev.lucee.org/t/cfdump-not-rendering-when-output-false/3589

Newer Lucee versions disable the output buffer by default. So issuing a ``<cfabort>`` will stop any output being generated. This breaks, at a minimum, the missing template and maintenance handlers.

It can be manually re-enabled by setting `this.bufferOutput = true` in Application.cfc (or /config/app.cfm) but I'm not sure if this is the best option - as much as it's a frustrating change it is a makes sense from a performance perspective. I am told newer Adobe CF behaves the same but haven't tested.

neokoenig:
Also worth noting this change was accidentally put in a in a 5.26 version (which has since been replaced by a patch release I believe) - but from 5.3 it will be the default behaviour.

perdjurner:
What do we need to do exactly?

Just stop using abort (and come up with another way to avoid running code that shouldn't run)?

neokoenig:
We *could* add `this.bufferOutput = true` by default for development mode... and then not set anything for the other modes...

andybellenie:
Annoying as it is, I don't see how we can continue to use abort now that this is a recommended production option.

reubenbrown13:
It is not pretty, but here is my fix in the onError file on lines 70 and 86: 
`
if ( StructKeyExists(server, "coldfusion") AND ListFirst(server.coldfusion.productVersion) GTE 2018 ) {
	WriteDump(arguments.exception);
	abort;
} else {
	Throw(object=arguments.exception);
}`

perdjurner:
Possible fix for discussion.

![image](https://user-images.githubusercontent.com/160595/57070197-f58d7380-6cd6-11e9-96d3-25b199b8265a.png)

That fix is for maintenance mode but the general approach would probably apply to other places where `abort` is used as well.

Thoughts?

I can make a PR if you don't see any problems with it.

neokoenig:
Good to see you around @perdjurner ! So basically force an early return? I like the idea - would need to be on a environment setting though presumably (i.e, returnContentOnAbort = set to true for development etc) ? @andybellenie @chapmandu @dbelanger ?

perdjurner:
That's right, exiting the function early so that we don't have to call abort basically.

My proposed code there does exactly the same thing as before, just in a different way to avoid the abort call, so I don't even think an on / off setting for it would be needed.

[...]

PR #902, opened 2018-10-25
Richtext option for TextArea()

reubenbrown13:
I have added an option in the textArea() function to use ckEditor5 for a basic rich text editor.  The CDN to use is set in the setting.cfm file so it can be updated and changed per user's needs.  The script may need to be loaded in another file other than wheels/styles/header if you are using the example-app as your starting point.

### DIFF
--- a/config/settings.cfm
+++ b/config/settings.cfm
@@ -10,4 +10,7 @@
 	// set(dataSourceUserName="");
 	// set(dataSourcePassword="");
 
+	// set CKEditor URL for the CDN to Use
+	set(ckeditorURl="https://cdn.ckeditor.com/ckeditor5/11.1.1/classic/ckeditor.js");
+
 </cfscript>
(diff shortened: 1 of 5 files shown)


reubenbrown13:
Also added a fix for the output issue in the onError file.  Not pretty, but is functional and fixes issues #901 and #848

neokoenig:
Hi Reuben; thanks for your enthusiasm - however, we've got a policy of not including any 3rd party Javascript dependencies into the core: this is perhaps something more suited to a plugin? As a side note, for pull requests it's great if the pull requests can only be related to one issue! I'd encourage you to create a gitHub issue first wherever possible - for instance, we could have saved you some time by saying the CKEditor changes weren't appropriate.

reubenbrown13:
Ahhh... understand... .I had to add a richtext option for a project I am working on anyway.  I will take a look at what it takes to integrate it as a plugin.

reubenbrown13:
I took your advice and ripped out the code and set it up as a plugin.  I need to get it loaded on forgebox, but if you want direct access, here is my repo : https://github.com/reubenbrown13/cfwheels-ckEditor-plugin

reubenbrown13:
this pull needs to be omitted per the guideline for integrating this as a plugin.

Issue #906, opened 2018-11-06
MS SQL adapter doesn't support date, time, timestamp


reubenbrown13:
I noticed that the migration tool treats date, time, and timestamp as DATETIME on ms sql server.  I have updated my fork to address this issue so that they are using the corresponding data type.

neokoenig:
Thanks for the fix :)

PR #907, opened 2018-11-06
fixed date, time, and timestamp mappings

reubenbrown13:
This is to address issue #906


neokoenig:
(Full disclaimer, I know nothing about MSSQL) - is case sensitivity an issue here? Only that all the other bits of code there are noticeably UCASE?

reubenbrown13:
case sensitivity doesn't matter.  I did it in lcase so the changes would be obvious, but if you would like the consistency, I can update the fix.

neokoenig:
Many thanks!

PR #938, opened 2019-04-06
update timestamps on migrator create and update actions fixes #937

### LINKED ISSUE #937: Migrator: addRecord/updateRecord does not set createdAt/updatedAt Fields
I believe one of the core contracts of cfwheels is that the framework automatically manage createdAt, updatedAt, and deletedAt. So, when using the migrator to addRecord/updateRecord, the corresponding timestamp should be managed IMHO. 

> Note: I didn't list removeRecord here because looking at the code, it appears it actually hard deletes the record rather than simply sets the deletedAt field.

geirman:


geirman:
@chapmandu I didn't find a `$timestamp()` function, but did find `$timestampProperty()` in `/wheels/model/miscellaneous.cfm`, but I don't seem to have access to it. 

my first thought was to add it to the imports found in `/wheels/Migrator.cfc` but that didn't work. 

Any suggestions?

chapmandu:
That's odd.. It's used on line 393? It's the function that determines whether to use local or UTC time based on settings..

chapmandu:
It's in wheels/global/internal.cfm. Looks like those functions need to be made available to migrator..

geirman:
Ok, this is strange. One of the problems was that the app I was testing with was behind master. I didn't realize this until you pointed out line 393 and my version of Migration.cfc didn't have a line 393. But, the version I submitted the PR with did. My bad.

I've updated my app with master and now I'm seeing a different problem. The createdAt and updatedAt timestamps are both getting set, even when I'm just using addRecord. I'm trying to figure out what's going on here. 

a bit backwards, but the code on the right generated the results on the left

chapmandu:
Could it be the setUpdatedAtOnCreate setting?

geirman:
Yes, that would be it! I'd say that's unexpected (default) behavior. I think we can close this since this appears to be a non-issue

PR #943, opened 2019-05-02
Adding OPTIONS method

timsayshey:
Supporting options is a must. This is used by JS libraries like Axios to validate the endpoint before hitting it.

### DIFF
--- a/wheels/mapper/matching.cfm
+++ b/wheels/mapper/matching.cfm
@@ -29,6 +29,34 @@ public struct function get(
 }
 
 /**
+ * Create a route that matches a URL requiring an HTTP `OPTIONS` method.
+ *
+ * [section: Configuration]
+ * [category: Routing]
+ *
+ * @name Camel-case name of route to reference when build links and form actions (e.g., `blogPost`).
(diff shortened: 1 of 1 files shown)


neokoenig:
Hey Tim; I've got basic options requests working (in axios / vue) via 
https://github.com/cfwheels/cfwheels/blob/master/wheels/events/onrequeststart.cfm#L119 but it does require `allowCorsRequests()` to be true; additionally, turning on `accessControlAllowMethodsByRoute` will make the `Access-Control-Allow-Methods` header returned the allowed verbs for that route.

See https://github.com/cfwheels/cfwheels/blob/ba13520df7f13707519edca872f9a910177eb86c/wheels/global/internal.cfm#L1158

So my config might look like:
```
set(allowCorsRequests=true);
set(accessControlAllowMethodsByRoute=true);
set(accessControlAllowCredentials = true); 
set(accessControlAllowOrigin="https://app.mydomain.com,https://staging.app.mydomain.com");
```

For me, this provides all the functionality I need - what would adding options to mapper achieve which wouldn't automatically be handled here?

timsayshey:
Thanks @neokoenig! That's what I was looking for

PR #947, opened 2019-05-09
allow definable 'key' name in route definitions. fixes #946

### LINKED ISSUE #946: Feature Request: Definable [key] in Routes
A common pattern in many apps is to use unique `slugs` in place of `keys` or `id` when referencing a data object. An example is...

https://blog.com/posts/my-great-post (where `my-great-post` is the `slug`, replacing the post's id as the key)

There are at least a couple benefits, probably more...

1. They're human readable
2. They don't offer any insight into the datastore

## The Challenge
Currently, CFW hard codes the `[key]` part of the routes, so replacing it with `[slug]` currently requires extra, manual effort. 

## The Solution
With just a few simple changes to the code, we can make this a trivial thing.

1. Add `keyName='key'` to the attributes list of both the `resources()` and `resource()` functions
2. Replace `[key]` of [/wheels/mapper/resources.cfm#L37](https://github.com/cfwheels/cfwheels/blob/master/wheels/mapper/resources.cfm#L73) with `[#arguments.keyName#]`
3. Update documentation

Done! And it's backwards compatible since it defaults to existing behavior. 

With this in place, generating a slug-based (rather than key-based) collection of routes would be trivial. 

```
.resources(name='posts', keyName='slug')
```

## Added Benefits?
Maybe debatable, but I think it would be nice to be able to do the following (if we wanted to) since it's more descriptive and searchable than generically referencing `key` everywhere...

```
.resources(name='posts', keyName='postId')
.resources(name='events', keyName='eventId')
.resources(name='jobs', keyName='jobId')
```

geirman:


chapmandu:
Looks good to me.

neokoenig:
@chapmandu I thought the consensus from https://github.com/cfwheels/cfwheels/issues/946 was not to implement this? In which case we should close the PR

chapmandu:
Hmm.. I do recall that discussion now that you mention it. Closed.

PR #952, opened 2019-05-18
remove all non-numeric characters from param, fixes #900

### LINKED ISSUE #900: ObfuscateParam fails on integer check when input string is a plus-prefixed integer value
ObfuscateParam fails to execute when when the input string is a plus-prefixed integer value.

For example when `arguments.param = "+123"` it passes the integer check. Next it throws an exception "The value 321+ cannot be converted to a number" when trying to execute line `loc.a = (10^loc.iEnd) + Reverse(arguments.param);` as it tries to sum an integer and a string.

geirman:


neokoenig:
Dumb question - is there a chance this breaks/ changes existing IDs? Or is the point that those IDs would never have worked anyway?

geirman:
Not a dumb question at all. I can't think of any reason why there would be a problem with regression for legitimate, numeric IDs in this case because the original function only ever worked if the param was an integer. This just ensures it is by removing any non-numeric characters. 

If someone is passing in "+123" and expecting to get "+123" back though, as they would today, then that will break because they'll instead get the obfuscation of "123" instead.

geirman:
🎉🎉🎉

PR #958, opened 2019-07-12
CFWheels default 2.0.1 installation uses deprecated code application.getApplicationSettings().sessionManagement

djuggler:
Expression Exception - in {root}/wheels/controller/verifies.cfm : line 85

     Element SESSIONMANAGEMENT is undefined in a Java object of type class coldfusion.runtime.StructBean.

CF 10 and greater now uses the getApplicationMetaData() to access application settings


neokoenig:
Never seen this error, but always better for use to be using documented functions / features ! Cheers!

Issue #956, opened 2019-06-13
Plugins route does not accept POST method


chapmandu:
Some plugins contain embedded forms that post to themselves, but the POST method is not accepted by the mapper for the plugin route.

neokoenig:
This has presumably never worked for 2.x?
Or is this specifically to do with the new GUI?

chapmandu:
I'm assuming it's never worked in 2.x.. But I had not tried it until recently. I suppose it depends on whether the old wildcard route accepted posts.

PR #966, opened 2019-10-10
Add configuration option for improve performance of properties #956

chapmandu:


neokoenig:
Seems reasonable to me!

PR #968, opened 2019-10-21
Puts scope dumps in PRE block to preserve line feeds.

bpamiri:
This PR surrounds the scope dumps in the error emails with a PRE tag which will cause the block to preserve their line feeds thus making the email easier to read.


neokoenig:
looks good to me - anyone else?

SebastienFCT:
Looks good to me! And I like the change!

Screen Shot 2019-10-21 at 11 55 54 PM

PR #969, opened 2019-10-23
Uses the X-Forwarded-Proto header if it exists to set the protocol to…

bpamiri:
Modify the isSecure() function in wheels/controller/miscellaneous.cfm to use the cgi.http_x_forwarded_proto header if it exists in determining the protocol to use. Then modified a couple of instances that use cgi scripts directly to use isSecure() to determine what protocol to use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
     - PROFILE=jetty,lucee,subfolder PARAM=testParallelStart=true
     - PROFILE=jetty,lucee,mysql PARAM=none
     - PROFILE=jetty,lucee,postgresql PARAM=none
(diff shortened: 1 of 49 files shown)


chapmandu:
Thanks! Be sure to add a line to the changelog.md

neokoenig:
I've not used this header at all. Happy to leave it up to @chapmandu

bpamiri:
I moved http_x_forwarded_for and http_x_forwarded_proto into $cgiScope and refactored so the two are referenced from the request scope.

bpamiri:
> Thanks! Be sure to add a line to the changelog.md

I've brought this branch up to date and modified the changelog. This should make it easier to merge once approved.

chapmandu:
Merged! Thanks for the PR and the changes.. 👍

PR #977, opened 2020-02-06
Don't require an environment settings file

chapmandu:
There are times when this file is redundant. Eg: in production when loading settings from environment variables. This change just checks for existence of the file, it doesn't break backwards compatibility. I'd like someone to test it on ACF.


andybellenie:
I don't have ACF but looks good to me

neokoenig:
This will also presumably stop that error when you try and do `?reload=idontexist&password=foo` (where it throws an error about not being able to find the appropriate configuration file.)

Issue #979, opened 2020-02-24
turn imageTag() exception on/off


michaeldiederich:
Currently, the function ```imageTag()``` throws an exception if a local image/file doesn't exist or is not an image file.

I would request, that the exception could be turn on/off by a global setting. Or could be turn off in the developer mode. Right now, I want to test something in my development environment and I have one file missing, but I don't care about the missing file, I want to check something else and not update my images folder.

wheels\view\assets.cfm line 195 ff.
cfwheels 2.0.2

neokoenig:
Seems reasonable.

I'd propose we could add a `required` (something like `strict`) attribute on the `imageTag()`, which (to keep current behaviour) could default to` true`, and only if that attribute is `true` would it throw an exception

That way in development you could just set:
```
set(function="imageTag", required = false)
```

Thoughts?

chapmandu:
Closed via f05ed8e99e39067e4d0fe7113504bcdce93f8675

PR #984, opened 2020-03-05
adds required arguments to imagetag #979

chapmandu:


neokoenig:
@dbelanger It should be true, as otherwise it breaks existing behaviour... (even if that behaviour is really annoying)

chapmandu:
If you chaps are happy with this I'll merge it into master

neokoenig:
It's a go from me

wheels/tests/view/assets/imagetag.cfc
```diff
@@ -111,4 +111,25 @@ component extends="wheels.tests.Test" {
 		assert("e eq r");
 	}
 
+	function test_raises_exception_on_missing_image() {
+		path = "missing.jpg";
+		actual = raised("_controller.imageTag(source=path)");
+		expected = "Wheels.ImageFileNotFound";
+		assert("actual == expected");
```
chapmandu:
```suggestion
		assert("actual eq expected");
```

wheels/tests/view/assets/imagetag.cfc
```diff
@@ -111,4 +111,25 @@ component extends="wheels.tests.Test" {
...
+	}
+
+	function test_raises_exception_on_unsupported_image() {
+		path = "../wheels/tests/_assets/files/cfwheels-logo.txt";
+		actual = raised("_controller.imageTag(source=path)");
+		expected = "Wheels.ImageFormatNotSupported";
+		assert("actual == expected");
```
chapmandu:
```suggestion
		assert("actual eq expected");
```

wheels/tests/view/assets/imagetag.cfc
```diff
@@ -111,4 +111,25 @@ component extends="wheels.tests.Test" {
...
+	}
+
+	function test_accepts_missing_image() {
+		path = "missing.jpg";
+		actual = _controller.imageTag(source=path, required=false, alt="This may be broken");
+		expected = '<img alt="This may be broken" src="#imagePath#/#path#">';
+		assert("actual == expected");
```
chapmandu:
```suggestion
		assert("actual eq expected");
```

dbelanger:
Technically this change is functionality correct but I'm still not convinced default to "true" was the way to go.  While we protect existing behaviour, that existed behaviour isn't what's supposed to happen at the browser level.   Required = True is a good thing to have as it will force an error and we're being clear about that but if the standard is to return a silent 404 on the browser when the image file isn't present, I personally believe it should just return the dead link.

neokoenig:
As a compromise, how about we commit the change in 2.2, but mark it as pending change in 2.3? i.e, like a deprecation notice (but not) ?

PR #988, opened 2020-04-08
Renames findLast() to findLastOne()

neokoenig:
Allows wheels to run on lucee@5.3.5+92 upwards


adamzenu:
I think we should also tackle findFirst to bring it in line with findLastOne.

neokoenig:
> I think we should also tackle findFirst to bring it in line with findLastOne.

Tempting, but I'd also rather just duplicate findFirst -> findFirstOne, then deprecate findFirst gradually, and then remove on next proper release? That way we're giving some heads up on a change (rather than what we're doing at the moment..)

neokoenig:
Merging. FYI I manually added this to the 2.1 branch to do a 2.1 final build (as there were other things in the master which weren't going in)

PR #990, opened 2020-04-13
Update minimum Lucee version.

perdjurner:
See this:
https://luceeserver.atlassian.net/browse/LDEV-2052

It creates a problem for us when unzipping plugins, they're not overwritten properly.


neokoenig:
Merging, note to document

PR #997, opened 2020-04-17
Make token invalid by changing start of string instead of end.

perdjurner:


neokoenig:
Not quite sure I understand the reason for this?

perdjurner:
It fixes all tests on ACF 2018 Update 5 and above.

I suspect ACF started ignoring any characters after the end of whatever length a token is in Update 5 (seems logical since being able to pass in longer tokens might have been some kind of security issue).

neokoenig:
Ah I see... gotcha

PR #1011, opened 2020-05-02
Adds Docker Test Suite

neokoenig:
Proposed merge of local docker test suite into master; this won't end up in builds obviously, but not having it in master is a pain :)


chapmandu:
Hell yes!

PR #1013, opened 2020-05-07
Wheels.RouteNotFound: possible XSS attack

michaeldiederich:
In my opinion, there was a XSS possibility for the `Wheels.RouteNotFound` error page. I escape the arguments.path now. ESAPIEncode() is the suggested function from the lucee documentation.


neokoenig:
Whilst I applaud this, you should never been displaying the wheels internal error pages to users, they're for development only and should never be triggered in production. Or am I missing something?

michaeldiederich:
I guess so. If someone would use the modes for any other reason, there could be a potential. Anyway, if you use a XSS scanner like arachni for your local project, it is mostly in developer mode. As this not a big issue, I just escaped the string and created a pull request.

PR #1018, opened 2020-05-13
GitHub actionsCI pipeline

chapmandu:
Here is a working CI pipeline using Github actions and @neokoenig 's Docker test environment.

Some items that would be nice:

- ~~A summary of any failed tests (like the ones you can copy paste from the test-ui)~~
- ~~Move the case statements into functions that can be included in the main bash scripts~~
- Possibly port the bash scripts to commandbox tasks (coded, but requires some docker config for the commandbox task runner to be able to make requests to the docker containers)
- ~~Get sql server tests passing (these failures will be resolved by fixing the config in the sqlserver docker container db setup)~~


dbelanger:
I can't really review this because I don't know enough about Pipelines to be of any use here - I will be using this to learn however!

chapmandu:
> I can't really review this because I don't know enough about Pipelines to be of any use here - I will be using this to learn however!

Heavily infuenced by https://github.com/foundeo/cfml-ci-examples

PR #1023, opened 2020-06-06
Fixes #978 Disable Flash Storage

### LINKED ISSUE #978: Add ability to completely disable Flash
If you're running wheels as a stateless API (i.e with no session management and using your own authentication cookies), you may not want to use flash at all, and deliberately don't want your API to set the `flash` cookie.

Propose to add a `enableFlash` setting defaulting to `true` to allow the developer to completely disable the flash. It won't remove the flash functions from memory, just won't set the cookie, or always return an empty struct etc.

neokoenig:
A simple fix; basically, if flashstorage = "none" don't write the flash to session or cookie; this has the advantage of not massively breaking other functionality. It means wheels apps can now actually pass security audits where `SameSite = strict` is required on all cookies (by handling our own cookies and not using wheels).

### DIFF
--- a/wheels/controller/flash.cfm
+++ b/wheels/controller/flash.cfm
@@ -173,7 +173,7 @@ public any function $writeFlash(struct flash = {}) {
 	} else {
 		if ($getFlashStorage() == "cookie") {
 			cookie.flash = SerializeJSON(arguments.flash);
-		} else {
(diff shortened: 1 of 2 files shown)


chapmandu:
Looks good to me.

Issue #1031, opened 2020-06-19
accessControlAllowOrigin - allow subdomain wildcards


neokoenig:
`application.wheels.accessControlAllowOrigin` can currently be a wildcard or simple value/list/array of permitted values and will match the incoming origin and add the appropriate CORS header.

The CORS spec specifies that you are only allowed either a `*` wildcard, or a specific URL , i.e https://www.foo.com:8080 - it doesn't in itself allow for wildcard subdomains.

However, I think it would be really useful to be able to specify this in our `accessControlAllowOrigin` setting:  e.g

```
set(accessControlAllowOrigin = [
  "https://*.domain.com",
  "https://www.anotherdomain.com"
]);
```
So that you can match `foo.domain.com` and anything else: the function would then set the header to the incoming value if there was a wildcard match to comply with the CORS spec (rather than just do a generic wildcard).

This is useful for staging environments, and also SaaS style subdomains

neokoenig:
Now added - also updated https://guides.cfwheels.org/docs/cors-requests

PR #1035, opened 2020-07-04
Adds $wildcardDomainMatch functions

neokoenig:
See #1031 

Adds 

`$fullDomainString` 
`$fullCgiDomainString`
`$wildcardDomainMatch()`
`$wildcardDomainMatchCGI()`

### DIFF
--- a/wheels/global/internal.cfm
+++ b/wheels/global/internal.cfm
@@ -1265,6 +1265,107 @@ public void function $throwErrorOrShow404Page(
 	}
 }
 
+/**
+ * Wildcard domain match: check if the current cgi.server_name and port satisfies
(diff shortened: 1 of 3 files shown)


neokoenig:
I'm in two minds whether to allow for suffix wildcard matching (i.e https://www.domain.* ):
Part of me is thinking about how that would match phishing requests (I.e https://www.domain.bad.com)

It's in there for now, but I may well remove it. The use case scenario would be domain.net, domain.org etc. but seeing as you're not going to have millions of TLDs, it's probably not too much to ask to make them list them seperately

neokoenig:
Ugh. Looks like ACF doesn't have ArrayReverse(). FFS.

PR #1037, opened 2020-07-14
runs migrations after app onapplicationstart

chapmandu:
It would be nice to have services defined in /events/onapplicationstart.cfm (like error reporting/performance monitoring) available to the migration templates.


neokoenig:
I don't see any reason not to do this except if the Dev has migrator specific code in their onApplicationStart(), then the component won't be initialized yet and thus will break?

chapmandu:
I can't for the life of me think of a use case for that.. I could make it a setting or is that going too far?

neokoenig:
A possible use case might be if they're not using the automatic production migrations, but have rolled their own using the already initi'd migrator in application.wheels scope... Agree it's probably fairly unlikely.

chapmandu:
Maybe I'll rejig it so the component is initialised before but the auto migration happens after the app onapplicationsrart..

chapmandu:
Have a gander at this @neokoenig

neokoenig:
That looks sensible

PR #1040, opened 2020-07-31
restore test application scope on requestend

chapmandu:
Allow a user to do the application scope swapping manually.

Sometimes when an exception occurs in a test case, later test cases error due to application.wheels being restored.

I have implemented this scope swapping manually in my own base Test.cfc component and /events/onrequestend.cfm

### DIFF
--- a/wheels/events/onapplicationstart.cfm
+++ b/wheels/events/onapplicationstart.cfm
@@ -264,6 +264,10 @@ public void function onApplicationStart() {
 	application.$wheels.viewPath = "views";
 	application.$wheels.controllerPath = "controllers";
 
+	// Test framework settings.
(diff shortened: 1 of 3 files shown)


chapmandu:
It's to ensure that the any changes to the application scope that may have been made during test case setup are torn down even if the test runner throws an exception. We run our test suite with a single http request for each package so if one package throws an exception, all later tests fail with an odd "findAll could not be found". Only a reload resolves this.

neokoenig:
Looks fine to me - this will actually solve another issue too, which is where if you run the core wheels test suite, the datasource name gets set to wheelstestdb (and again, requires a reload to set it back to the app's default db).

Issue #1063, opened 2021-04-23
Timestamp settings (UTC) ignored for deletedAt


dbelanger:
This went unnoticed for us for quite some time because honestly, we rarely check the timestamp of a soft deleted record.

However, I recently needed to compare records when I noticed that my **deletedAt** timestamp was before my **createdAt**.  Obviously, that should never be the case.

We added timestamp modes and UTC support was requested as an enhancement in #293 and implemented in #695 / #696 however we only did it for **deletedAt**.

There is only one small change necessary to make this happen however we should warn users that the **deletedAt** will now also default to the timestamp mode selected (just like **createdAt** and **updatedAt** do).  

NOTE: We also use the timestamps in the migrator however the "removeRecord" always hard deletes the records which I believe to be correct only because it's in the migrator

neokoenig:
Pls add to changelog for 2.3 release!

PR #1064, opened 2021-04-23
1063: Make deletedAT UTC compliant

dbelanger:
The will resolve issue #1063 where createdAt and updatedAt have the timestamp mode selector but deletedAt didn't.


dbelanger:
Please note that users should be warned about this change in case any of them have SQL set up to check deletedAt dates (for anything other than IS NULL of course!)

chapmandu (approved):

Issue #1065, opened 2021-04-28
404-handling overlooks... missing files


adamcameron:
There's a slight bug here in the [`$abortInvalidRequest` handler in `internal.cfm`](https://github.com/cfwheels/cfwheels/blob/master/wheels/global/internal.cfm#L437):

```
public void function $abortInvalidRequest() {
	local.applicationPath = Replace(
		GetCurrentTemplatePath(),
		"\",
		"/",
		"all"
	);
	local.callingPath = Replace(GetBaseTemplatePath(), "\", "/", "all");
	if (
		ListLen(local.callingPath, "/") > ListLen(local.applicationPath, "/")
		|| GetFileFromPath(local.callingPath) == "root.cfm"
	) {
		if (StructKeyExists(application, "wheels")) {
			if (StructKeyExists(application.wheels, "showErrorInformation") && !application.wheels.showErrorInformation) {
				$header(statusCode = 404, statustext = "Not Found");
			}
			if (StructKeyExists(application.wheels, "eventPath")) {
				$includeAndOutput(template = "#application.wheels.eventPath#/onmissingtemplate.cfm");
			}
		}
		abort;
	}
}
```

It's possible to fall through the logic and hit that `abort`, in which case one just gets a largely empty response (some bled whitespace mismanagement is all), and a **200-OK** status. This is easy to replicate, just browse to `/any/path/does/not/matter/as/long/as/it/does/not/exist/index.cfm`. Doing this yer - possibly _by definition_ - not going to be in the `application.wheels` context, so you will never set the 404 header.

I can't help but think that in situations where `application.wheels` actually exists, it means you intrinsically must be on a good URL? Otherwise how would the application be identified to have an application scope? Not so sure about that, but perhaps more logic here needs to be checked.

Also it might also be able to fall right through to the bottom of the method entirely, in which case... there's no handling at all. I have not tried to induce this.

I'm pretty sure you want to have this just before the end of the method, outside any of the flow control statements:

```
$header(statusCode = 404, statustext = "Not Found");
abort;
```

As an aside: this code is triggered from `onApplicationStart`. I suspect it ought to be being triggered in `onRequestStart`, given it's related to requests, not the application as a whole..? Different issue though.

adamcameron:
[...]

PR #1066, opened 2021-04-29
Returns 404 header status even in development mode

neokoenig:
#1065 In development (or any) mode, always returns 404 header status on $throwErrorOrShow404Page as all the requests which end up here are 404-able, i.e route not found, view file not found etc, even if we're displaying a friendly developer error


neokoenig:
Hmm, breaks the test suite by returning a 404  even if the tests are successful...more investigation needed

PR #1106, opened 2022-02-14
Fix params bug in CLI API

bpamiri:
This PR fixes a bug in the API used by the CLI to pass commands to the server. The wheels > public > views > cli.cfm file used the params structure directly but that struct doesn't exist. I've modified the calls to `params` to call `request.wheels.params` instead.


chapmandu:
@bpamiri This PR is not showing that any files have changed. What am I missing?

bpamiri:
@chapmandu you weren't missing anything. I on the other hand am missing half a brain. The intended change is now included in the PR. Sorry for the confusion.

chapmandu (changes requested):
Could you leave the indenting intact. I believe there's an .editorconfig file that should help keep standards applied.

chapmandu:
@neokoenig Whilst Peter's solution seems to fix the symptom, why would params scope not be available in this view?

neokoenig:
> @neokoenig Whilst Peter's solution seems to fix the symptom, why would params scope not be available in this view?

IIRC, there's something weird about the way the public cfc is created; basically it doesn't go through the whole wheels controller cycle ( because it can't, otherwise it would interfere with the main app).

Looking at https://github.com/cfwheels/cfwheels/blob/master/wheels/public/includes.cfm#L98 it sure looks like I hit the same issue when doing the rest of the GUI. I'm just trying to work out how wheels/public/views/cli.cfm worked at all, ever. Cos it did at some point.

PR #1111, opened 2022-03-22
syntax-fixes

chapmandu:
fixes some cfformat syntax


bpamiri:
@chapmandu can you let me know a little more about the back story about this PR. I know on one of my previous PRs CFFormat had modified the formatting to the point that you asked me to keep the formatting the same. So, I’m wondering what this PR covers so I can make sure to follow the right procedures going forward.

chapmandu:
The cfformat config settings have evolved to make the syntax readable and consistent. These standards were loosely agreed upon by the wheels core team around the time the code base was converted from tags to script. I implemented these standards into cfformat rules and ran the formatter on the entire codebase. Cfformat does have some limitations, especially around how it wraps long lines contains function calls and operators. 

This PR does 2 things. Fixes some of the quirky formatting, and applies the tweaked rules to the entire codebase.

chapmandu:
Ideally cfformat would run as a pre-commit GitHub action.

bpamiri:
Thanks for the info. I can work to integrate this into the GitHub actions. I have a PR that adds H2 testing to our test suite I can work on adding a step for CFFormat as well. I found sample code like this:

`steps:
- uses: Ortus-Solutions/commandbox-action@v1.0.2
  with:
    cmd: cfformat run models,handlers,interceptors,tests/specs --overwrite`

Do you have a list of folders this should check. At first I thought we should include all our top level folders but perhaps we only need to include the wheels folder if this is running in the make cfwheels repo.

chapmandu:
I'll compile a list of directories.. or better still exclusions. All the cfcs are fine to format, but I generally try to avoid view files that contain html or css. Since there are a lot of cfm includes that are not views, you have to be quite careful.

Issue #1098, opened 2022-02-02
linkTo Includes the Controller, but Disregards the Action


hjozwiak:
Hello,

I have been working my way through the tutorial and I discovered that the linkTo function is disregarding the action that you pass it.
In particular, <p>Time to say <cfoutput>#linkTo(text="goodbye", action="goodbye")#?</cfoutput></p> should have given me a link to http://localhost:60000/say/goodbye, but instead I got http://localhost:60000/say without the action.

chapmandu:
I think you may need to include the controller argument..

`linkTo(text="goodbye", controller="say", action="goodbye")`

I'd suggest looking at routes.. it would end up something like (depending on your routes) `linkTo(text="goodbye", route="sayGoodbye")`

adamcameron:
Hi Other Adam 
I ran into this as well (as did another person who pinged on the cfml slack channel, hence I'm looking into it).

Note the tutorial specifically says this:
```
<p>Time to say <cfoutput>#linkTo(text="goodbye", action="goodbye")#?</cfoutput></p>
```

This is in the [Linking to Other Actions > Linking Hello to Goodbye](https://guides.cfwheels.org/docs/beginner-tutorial-hello-world#linking-hello-to-goodbye) para.

If that needs to say something different, it might be best to update it? Not related to this issue, that said.

Note I tried passing the `controller` param too, to no avail.

I looked at the code, and there seems to be a coupla bugs to me. I'll copy and paste what I said in the cfwheels subchannel here. Hopefully it might help someone sort it out.

---

The logic to build the link in `linkTo` (https://github.com/cfwheels/cfwheels/blob/v2.2.0/wheels/view/links.cfm#L24) ends up calling `uRLFor` (from https://github.com/cfwheels/cfwheels/blob/v2.2.0/wheels/view/links.cfm#L59). At this point it is still passing through everything it needs to create the correct URL.

 
In `uRLfor` (https://github.com/cfwheels/cfwheels/blob/v2.2.0/wheels/global/misc.cfm#L599), the logic flows through to where it checks what route we are on (https://github.com/cfwheels/cfwheels/blob/v2.2.0/wheels/global/misc.cfm#L669). At this point we are on `wildcard` (which is resolved further up I guess), and this is not present in arguments (so the if block doesn't run), but it is present in `local.params.route`, so that `else if` block executes.


What that does is call `$findRoute(route = local.params.route)`. Notice all it passes into that method is the `route`. No other information.


[...]

PR #1119, opened 2022-05-06
issue-1098-urlfor

chapmandu:
Attempts to address https://github.com/cfwheels/cfwheels/issues/1098 but seems to good to be true?!


chapmandu:
I'm a little confused.. it's late on a Friday so I'm not sure if: 
- the tutorial is wrong
- there is still a bug
- the new test case expected value is wrong?

but the new and existing test cases pass with this tweak..

chapmandu:
Seems it's already resolved via https://github.com/cfwheels/cfwheels/commit/d351b474f0f9581754ce2a53cfd8a03a6b9f9a34#diff-53fb5848da39e7b3637d1ba6502f4286c0dc9cefb64baf007c3528dad1f9c956L669

PR #1150, opened 2022-06-21
Fix two broken links in `README`

jbampton:


bpamiri:
Thanks for the fix @jbampton

PR #1183, opened 2022-09-13
Allow datasource argument in finders

chapmandu:
- Implement a `datasource` argument in find* model methods
- Write some tests that uses the H2 datasource as the alternative
- It passes the value down to the query execute function


bpamiri:
@chapmandu I pushed this out last night as v2.5.0-dev and updated the API docs so the 2.5-dev docs are on the site as well. Thanks for the work.

chapmandu:
@bpamiri For some reason, this test case is not in the develop branch?

https://github.com/cfwheels/cfwheels/blob/main/wheels/tests/model/read/datasource.cfc

bpamiri:
I back ported your changes and added this missing test to develop. Thanks for jumping on this so quickly.

PR #1200, opened 2022-10-03
Docker compose updates

chapmandu:
- ~~Fixes typo in docker root volume name~~
- Upgrade to `docker compose`
- Comment out development volume.. it seems to break the github action
- Remove volume `/src/docker` -  it doesnt seem to be used
- Remove 2 redundant workflow steps now that `--wait` option is available to `docker compose`
- Fix BOX_SERVER_CFCONFIGFILE path to prevent commandbox exception

Original commit 5c844fb


bpamiri:
Updated the Vue CLI version because the existing version was causing errors for building the test-ui app on Mac Intel and Mac M1. I also took the opportunity to update the Readme file with the new `docker compose` commands as well as more definition around the datasources created for the test runner. Lastly I added the URL of the test runner to the test-ui console output to make it easier for the user to copy the URL and run the tests manually.

Issue #1182, opened 2022-09-09
Race condition in unlocked application-scope usage


adamcameron:
https://github.com/cfwheels/cfwheels/blob/main/wheels/global/internal.cfm#L758-L759

```
for (local.key in application.wheels.cache[arguments.category]) {
    if (Now() > application.wheels.cache[arguments.category][local.key].expiresAt) {
```
You have a race condition there. The keys that the loop will be iterating over are set once at the beginning of the loop, and given they come from the application scope there is no guarantee all said keys will still exist throughout the entire lifetime of the loop.

Subsequent code in that very function acts to remove items from `application.wheels.cache`, so there's potential for this function to break itself in a single request; but given it's shared-scope code, and given this is a global caching function, it's likely to have lots of requests hitting it simultaneously.

This is not-theoretical: it is giving our application problems.

Access to shared-scopes where there are race conditions like that need to be locked.

I _think_ you can largely mitigate this immediate issue by simply putting an exclusive application scope lock around the function's body. I have only worked through this in my head though, so you would need to test this.

NB: this is not the only place in the codebase where `application.wheels.cache` is accessed directly, so this might not fix it I'm afraid. I did not analyse the other uses, I just did a search in the code and noted the other matches.

Overall, you really ought not be accessing the application scope directly and willy-nilly in your code base. It should be behind a proxy, and the proxy should deal with the locking. Obvs this would be a non-trivial undertaking to implement (and probably some fundamental architecture changes in the app), and I don't think it ought to be in scope for addressing this particular issue.

Lemme know if you need more info.

Demo ApplicationScopeProxy that I threw together when analysing this here: https://blog.adamcameron.me/2022/09/cfml-speaking-of-application-scope.html

Cheers.

chapmandu:
@adamcameron Your blog post was a real eye-opener.. I have always assumed that application scope was "set & forget" (pun intended) and that the language would handle all the locking.. 

I'm going to tinker with a PR using your example, my production app has also experienced this same exception.

I really like the way you have written the test cases using `RunAsync`.. really clever. I'll have to port the test cases to "rocketunit" and deal with older ACF versions.. but.. this is great stuff.

chapmandu:
It appears the fix for THIS specific issue was simpler that implementing an application scope proxy, Though I'm very interested in the concept for future use.

PR #1205, opened 2022-10-11
issue-1182-adds-simplelock-to-sql-caching

chapmandu:
Attempt to resolve https://github.com/cfwheels/cfwheels/issues/1182

- It seems that this is the only unlocked invocation of `$addToCache`


chapmandu:
@adamcameron I have dogfooded this fix in our high-traffic production application and have not seen the exception since..

image

PR #1222, opened 2022-11-13
Findall performance bottleneck

chapmandu:
Addresses major performance bottleneck when using model finders on tables with a lot (hundreds) of columns.

Creates new structs that mimic property lists
Uses StructKeyExists rather than ListFindNoCase when looping through columns after findAll


chapmandu:
I have been dogfooding this for the past couple of weeks in my production apps, on tables with large numbers of columns and large select statements, it runs about 30% faster under load.

PR #1223, opened 2022-11-14
refactor-queryCallback-with-inbuilt-query-functions

chapmandu:
- Speed this up by eliminating list functions
- Use ACF2016+ inbuilt query functions

Performance test results:

(Before) With callback (ListFindNoCase)
afterfind-callback-ListFindNoCase	**15145ms**

(**After**) With callback (QueryKeyExists)
afterfind-callback-QueryKeyExists	**1474ms**

Without callback (afterfind callbacks are evil due to query row to struct creation on each row)
without-callback	**232ms**

Test code
```
// test.cfm
cftimer(label = "afterfind", type = "debug") {
	loop from=1 to=20 index="i" {
		model("User").findAll(maxrows = 100, reload = true);
	}
}

// models/User.cfc
afterFind("do");
...
function foo() {
	arguments.foo = true;
	arguments.bar = true;
	arguments.baz = true;
	arguments.qux = true;
	return arguments;
}
```


bpamiri:
@chapmandu are you still working on this PR?

chapmandu:
I'll have another look today and see where it's at.

chapmandu:
I'm going to do some performance comparisons to see what gains are to be had.

Issue #591, opened 2016-02-17
Non-existent columns in select statements should cause exception


chapmandu:
It seems one can select any column name with impunity..

`user = model("User).findAll(select="id, name, foobar");`

Should this cause an error if the "foobar" column doesn't exist as either a database column or calculated property?

perdjurner:
Yes, that would be good I think.

perdjurner:
Note that this would only apply when not using "table.column" or "column AS something" since in those cases Wheels passes through the select clause unchanged.

Wonder if this can go in 1.4.5?

chapmandu:
Just some more info.. this behaviour is working as desired in WHERE clauses..

`model("User").findAll(where="foo = 'bar'");`

Causes the desired exception:
"Wheels looked for the column mapped to the foobar property but couldn't find it in the database table."

jkeczan:
Just chiming in on this one. I have been out of the cfwheels community a bit lately due to my company transitioning our servers to NodeJS, but I honestly miss this framework too much to not keep up with it. 

Anyways, I think this particular behavior would be good to have a setting for because of the potential to write NoSQL based adapters in the future. May not be in the near future, but it would be good to go if we ever got there

perdjurner:
Closing this as we're using the issue tracker for bugs only from now on.
Please implement your feature suggestion as a plugin or make a pull request.
If you need feedback before starting, please consult the Google Group.

Thanks.

chapmandu:
This is still a legit issue.

zainforbjs:
@chapmandu I reviewed the code in file "wheels/model/sql.cfm" and the function **$createSQLFieldList**. I noticed that the throw function was missing which causes exception in case of an invalid column. Same throw code is used to throw an error in the "where" clause so I added that in "select" and "group" case.

Here is a PR with my changes. https://github.com/cfwheels/cfwheels/pull/1226/commits/e100ddd9bb1ec38b608332c638444c3579581401

chapmandu:
Nice! I think you'll need to use 'raise()' for the test to pass.

zainforbjs:
@chapmandu Thanks for the correction. I will change my test function.

zainforbjs:
@chapmandu Added the test case.

Here is a link. https://github.com/cfwheels/cfwheels/pull/1226

zainforbjs:
@bpamiri This has been already fixed in our recent merges, we can close this.

PR #1226, opened 2022-11-16
Invalid column not throwing exception in select argument

zainforbjs:
Added an exception in case the column specified in the select or group argument passed in the findAll() function does not exist in the database.


bpamiri (approved):
Thanks

bpamiri:
Closes #591

bpamiri (changes requested):
@zainforbjs this is breaking on columns added via a class property as in calculated or aliased columns. Do a search for firstID and you'll see it is defined as a property on the Post model but is getting caught by this exception because the column doesn't exist in the DB.

zainforbjs:
@bpamiri Added some conditions to fix the 6 test cases that were failing. Added details in the commit.

bpamiri (approved):

Issue #1029, opened 2020-06-16
key() should return a numeric type when possible


chapmandu:
foo.key() should return a numeric type if the primary key column is numeric.

It currently returns a string which causes undesired results when returning JSON responses

Eg:
```
{"id": "1", "name": "Harry Potter"}
```
rather than
```
{"id": 1, "name": "Harry Potter"}
```

zainforbjs:
@chapmandu I tried to create the same scenario that you mentioned and one thing I noted is that this issue occurs on CF2018 and not on CF2016. 
CF2016 returns numeric in the JSON string if we use .key() function, but CF2018 returns string.

Here is a PR. https://github.com/cfwheels/cfwheels/pull/1227

PR #1227, opened 2022-11-16
Return a numeric value if the primary key is Numeric

zainforbjs:
This change will return a numeric value if the primary key is Numeric and a String otherwise. This change is due to the issue that returns the primary key as a string in a JSON response


bpamiri:
Closes #1029

Issue #1115, opened 2022-04-28
class attribute overwrite settings in settings.cfm file


blastiko:
Currently if i use the class attribute in a tag constructor (like checkbox, textfield, linkto and so on) this overwrite (if present) settings in settings.cfm file.
For example, if in settings.cfm i have:
set(
	functionName="textField",
	class="form-control",
	labelPlacement="before",
	labelClass="form-label"
);

and in a cfm file i write:
#textField(.... class="text-dark"...)#

the html result is:
<input .... class="text-dark" type="text" value="" ..... >
losing the sets in settings.cfm file

I suggest to introduce new attribute addClass (or similar name) that not overwrite sets in settings.cfm but add class/s present/s in this new attriburte to that is present in settings.cfm file.

zainforbjs:
@blastiko I went according to your suggestion and added a new attribute called "addClass" in fields' functions. This will add the values passed in the "addClass" attribute to the class attribute.

Here is a PR with my changes. https://github.com/cfwheels/cfwheels/pull/1228/commits/2b01829ae8d157c99efd9f04cab18877791bf7ed

bennadel:
Whaaaaaa?!?! I had no idea this existed! Just found it in the source code. This is exactly what I needed.

PR #1228, opened 2022-11-16
Adding addClass attribute in the function textField

zainforbjs:
Added functionality to check if "addClass" attribute is present then add that attribute's value to class so that it does not overwrite the default class value present in config/settings.cfm


bpamiri:
Closes #1115

Issue #942, opened 2019-05-02
paginationLinks() Does Not Work Out-of-the-Box


geirman:
In my controller...
```
function index() {
	param name='params.page' default=1;
	quotes=model("quote").findAll(order="createdAt DESC", page=params.page, perPage=5, handler="quotesQry");
}
```

My view simply calls `#paginationLinks()#` but each page's link looks like this...

```
http://127.0.0.1:60000/quotes/index?page=2
```

> Which throws this error...
> 
> Wheels.QueryParamValue
> 
> The value for cfqueryparam cannot be determined
> Suggested action
> 
> This is usually caused by a syntax error in the WHERE statement, such as forgetting to quote strings for example.
> Error location
> 
> Line 20 in controllers/Quotes.cfc
> 
> 18: 	**/
> 19: 	function show() {
> 20: 		quote=model("quote").findByKey(deobfuscateParam(params.key));
> 21: 	}
> 22: 
> 

Note that `show` route is being triggered. Instead, the link should be formatted like so...

```
127.0.0.1:60000/quotes/?page=2
```

Which works.

geirman:
Going through the code, I noticed an undocumented feature...at least it's undocumented in that `route` isn't part of the API documentation. If I pass the route in like so, then it works.

```
#paginationLinks(route="quotes")#
```

I believe it should also work without that since the documentation specifically states that it shouldn't be necessary...

![2019-05-02_1640](https://user-images.githubusercontent.com/1640318/57113365-0d75ef00-6cf9-11e9-842b-9194b5181c51.png)

neokoenig:
I agree route should be part of the documentation (or we should try and default to it). If it works in the same was as urlFor/linkTo, then specifying the route is more efficient, as it saves an extra loop/lookup. Generally speaking, there's a bit of a disconnect in some of the documentation for 2.x which still refers a fair bit to controller/action - which of course you can still technically do...

geirman:
Got it. Unfortunately controller/action doesn't work in this case for some reason. It still should though. But because the URL is formatted `/controller/action?page=2` it attempts to display the `show()` action even though in this case action was index (e.g. `/quotes/index?page=2`)

zainforbjs:
@geirman I investigated the issue and followed the suggestion provided by @neokoenig to default the route.

Here is a PR with my changes https://github.com/cfwheels/cfwheels/pull/1246/commits/c9f12ea0e0533da711c7d4cb421469c37f577d85

[...]

PR #1246, opened 2023-01-16
set the default route if it is not passed in the function

zainforbjs:
The paginationLinks() function does not set the correct URL on the anchor tag if route is not passed in. Added condition to default the route, if it is not passed in, to the route defined in the request.wheels.params, if the action is index.


bpamiri:
Closes #942

Issue #1048, opened 2020-12-01
Invalid Keys shown in included relationships when using findAll()


bshea5:
Below, you can see request.id and vendor.id are using the value from "id". This occurs when using the select clause and returning as a object/struct. I'm guessing there is a namespace conflict that isn't handled properly by the ORM, since the tables all use an "id" field as there key. 

select: delistedat,id,modelnumber,name,primarylabids,requestid
include: requests,vendor

![image](https://user-images.githubusercontent.com/3984775/100786072-99716080-33df-11eb-8a2e-c3ee11244515.png)

ACF 2016
CFWheels 2.1

dbelanger:
When you specify your own select columns and one of those columns is repeated in both tables, you need to clarify which of those columns you want OR add an alias for each repeateded column.   I'm guess the main ID is some other table (because you have two includes) so if you want them both, I would do this:

`select: delistedat,vendors.id AS vendorid,requests.id AS requestid,modelnumber,name,primarylabids,requestid`

Wheels will only try to correct duplicates if you don't select your own columns (prefacing the name of the repeated column with the table name in the results)

neokoenig:
If you're implicitly selecting the fields,  you'll need to do `requests.id AS requestid` etc..

neokoenig:
lol @dbelanger beat me to it

bshea5:
lol, you guys are fast.

Doesn't doing this invalidate any calculated props though?

neokoenig:
Think they should still work....?

bshea5:
Doesn't seem to for my calc prop, 'primarylabids'. I tried it with a simpler one too, but same problem.
`Message: [Macromedia][SQLServer JDBC Driver][SQLServer]Invalid column name 'primarylabids'.`

dbelanger:
FYI: It's important that AS is uppercase.
Next, you can't nest your aliases.  Ex: requestid,primarylabids should simply be "primarylabids" to appear where it should

bshea5:
@dbelanger , I'm not sure I follow your 2nd suggestion. Using the above example, how would the select param need to change to make this work, or are you saying that nesting an alias in the call will not work?

dbelanger:
My second suggestion is simply not to treat your alias name as a struct (no dot notation).  It should be a simple one word string (ex. primarylabels).

bshea5:
[...]

PR #1249, opened 2023-02-02
Removed the original IF/ELSE condition that invalidates calculated props and added condition

zainforbjs:
The original issue was due to the alias not being passed in to identify the same columns in multiple tables. When we pass in the alias/dot notation in the select clause, it does not add the calculated properties due to the below condition which causes the orignal name of calculated property to be passed in the final query instead of the definition of calculated property, and that gives an invalid column when executed. Commented the below if and else condition and made fixes in case "." and " AS " is passed in.


bpamiri:
closes #1048

PR #1293, opened 2023-10-16
Feature/update luce 6 to latest snapshot for tests

bpamiri:


bpamiri:
Updated Lucee to the latest Snapshot which cause the tests to fail. Had to remove the $ from the spec names to get the test suite to pass. With the $ in the names the function getting the meta data of the components would fail.

Issue #1319, opened 2023-11-08
Problem with using a different datasource in a model cfc in v2.5


edthetechie:
**Describe the bug**
I updated one of my client sites today from 2.4 to 2.5 and it all seems to work except that I'm getting an error on pages where I need to connect to a different database than the default one in settings.cfm 
My model config looks like this:
```
component extends="Model" {
  function config() {
    dataSource("other_database_name");
  }
}
```

In 2.4 it works fine but in 2.5 I get an error saying:
[Macromedia][SQLServer JDBC Driver][SQLServer]Invalid object name 
Which leads me to believe that the correct datasource isn't being accessed

**To Reproduce**
1. In the model's CF add a different datasource to the default one
2. Write a test findAll() function to request rows from a table in the database added in step 1

bpamiri:
Ed,

Can you try `findAll(datasource="other_database_name")` to see if that works. I just did some tests and it appears that the datasource setting in the model is being ignored but if I add it to the findAll() it work. Just wanted to make sure you can confirm this behavior on your end.

-Peter

edthetechie:
Hi Peter,

Yes that works. Will I need to change my code or is this something that can be fixed in a future version? I don't mind waiting to upgrade this client if necessary.

-= Ed

zainforbjs:
@edthetechie Great to hear that works for you. For sure, we can have that fix in future version. Thank you

edthetechie:
No problem at all, I'm currently testing my client sites on the new version so you might see more issues from me but hopefully not. So far everything seems really smooth.

zainforbjs:
@edthetechie Looking forward. Appreciated.

edthetechie:
Thanks for all your hard work, Zain!

chapmandu:
It's possible I broke this when adding the datasource argument to finder functions.

chapmandu:
https://github.com/cfwheels/cfwheels/pull/1324

PR #1324, opened 2023-11-09
issue #1319 model datasource bug

chapmandu:
- Use class datasource as default finder datasource
- Adds test asset model for using alternate datasource
- Unit test case

NOTE: Can this condition be removes so this tests ACF too?
```
} else if (application.wheels.serverName contains "Coldfusion") {
	// seems ACF can't handle H2 datasources
	isTestable = false;
```

### DIFF
--- a/wheels/model/read.cfm
+++ b/wheels/model/read.cfm
@@ -47,7 +47,7 @@ public any function findAll(
 	boolean callbacks = "true",
 	boolean includeSoftDeletes = "false",
 	struct useIndex = {},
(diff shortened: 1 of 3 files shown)


zainforbjs:
@chapmandu can you please confirm that base branch should be develop not main, and create another PR? Thanks

chapmandu:
I'll create another branched from main.

chapmandu:
Close in favour of correctly forked https://github.com/cfwheels/cfwheels/pull/1327

PR #1325, opened 2023-11-10
Fix for datasource() function not working

zainforbjs:
Fixed the issue of not being able to use another datasource than the default one when using datasource() function in the config() function of the Model.


chapmandu:
Too slow.. ;)

https://github.com/cfwheels/cfwheels/pull/1324

bpamiri:
I've merged the fix from @chapmandu so closing this PR. @zainforbjs thanks for jumping on this. We'll get better at coordinating fixes like this.

PR #1327, opened 2023-11-10
issue #1319 model datasource bug

chapmandu:
- Use class datasource as default finder datasource
- Adds test asset model for using alternate datasource
- Unit test case

NOTE: Can this condition be removes so this tests ACF too?
```
} else if (application.wheels.serverName contains "Coldfusion") {
	// seems ACF can't handle H2 datasources
	isTestable = false;
```

### DIFF
--- a/box.json
+++ b/box.json
@@ -1,6 +1,6 @@
 {
     "name":"CFWheels",
-    "version":"2.5.0",
+    "version":"2.5.1",
     "author":"CFWheels Core Team and Community",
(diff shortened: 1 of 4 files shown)


bpamiri:
@chapmandu can you also update the version in box.json to 2.5.1. That way when I merge this into main, it will generate a new release and push the new version up to ForgeBox.

bpamiri (approved):

PR #1350, opened 2023-11-17
Update reading-records.md

MvdO79:
Added documentation for ignoreColumns model config method.

Let me know if the positioning in the guides is right or if another location is desired.

### DIFF
--- a/guides/database-interaction-through-models/reading-records.md
+++ b/guides/database-interaction-through-models/reading-records.md
@@ -174,6 +174,30 @@ model("author").findAll(
     useIndex={
         author="idx_authors_123",
         post="idx_posts_123"
-    }
+		}
 );
+```
+
+### Ignoring Column(s)
+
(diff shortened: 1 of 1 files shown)


guides/database-interaction-through-models/reading-records.md
```diff
@@ -174,6 +174,30 @@ model("author").findAll(
...
         author="idx_authors_123",
         post="idx_posts_123"
-    }
+		}
 );
 ```
+
+### Ignoring Column(s)
+
+For the above finder methods you are able to ignore specified column(s). For instance this can be useful when working with an existing database that has many irrelevant columns. Ignoring column(s) is achieved by adjusting the desired model's config function like below.
```
MvdO79, line 183: This can be confusing, needs to be: "For all select query methods like the above finder methods"?

MvdO79: Or: "For all finder methods"...

bpamiri (approved):

PR #1443, opened 2024-05-29
Sorting the test bundles before execution

zainforbjs:
Sorting the available test bundles before execution so some tests don't fail on linux and MacOS machines.


bpamiri:
Integrated into PR 1447. Thank you for your contribution.

Issue #432, opened 2014-11-26
Cache Model Initialization turned off creates race condition


jduteau:
With cacheModelInitialization set to false, every onRequestStart will clear the model initialization cache (wheels/events/onrequeststart.cfm:84) which does a StructClear on application.wheels.models (wheels/global/internal.cfm:811).  Unfortunately, there are a bunch of accesses to application.wheels.models that assume that the model exists in the cache by the time they are called.  The specific one that is causing errors in our project is in $initModelObject in wheels/model/initialization.cfm (line 237 is the culprit).  $initModelObject is assuming that model("XXX") has been called and that the model with name 'XXX' will be found in application.wheels.models.  It has a bare struct reference: object = application.wheels.models[arguments.name] which fails if, between the time that the model object was accessed and the time that this method executes, the model cache has been cleared.

farism:
Just thought I'd leave my feedback on my real world (development) use case: concurrent asynchronous requests to my REST api. The first request to finish will overwrite the initialization cache for previous requests, and those requests will consistently fail. Not a huge deal in production, but a bit inconvenient while developing and scaffolding out all your models.

perdjurner:
We have removed design mode which had this buggy behavior.

Closing since we will document that `cacheModelInitialization` is prone to race conditions.

andybellenie:
I think we should leave this open, it's still a bug even if the importance his been diminished by design mode removal. It may never be fixed but it's good to keep a record.

perdjurner:
Makes sense.

zainforbjs:
@jduteau The bug with multi requests over model has been fixed, and it's working smooth with limited(30-35) number of requests by performing some tests. I am open to see if you guys take the latest v3.0 and posts your feedback. Thank you for your supports.

PR #1568, opened 2025-05-30
Fix: Model Cache Multi Requests

zainforbjs:
Fixed by adding a check to ensure the model exists in `application.wheels.models` before accessing `$classData()`. 
When `application.$wheels.cacheModelConfig` is set to `false`, the model cache can be cleared, and in rare cases (e.g. race conditions), the model may not be initialized yet. This update safely loads the model if it’s missing and then calls `$classData()`, preventing errors in such scenarios.
We have also removed the use of `simpleLock` since locking is no longer necessary. The logic now ensures models are loaded explicitly when needed, avoiding the previous race condition without requiring locking.


bpamiri:
@zainforbjs I don't have a strong feeling about this but this review is from you know where and the concerns and recommendations are interesting to consider.  

Summary

  This PR addresses a race condition that occurs when application.$wheels.cacheModelConfig is set to false. The issue happens when multiple
  requests try to access a model that hasn't been initialized yet in the application cache.

  Key Changes

  1. Removed $simpleLock usage - The previous implementation used locking to handle concurrent access
  2. Added explicit model existence check - Now checks if the model exists in application.wheels.models before accessing it
  3. Added fallback model loading - If the model doesn't exist, it calls model(arguments.name) to load it, then re-checks

  Analysis

  Pros:
  - Eliminates the need for locking, which can improve performance
  - Provides a clear fallback mechanism when models aren't cached
  - Handles the race condition by explicitly loading missing models

  Potential Concerns:
  1. Double-checking pattern - The code checks for model existence twice, which might be redundant in some cases
  2. Error handling - If model(arguments.name) fails to load the model, the second check might still fail without proper error messaging
  3. Performance impact - Calling model() during initialization could add overhead compared to the locked approach

  Recommendation: The approach is sound but could benefit from:
  - Adding error handling if the model still doesn't exist after calling model()
  - Consider logging when this fallback mechanism is triggered for debugging
  - Testing under high concurrent load to ensure the race condition is truly resolved

zainforbjs:
Here is the Feedback from @bpamiri 

I just looked at it. Take a look at the following:

**Recommendations**
  1. Consider Double-Checked Locking: A hybrid approach that only locks when model needs loading:
 ```
 if (!StructKeyExists(application.wheels.models, arguments.name)) {
      lock name="modelLoad_#arguments.name#" type="exclusive" timeout="10" {
          if (!StructKeyExists(application.wheels.models, arguments.name)) {
              model(arguments.name);
          }
      }
  }
```
  2. Add Logging: Track when fallback loading occurs to monitor frequency in production
  3. Improve Error Messages: Include the actual error from model() in the thrown exception
  4. Test Coverage: Add specific tests for concurrent model initialization scenarios
  5. Documentation: Add comments explaining why locking was removed and the trade-offs

  **Verdict**
  The PR addresses a real issue but introduces potential new race conditions. While it may work in most scenarios,
  high-concurrency environments could still experience issues. Consider implementing double-checked locking pattern
  for a more robust solution.

zainforbjs:
Thanks for the feedback @bpamiri 
We’ve updated the code to use a double-checked locking pattern to handle race conditions more safely. Now, the model is only loaded if it’s not already in `application.wheels.models`, and a lock ensures it’s done safely under high concurrency.

We also added better error messages that include the original model() error for easier debugging. To validate this, we added a stress test that simulates concurrent access. It randomly clears the model cache and ensures no thread fails.

The change is documented in the changelog and included in [PR #1568](https://github.com/wheels-dev/wheels/pull/1568), resolving [issue #432](https://github.com/wheels-dev/wheels/issues/432).
Let me know if you see anything else we should adjust!

PR #1856, opened 2026-02-03
feat(sql): escapes reserved words

chapmandu:
Ensures MySQL reserved words (like groups, order, key, etc.) are properly escaped with backticks when used as table names, preventing SQL syntax errors.

Changes:

$escapeReservedWords() in MySQL adapter checks table names against reserved word list
$fromClause() calls the adapter's escape function when building SQL
Test:

Added test: "escapes MySQL reserved word 'groups' in from clause"

This feature does not define reserved words for all db adapters, only MySQL


chapmandu:
Nice @bpamiri 

Closed in favour of https://github.com/wheels-dev/wheels/pull/1874

PR #1866, opened 2026-02-09
Fix typo in README.md from 'functional' to 'functions'

michaeldiederich:
Notice for Filename for getting Tests running


bpamiri:
Thanks for catching this inconsistency @michaeldiederich! After looking into it, I agree with @zainforbjs — `functional/` is the correct name. No major framework uses `functions/` as a test directory name (`functional/` aligns with Django and legacy Rails conventions, while Laravel uses `feature/`, modern Rails uses `system/`, and NestJS uses `e2e/`). The template folder was the typo, not the README.

I've pushed the rename in #1872 which should close this issue.

PR #1896, opened 2026-02-25
Add CockroachDB RETURNING-based identity select and CI test matrix

zainforbjs:
## Summary

Completes CockroachDB adapter support with proper identity select and full CI coverage.

- **RETURNING clause**: CockroachDB doesn't support `pg_get_serial_sequence()`/`currval()`, so INSERTs now use `RETURNING pk` to retrieve generated keys — no race condition
- **Version detection fallback**: CockroachDB uses the PostgreSQL JDBC driver, so `SELECT version()` disambiguates when `driver_name` reports "PostgreSQL"
- **CI test matrix**: CockroachDB added to Docker Compose, all CFConfig datasources, and the test workflow with proper readiness checks
- **Adapter name lists**: CockroachDB added to sql.cfc and update.cfc adapter checks for UPDATE with include

### Changes from original PR
- CockroachDBModel now extends `PostgreSQLModel` (not `Base`) — inherits query helpers, only overrides what differs
- Boolean maps to `cf_sql_boolean` (not `cf_sql_bit`) — CockroachDB enforces strict type checking
- CockroachDBMigrator kept as-is from develop (extends `PostgreSQLMigrator` with `unique_rowid()`)
- Removed unrelated `server.json` change
- Removed `currval()` fallback (unsupported by CockroachDB)
- Replaced `sleep 10` in start.sh with proper readiness loop
- Adapted to new engine-grouped CI workflow structure

Closes #1971, closes #1973

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1989, opened 2026-03-16
Override $identitySelect for CockroachDB adapter

### LINKED ISSUE #1971: CockroachDB: $identitySelect may fail with unique_rowid()
## Summary

CockroachDBModel inherits PostgreSQL's `$identitySelect()` which calls `pg_get_serial_sequence()` and `currval()`. These may not work correctly with CockroachDB's `unique_rowid()` default values.

## Details

- PostgreSQL uses sequences behind `SERIAL` — `currval()` retrieves last inserted value
- CockroachDB's `unique_rowid()` doesn't use sequences, so `pg_get_serial_sequence()` returns NULL
- Need CockroachDB-specific identity retrieval, likely using `RETURNING` clause
- CockroachDB supports `INSERT INTO ... RETURNING id` natively

## Files

- `vendor/wheels/databaseAdapters/CockroachDB/CockroachDBModel.cfc` — needs `$identitySelect()` override
- `vendor/wheels/databaseAdapters/PostgreSQL/PostgreSQLModel.cfc:124-171` — inherited method

bpamiri:
## Summary
- CockroachDB inherits PostgreSQL's `$identitySelect()` which uses `pg_get_serial_sequence()` and `currval()` — neither of which CockroachDB supports
- Overrides the method in `CockroachDBModel.cfc` with a MAX-based query, since CockroachDB's `unique_rowid()` generates monotonically increasing values
- Follows the same pattern used by the Oracle adapter, with BoxLang compatibility from the PostgreSQL adapter

Closes #1971

## Test plan
- [ ] Verify `model("Foo").create(...)` returns the correct primary key on CockroachDB
- [ ] Verify explicit PK inserts still work (no identity select triggered)
- [ ] Confirm no regression on PostgreSQL (parent adapter unchanged)

### DIFF
--- a/vendor/wheels/databaseAdapters/CockroachDB/CockroachDBModel.cfc
+++ b/vendor/wheels/databaseAdapters/CockroachDB/CockroachDBModel.cfc
@@ -1,6 +1,57 @@
 component extends="wheels.databaseAdapters.PostgreSQL.PostgreSQLModel" output=false {
 
 	/**
+	 * Override PostgreSQL's $identitySelect which uses pg_get_serial_sequence() and currval().
+	 * CockroachDB doesn't support these functions. Since CockroachDB's unique_rowid() generates
+	 * monotonically increasing values, we use a MAX-based query (similar to the Oracle adapter).
+	 */
+	public any function $identitySelect(
+		required struct queryAttributes,
+		required struct result,
+		required string primaryKey
+	) {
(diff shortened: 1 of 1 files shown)


bpamiri:
Closing in favor of #1896's RETURNING-based approach, which avoids the race condition inherent in SELECT MAX(pk).

PR #2378, opened 2026-04-30
fix(config): config-include output safety net + visible package failures + tutorial doc fixes from VM bake

bpamiri:
## Summary

A bundle of related fixes surfaced by a fresh-VM tutorial bake (the same one that produced the journal+report you've been reviewing). Each is small on its own; together they unblock the chapter 6b auth flow and the bonus basecoat chapter, and they make two whole classes of silent failures loud the next time someone hits them.

### Framework changes
- **`$includeConfig` safety net** on the four `onApplicationStart` config-file loads — captures cfinclude output, warns to wheels.log if non-empty (almost always means the user's `config/*.cfm` is missing its `<cfscript>` wrapper). Stops the visible "page leak" symptom that surprised the bake.
- **Surface PackageLoader failures at boot** — fix `file="application"` writes that don't materialize on the default Lucee log layout; add an aggregate WARN to wheels.log + per-package ERROR to wheels-errors.log so a developer chasing a "No matching function [...]" error has a concrete breadcrumb.

### Doc changes
- **Chapter 6b** — wrap the `services.cfm` example in `<cfscript>...</cfscript>`, rewrite the troubleshooting block to point at the missing-wrapper case first.
- **Chapter 4** — add `allowBlank=true` to the `validatesLengthOf` example and explain the framework's blank-handling default. Without this, blank submissions fire BOTH "can't be empty" AND "is the wrong length" even though zero characters is well under the 120-char cap. Whether the framework default should change is a separate v4 conversation (compat risk).
- **Bonus chapter** — switch the basecoat CSS curl URL from `basecoatui.com` (serves HTML) to jsdelivr (real CSS); replace the misleading "Lucee 7 rejects `mixin=` attribute" troubleshooting note with the actual blocker pointer (`$UIBUILDID` / wheels-basecoat 1.0.3).

## Verification

All changes verified end-to-end on a real VM (macOS 26.2 / Lucee 7.0.0.395 / Wheels 4.0.0-SNAPSHOT+1652):

- ✅ Chapter 6b login flow: `POST /login` → `302 /posts` (was: 500 "service not registered")
- ✅ services.cfm without wrapper: visible page leak GONE, warning surfaces in wheels.log with first-200-chars preview
- ✅ Induced package failure: aggregate WARN in wheels.log + per-package ERROR in wheels-errors.log
- ✅ Chapter 4 with `allowBlank=true`: empty title fires ONE error ("can't be empty"), 200-char title fires ONE error ("wrong length"), valid title fires zero
- ✅ Bonus chapter jsdelivr URL: 43.5 KB of real CSS instead of HTML

## Related work

Three follow-ups in their own PRs (not in this one):

- **LuCLI** — the bonus chapter's "save CSS to `public/assets/`" instruction is dead on arrival because the shipped rewrite template uses negation-style `RewriteCond` chains that Tomcat 11's `RewriteValve` doesn't honor. Fixed in [cybersonic/LuCLI#60](https://github.com/cybersonic/LuCLI/pull/60).
- **wheels-basecoat 1.0.3** — `$uiBuildId`/`$uiLucideIcon` private→public so `PackageLoader` carries them across the mixin boundary. Fixed in [wheels-dev/wheels-basecoat#3](https://github.com/wheels-dev/wheels-basecoat/pull/3).
- **Chapter 7 SignupFlowSpec** — diagnosed but deferred. The documented browser spec hits a real Playwright/Turbo Drive interaction (POST fires, user IS created, but URL doesn't navigate via Turbo's History.pushState path). Needs a chapter-7 doc-design pass.

## Findings I did NOT change

Two items from the report turned out to be misdiagnoses (verified on VM):

- **Finding #11** (basecoat `mixin="controller"` rejected by Lucee 7): not actually a failure on Lucee 7.0.0.395. The attribute compiles and PackageLoader loads the package successfully. The real blocker was always Finding #14 (private helpers).
- **Finding #13** (`linkTo` / `buttonTo` rejects `class=`): three different variants verified on VM — `linkTo(text=, route=, key=, class=)`, `linkTo(text=, controller=, action=, key=, class=)`, `buttonTo(text=, route=, key=, method=delete, class=)` — all render the `class` attribute on the resulting tag. No framework change needed.

## Test plan

- [ ] CI passes on the full matrix (Lucee 5/6/7, Adobe 2018/2021/2023/2025, BoxLang)
- [ ] PackageLoader spec — induce a failure, assert wheels.log + wheels-errors.log content
- [ ] $includeConfig spec — induce missing-wrapper, assert no leak + warning fires
- [ ] Verify-docs harness still passes for the modified chapters

### DIFF
--- a/cli/lucli/templates/app/rewrite.config
+++ b/cli/lucli/templates/app/rewrite.config
@@ -1,0 +1,54 @@
+# Wheels URL Rewrite Rules (Tomcat RewriteValve)
+# All requests that don't match a static resource or a CFML file are
+# forwarded to the front controller (index.cfm) with the original path
+# preserved as PATH_INFO.
+#
+# Example: /posts/1/edit -> /index.cfm/posts/1/edit
+#
+# Documentation: https://tomcat.apache.org/tomcat-9.0-doc/rewrite.html
+#
+# Implementation note: Wheels apps ship this file at the project root so
+# LuCLI's CatalinaBaseConfigGenerator picks it up verbatim (the override
+# path documented at LuCLI/src/main/java/.../CatalinaBaseConfigGenerator.java
+# — when projectDir/rewrite.config exists, LuCLI uses it instead of its
+# bundled default template). This decouples Wheels from LuCLI's release
+# cycle and lets Wheels own its routing semantics.
+#
+# These rules use positive-match passthrough rules with the [L] flag rather
+# than the more idiomatic-looking RewriteCond !-negation chain. Tomcat 11's
+# RewriteValve does NOT honor stacked `RewriteCond %{REQUEST_URI} !pattern`
+# entries before a single rewrite RewriteRule the way Apache mod_rewrite
+# does — the conditions effectively don't gate the rule, the rule fires for
+# every request, and public/ static files (.css, .js, .txt, .png, etc.)
+# silently 404 even when their extensions are explicitly listed in the
+# negation. Each "skip" condition is therefore its own RewriteRule with
+# [L], which exits rule processing and lets Tomcat's default servlet serve
+# the file.
+
+# Skip rewriting if the request is already accessing the front controller
+# (covers both `/index.cfm` itself and `/index.cfm/path-info`). Without
+# this rule, framework-emitted URLs like `/index.cfm/posts/1` would get
+# rewritten to `/index.cfm/index.cfm/posts/1` when URL rewriting is
+# partially configured.
+RewriteRule ^/index.cfm.*$ - [L]
+
+# Serve direct CFML file requests unchanged
+RewriteRule ^/(.*)\.(cfm|cfc|cfml)$ - [L]
+
+# Pass static-extension files through to the default servlet
+RewriteRule ^/.+\.(css|js|jpg|jpeg|png|gif|ico|svg|webp|avif|woff|woff2|ttf|eot|otf|pdf|zip|json|xml|txt|map|mp3|mp4|webm|wasm)$ - [L]
+
+# Pass requests under conventional Wheels static directories through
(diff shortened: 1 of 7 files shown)


bpamiri:
## Update: added Wheels-side `rewrite.config` template (commit 1ba7bfd)

Pushed `cli/lucli/templates/app/rewrite.config` so `wheels new` ships its own routing config rather than depending on LuCLI's bundled default. Background: the same Tomcat-11-RewriteValve bug being fixed upstream in [cybersonic/LuCLI#60](https://github.com/cybersonic/LuCLI/pull/60) was making the bonus-chapter's `curl -o public/assets/...` instructions dead on arrival. Two reasons to also ship a Wheels-side fix:

1. **Decouples Wheels from LuCLI's release cycle.** `wheels new` apps work correctly on first boot regardless of when LuCLI ships its fix.
2. **Wheels has its own conventions** (`stylesheets/`, `javascripts/`, `files/`, `miscellaneous/` static dirs; modern web extensions like `webp`/`avif`/`wasm`; root-file passthroughs for `robots.txt`/`favicon.ico`/etc.) that don't belong in a generic LuCLI template.

LuCLI's `CatalinaBaseConfigGenerator.java` already supports the override path: when `<projectRoot>/rewrite.config` exists, LuCLI uses it verbatim instead of the bundled default.

Verified against a 12-case URL matrix on Tomcat 11.0.13 + Lucee 7.0.0.395 — see commit message for the full table. All 12 pass with correct status + Content-Type.

Existing Wheels apps can adopt the fix by copying the template file into their project root; new apps get it for free. May be worth a `wheels upgrade`-style migration in a follow-up so existing apps pick it up automatically.

PR #2543, opened 2026-05-10
fix: ship registry reader with framework so Tools→Packages works in fresh apps

### LINKED ISSUE #2530: Debug Panel "Packages" tab does not list registry packages (pipeline test)
> **Pipeline test note:** This is a duplicate of [#2526](https://github.com/wheels-dev/wheels/issues/2526) opened to exercise the wheels-bot pipeline end-to-end after PR [#2529](https://github.com/wheels-dev/wheels/pull/2529) lifted the Phase 4 gates. The original #2526 has stale `wheels-bot:triage:2526` markers from earlier triage runs that block re-triggering. The bot should triage and propose-fix this fresh issue automatically — if a draft PR results, please link it to both issues at merge time.

---

**Describe the bug**

The Debug Panel's "Packages" tab is supposed to give developers visibility into both (a) the packages currently installed in `vendor/` and (b) the packages available from the wheels-packages registry, with links to each package's index page.

The first half works — installed packages render correctly. The registry-packages section, however, never renders any packages at all. We currently have 6 first-party packages registered (`wheels-sentry`, `wheels-hotwire`, `wheels-basecoat`, `wheels-legacy-adapter`, `wheels-i18n`, `wheels-seo-suite`), so this list should not be empty.

**To Reproduce**

1. Run any Wheels app in development mode (`wheels start`).
2. Open the Debug Panel at the bottom of any page.
3. Scroll to the "Packages" section.
4. Observe: only the installed packages are listed. There is no list of packages registered with `wheels-dev/wheels-packages`.

**Expected behavior**

The Packages section should show two tables:

1. **Installed packages** — discovered from `vendor/*/package.json` (already working).
2. **Available packages** — pulled from the wheels-packages registry, with each row linking to that package's index page so the developer can read about it / install it.

**Actual behavior**

Only the installed-packages table renders. The available/registered packages section is missing — `debug.cfm` never calls the registry loader.

**Code reference**

- Template: `vendor/wheels/events/onrequestend/debug.cfm` (lines 313–341) — only iterates `application.wheels.packageMeta`; never calls the registry.
- Existing data source: `vendor/wheels/Public.cfc::\$loadRegistryPackages()` — already implemented and working.
- Registry client: `cli/lucli/services/packages/Registry.cfc` — fetches manifests from `wheels-dev/wheels-packages`.
- Working precedent: `packagelist.cfm` (the standalone packages browser view) already calls `application.wheels.public.\$loadRegistryPackages()` to populate its registry list. The same pattern needs to be mirrored in `debug.cfm`.

**Additional context**

The plumbing for fetching the registry already exists and is exercised by another view; this is a wire-up gap in the debug panel template, not a missing feature. A reasonable fix is to add a second table below the installed-packages table that iterates the result of `\$loadRegistryPackages()` and renders package name, version, description, and a link to the package index page on the registry.

bpamiri:
## Summary

The debug bar's **Tools → Packages** link (and the standalone packagelist page it opens) silently rendered an empty registry list in apps generated with `wheels new`. `Public.\$loadRegistryPackages` gated on `FileExists("/cli/lucli/services/packages/Registry.cfc")` — that path only exists in the framework dev repo. User apps ship `vendor/wheels/` without `cli/`, so the gate always returned an empty array and the registry section showed "No packages found in the registry."

This PR moves the registry reader into the framework so it ships with every generated app, and tightens where the registry list renders.

## Related Issue

Closes #2530

## What changed

- **New** `vendor/wheels/services/packages/{Registry,HttpClient,ManifestCache}.cfc` — pure-CFML services that fetch the wheels-packages registry over HTTPS with a 24h FS cache. The framework now has its own classpath-resolvable copies that ship with every `wheels new` app.
- **`vendor/wheels/Public.cfc`** — removed the `\$registryClientAvailable()` private fn and the file-existence gate; `\$getRegistryClient()` now instantiates `wheels.services.packages.Registry`.
- **`vendor/wheels/events/onrequestend/debug.cfm`** — removed the inline "Available from registry" subsection. The registry list is now scoped to the standalone Tools → Packages page (`packagelist.cfm`); the inline Environment > Packages section shows installed packages only. This keeps the debug bar compact and avoids triggering a registry walk on every dev-mode page load.
- **Test wiring** — `LoadRegistryPackagesSpec.cfc` now asserts the bundled Registry ships and is instantiable from a vanilla classpath. `debugPanelPackagesSpec.cfc` flipped its assertion: registry packages must NOT appear in the inline overlay. Removed obsolete `PublicWithoutRegistry.cfc` test asset.

## Why two copies (framework + CLI)

The CLI keeps its own copies under `cli/lucli/services/packages/` because the two contexts have different classpath mappings — the CLI uses LuCLI module paths (`modules.wheels.*`) and the framework uses `wheels.*`. Forcing them to share one copy means navigating cross-context Lucee mappings, which is risky and out of scope here. Dedup is a clean follow-up if the duplication becomes a maintenance burden.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement to existing feature
- [ ] Documentation update
- [ ] Refactoring

## Feature Completeness Checklist

- [x] **Tests** — Updated `LoadRegistryPackagesSpec.cfc` and `debugPanelPackagesSpec.cfc`. Existing tests for production gating, error capture, and bubble-up of unknown errors remain.
- [ ] **Framework Docs** — n/a (bug fix; no API surface change)
- [ ] **AI Reference Docs** — n/a
- [ ] **CLAUDE.md** — n/a
- [ ] **CHANGELOG.md** — could not modify in this turn
- [ ] **Test runner passes** — Could not run `tools/test-local.sh` locally because the brew-installed `wheels` bundles LuCLI as a native binary rather than exposing a standalone `lucli` on PATH. CI will run them on push.

## Test Plan

After homebrew picks up these changes:

1. \`wheels new test && cd test\`
2. \`wheels start\` (or manual lucli/box server)
3. Browse to the app, click the debug bar's **Tools → Packages**.
4. Verify both tables render: installed packages (empty) AND a "Browse registry" table listing the 6 first-party packages from `wheels-dev/wheels-packages` (sentry, hotwire, basecoat, legacy-adapter, i18n, seo-suite).
5. Verify the **Environment** tab's **Packages** section shows the installed table only — no registry list inline.

The first registry hit warms the 24h cache at `~/.wheels/cache/packages/`; subsequent page loads are FS-only.

## Screenshots / Output

n/a — UI behavior is described above; the user reported the empty state in the debug panel.


bpamiri:
Thanks @wheels-bot — addressed in 1462aecf5.

**Correctness fix**: extended `fetchManifest()`'s manifest guard to validate that `.versions` exists, is an array, and is non-empty. Threw the same `Wheels.Packages.RegistryMalformed` type so the existing per-package skip-on-malformed catch in `listAll()` actually catches every malformed shape now. Reviewer A's preferred location (in `fetchManifest`, not `listAll`) — keeps the invariant in one place and means cached manifests share the guard.

Mirrored the same validation into `cli/lucli/services/packages/Registry.cfc` so the two copies stay in sync per their in-file "keep in sync" note.

**Test coverage**: added `vendor/wheels/tests/specs/packages/RegistryFetchManifestSpec.cfc` with four cases — missing `versions` key, non-array `versions`, empty-array `versions`, and an end-to-end `listAll()` test that confirms a malformed entry is skipped instead of crashing the page. Required a framework-side `FakeHttpClient` test asset, mirrored from the CLI's stub but trimmed to the framework `HttpClient` surface (no `download()`).

**CHANGELOG**: the existing #2530 entry on line 153 was rewritten to match the final UX (registry list scoped to Tools → Packages standalone page; debug-bar Environment panel shows installed only). Appended a new entry for the validation fix.

bpamiri:
Thanks @wheels-bot — Reviewer B's nuance addressed in bf8d219a6.

B was right that my prior fix only validated on the fresh-fetch path. A manifest written by an older Registry version (that lacked the `versions` invariant) could survive in the on-disk cache, and `fetchManifest`'s cache-hit branch returned it without re-validation — same crash class, different entry point.

**Fix**: Extracted the manifest contract assertion into `$validateManifest()` and call it from both the cache-hit and fresh-fetch paths. Mirrored into the CLI copy. New spec case writes a stale-bad manifest directly to the cache, asserts it's marked fresh, then verifies `fetchManifest` re-validates on read. This covers both the existing-stale-entry concern B raised and any future required-field additions.

I went with the helper-function approach over B's "guard inside listAll()" suggestion because it keeps the contract assertion at the producer rather than spreading it across every consumer. The end-to-end `listAll()` skip-on-malformed test from the prior commit still passes — it now exercises the helper rather than inline checks.

PR #2598, opened 2026-05-12
fix(packages): use java.io.File.mkdirs() for cache dirs so Adobe CF works

### LINKED ISSUE #2567: Packages page fails on Adobe ColdFusion: `DIRECTORYCREATE` called with unsupported arguments
**Describe the bug**
When running a Wheels application on Adobe ColdFusion and navigating to the Packages page, an error is thrown:
```
Parameter validation error for the DIRECTORYCREATE function.
The function takes 1 parameter.
```

This indicates that the code is calling DirectoryCreate() with additional arguments that are supported in Lucee but not in Adobe ColdFusion.

**To Reproduce**
1. Run a Wheels application using Adobe ColdFusion.
2. Navigate to the Packages page.
3. Observe the error:
```
Parameter validation error for the DIRECTORYCREATE function.
The function takes 1 parameter.
```

**Expected behavior**
The Packages page should load successfully on Adobe ColdFusion without throwing any function compatibility errors.

**Screenshots**
Image

**Desktop:**
 - OS: Any
 - CFML Engine: Adobe ColdFusion

**Additional context**
Adobe ColdFusion’s DirectoryCreate() function supports fewer arguments than Lucee. If the code is using additional parameters such as mode or recursion flags, it will fail on Adobe ColdFusion.
For example:
`DirectoryCreate(path, true, true);`

works in Lucee but not in Adobe ColdFusion.

bpamiri:
## Summary

- `ManifestCache.$ensureDir` was calling `DirectoryCreate(path, true)`. The `createPath` flag is a Lucee extension — Adobe CF rejects the second argument with `"The function takes 1 parameter"`, crashing the **Tools → Packages** page on every fresh ACF install (#2567).
- Routes the recursive mkdir through `java.io.File.mkdirs()`, which has stable JVM-level semantics on Lucee, Adobe CF, and BoxLang. Same pattern as the existing `CreateObject("java", "java.lang.System")` call already in this file.
- Applies the identical fix to `cli/lucli/services/packages/ManifestCache.cfc`, the deliberate mirror called out in the framework file's docstring.

## Why the bot's "just drop the second argument" sketch wasn't enough

The wheels-bot triage on the issue suggested either dropping the second arg or branching on engine, claiming "Adobe CF creates parents by default." That's incorrect — Adobe CF's `DirectoryCreate(path)` only creates the leaf. If `~/.wheels/` doesn't yet exist (fresh install), a single-argument call would still fail on every engine. Java NIO bypasses the BIF entirely and recurses parents the same way everywhere.

## Regression coverage

- `vendor/wheels/tests/specs/packages/ManifestCacheEnsureDirSpec.cfc` — new spec exercising the multi-level-parent case the original `createPath=true` was meant to cover, plus the `manifests/` subdirectory creation path. Runs on every engine in the compat-matrix.
- Adds a parallel `it()` block to the existing `cli/lucli/tests/specs/packages/ManifestCacheSpec.cfc` so the CLI mirror keeps the same coverage.

## Docs

- `.ai/wheels/cross-engine-compatibility.md` gains a new entry documenting the BIF gotcha so the next contributor reaches for Java NIO before tripping on it.

## Test plan

- [ ] CI compat-matrix passes on Lucee 5/6/7 + Adobe CF 2018/2021/2023/2025 + BoxLang
- [ ] New `ManifestCacheEnsureDirSpec` is green in the **core** suite for every engine × DB combo
- [ ] New `ManifestCacheSpec.it` deep-nested case is green in the **CLI** suite on Lucee 7
- [ ] Manually verify the Tools → Packages page no longer 500s on Adobe CF (post-merge smoke)

Closes #2567


bpamiri:
Both findings addressed in [164cdfcdd](https://github.com/wheels-dev/wheels/pull/2598/commits/164cdfcdd):

- **mkdirs() return value**: `$ensureDir` now captures the boolean and throws `Wheels.Packages.CacheDir` when creation actually failed. The `!DirectoryExists` re-check swallows the benign concurrent-creation race. Same change mirrored into `cli/lucli/services/packages/ManifestCache.cfc`.
- **CHANGELOG**: entry added under `[Unreleased] → Fixed` describing the symptom, root cause, and fix.

[b1941ab55](https://github.com/wheels-dev/wheels/pull/2598/commits/b1941ab55) merges develop and resolves a parallel CHANGELOG insert from #2597.

PR #2600, opened 2026-05-12
fix(view): emit /wheels/guides redirect from body so Adobe CF stops erroring (#2569)

### LINKED ISSUE #2569: Guides page fails on Adobe ColdFusion with ‘Unable to add text to HTML HEAD tag’ error
**Describe the bug**
When running a Wheels application on Adobe ColdFusion and navigating to the Guides page, the following error is thrown:
```
Unable to add text to HTML HEAD tag.

ColdFusion was unable to add the text you specified to the output stream.This is probably because you have already used a CFFLUSH tag in your template or buffered output is turned off.
```

As a result, the Guides page does not render.

**To Reproduce**
1. Run a Wheels application using Adobe ColdFusion.
2. Navigate to the Guides page.
3. Observe the error shown above.

**Expected behavior**
The Guides page should render successfully on Adobe ColdFusion without throwing errors related to adding content to the HTML `<head>`.

**Screenshots**
Image

**Desktop:**
 - OS: Any
 - CFML Engine: Adobe ColdFusion

**Additional context**
This error usually occurs in Adobe ColdFusion when code attempts to add content to the <head> after output has already started. Possible causes include:
- Output being generated before `contentFor("head")` or similar helpers are called
- Use of `cfflush`
- Response buffering differences between Lucee and Adobe ColdFusion

bpamiri:
## Summary

- Drop the late `<cfhtmlhead>` call in [`vendor/wheels/public/docs/guides.cfm`](vendor/wheels/public/docs/guides.cfm). The wrapper view (`vendor/wheels/public/views/guides.cfm`) includes the layout header (which streams `</head>`) before this docs view runs, so injecting a meta refresh via `cfhtmlhead` arrives after the head has already been written. Lucee tolerates it; Adobe ColdFusion throws `"Unable to add text to HTML HEAD tag"` and the page fails to render.
- Emit the redirect from the body instead: a hidden `<div id="wheels-guides-redirect" data-url="…">` carries the (encoded) target, and a small inline `<script>` triggers `window.location.href` after 3 seconds. The URL flows through `encodeForHTMLAttribute`, matching the visible anchor a few lines below, so we don't introduce a new encoder.
- Add a convention guard spec at [`vendor/wheels/tests/specs/view/guidesRedirectHeadSpec.cfc`](vendor/wheels/tests/specs/view/guidesRedirectHeadSpec.cfc) that strips CFML comments and then asserts `docs/guides.cfm` no longer uses `cfhtmlhead`, plus a companion check that the body still emits a `window.location.href` redirect so the mechanism cannot silently disappear.

Fixes #2569.

## Test plan

- [x] `wheels validate` clean.
- [x] Full view-spec suite green locally on Lucee 7 + SQLite (508 pass, 0 fail).
- [x] Curling `/wheels/guides` on Lucee renders the moved-page panel, contains the hidden `wheels-guides-redirect` div with the encoded URL, and emits the `setTimeout` redirect — no `cfhtmlhead` left, no meta refresh in `<head>`.
- [ ] CI compat matrix (Adobe 2023/2025) renders the page without the `"Unable to add text to HTML HEAD tag"` error.


bpamiri:
Thanks for the review. Addressed both items in [cc8cde4e4](https://github.com/wheels-dev/wheels/pull/2600/commits/cc8cde4e4):

- **CHANGELOG entry** — added a one-line entry at the top of `[Unreleased] > Fixed` matching the format of recent neighbors (#2563, #2530), including the issue reference.
- **`Signed-off-by` trailer** — committed with `git commit -s`, so the fixup commit carries the trailer. The original commit (`07e84e061`) does not; since the PR is on track for a squash merge, the trailer will land in the final on-`develop` commit. Happy to amend the original via a force-push if you'd prefer a single signed commit on the branch, but that needs the standing "confirm before force-push" guardrail to be explicitly waived.

No behavior change since the last revision — Lucee 7 + SQLite view suite still 508/508, the rendered `/wheels/guides` body still emits the hidden `wheels-guides-redirect` div + `setTimeout` redirect, and `<head>` carries no meta-refresh injection.

bpamiri:
Thanks Reviewer B. Addressed the stale CLAUDE.md anchor in [af7508ba5](https://github.com/wheels-dev/wheels/pull/2600/commits/af7508ba5):

- The `Chr(60)` justification in [`guidesRedirectHeadSpec.cfc`](https://github.com/wheels-dev/wheels/blob/claude/romantic-merkle-eb2e74/vendor/wheels/tests/specs/view/guidesRedirectHeadSpec.cfc) no longer cites a fabricated `"Lucee Tag Scanner"` section. It now points at `MigratorViewIconsSpec.cfc` (line 41/46/61, same `Chr(60) & "svg"` / `Chr(60) & "i\s>"` pattern) as the in-repo precedent and describes the underlying "any compilation error in `tests.specs.*` takes the whole suite down" failure mode without inventing a doc heading.

On the DCO-framing observation: agreed it reads as self-contradictory in A's review. No code action there — leaving it as bot-on-bot calibration feedback for the next round.

Diff: <https://github.com/wheels-dev/wheels/pull/2600/commits/af7508ba5>

bpamiri:
Thanks Reviewer B. Round-2 read:

- The DCO false-positive in A's text (`07e84e061` lacking sign-off) is a fair catch on A's framing but not actionable on the code — the squash-merge path I described earlier already resolves it on `develop`.
- B's round-1 stale-spec finding was already cleared at the current head (`af7508ba5` → `c9562c72e`); B confirmed it's moot.
- `af7508ba5` post-dating A's review is informational, not a regression — A was re-triggered automatically and re-completed on the merge commit.

No code change needed. Joint `wheels-bot:converged-approve` marker stands; waiting on the in-flight Reviewer B round on the post-merge head (`c9562c72e`) and the Lucee 7 + SQLite check before flipping `UNSTABLE` → ready.

PR #2656, opened 2026-05-13
ci(web): add manual workflow to refresh visual regression baselines

bpamiri:
## Summary

- Adds [.github/workflows/refresh-visual-baselines.yml](.github/workflows/refresh-visual-baselines.yml) — a `workflow_dispatch` trigger that refreshes any of the 6 visual baselines under `web/tests/visual-baselines/` and pushes the new PNG(s) back to whichever branch the user selected in the "Run workflow" UI.
- Closes the gap when intentional content (new blog post, new guide page, planned layout tweak) makes the `visual-regression` check fail on a PR. Before this, the only fix was downloading the diffs artifact and hand-copying each `*.actual.png` over the matching baseline.
- Complements the existing `refresh-packages-baseline.yml` (which fires on `repository_dispatch: registry-updated` and opens a chore PR) — this new workflow is the manual escape hatch that commits directly to the PR branch that's currently failing.

## How to use it

1. Visual-regression fails on a PR.
2. Confirm via the `visual-regression-diffs` artifact that the change is intentional.
3. **Actions → Refresh visual baselines → Run workflow**:
   - **Use workflow from:** PR's source branch
   - **Which baseline(s) to refresh:** pick the affected site, or `all`
4. The workflow rebuilds the web/ sites, regenerates baseline(s), commits to the same branch, and pushes.
5. The push retriggers CI; `visual-regression` now passes. (Or click "Re-run failed jobs" on the original run.)

## Design notes

- **`type: choice` with 7 options** (`all` + the 6 site names). Single-select; multi-site refreshes mean running it twice. Default `all` because it's idempotent — already-correct baselines won't drift.
- **No explicit `ref:` on checkout** — `workflow_dispatch` already checks out `github.ref`, which is the branch the user picked in the UI. No extra input needed.
- **Routes every user-controllable value (`inputs.sites`, `github.ref_name`) through `env:`** before shelling out, so a branch name with shell metacharacters can't inject into `run:` blocks ([Actions injection background](https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/)).
- **No `[skip ci]` on the commit** — unlike `web-deploy.yml`'s blog auto-refresh, we *want* the push to retrigger CI so the PR's `visual-regression` flips green automatically.
- **Builds all sites** even when refreshing one. Saves a per-site `pnpm --filter` mapping table; adds ~30-60s to a low-frequency manual run.

## Test plan

- [ ] Manually trigger the workflow against this branch with `sites=blog` and verify it produces "No baseline drift detected" (no real change to commit).
- [ ] Intentionally modify a baseline-driving page on a throwaway branch (e.g. tweak `sites/blog` index header), let `visual-regression` fail on a draft PR, then trigger this workflow with `sites=blog` against that branch — confirm the commit lands, CI reruns, and the check flips green.
- [ ] Confirm `github-actions[bot]` has push access to `develop` for the case where the user runs this against `develop` directly (existing `web-deploy.yml` already pushes to `develop`, so this should already work).

### DIFF
--- a/.github/workflows/refresh-visual-baselines.yml
+++ b/.github/workflows/refresh-visual-baselines.yml
@@ -57,6 +57,17 @@ jobs:
   refresh:
     name: Refresh ${{ inputs.sites }} baseline(s) on ${{ github.ref_name }}
     runs-on: ubuntu-latest
+    # Serialize per-branch so two simultaneous refreshes can't both build,
+    # commit, and race on `git push` (the second would be rejected as
+    # non-fast-forward). Matches the pattern web-deploy.yml uses for the
+    # visual-regression job.
+    concurrency:
+      group: refresh-visual-baselines-${{ github.ref }}
+      cancel-in-progress: true
+    # Worst-case real path: install + Chromium download + full site build +
+    # six screenshots is well under 10 minutes. 30 is a generous ceiling
+    # that cuts off runaway billable minutes if pnpm or Playwright hangs.
+    timeout-minutes: 30
     permissions:
       contents: write
     steps:
@@ -147,7 +158,7 @@ jobs:
           # above, but writing through a file keeps that contract obvious.
           commit_msg_file="$(mktemp)"
           {
-            printf '%s\n\n' "chore(docs): refresh visual baseline(s) ($SITES)"
+            printf '%s\n\n' "chore(web): refresh visual baseline(s) ($SITES)"
             printf '%s\n' "Manually triggered baseline refresh via"
             printf '%s\n\n' ".github/workflows/refresh-visual-baselines.yml on branch $BRANCH."
             printf '%s\n' "Run when an intentional content/layout change makes the visual-regression"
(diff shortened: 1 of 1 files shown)


bpamiri:
Thanks for the review. Addressed in [b7aee2930](https://github.com/wheels-dev/wheels/commit/b7aee2930):

1. **Concurrency group** ✅ — Added `refresh-visual-baselines-${{ github.ref }}` with `cancel-in-progress: true`, matching the pattern from `web-deploy.yml`.
2. **`timeout-minutes: 30`** ✅ — Added at job level.
3. **`chore(docs):` → `chore(web):`** ✅ — Fixed the auto-generated commit message. Also matches the precedent in cd8ce3fa8.
4. **`Signed-off-by:` trailer** — Added on the fix commit (b7aee2930). The original commit (acdfe30421) still lacks it. The DCO check doesn't appear in this PR's required-checks list, and a squash-merge will fold both commits into a single signed final commit on `develop`, so I left acdfe30421 unrewritten rather than force-push. Happy to amend + force-push if you'd rather keep the per-commit DCO clean — let me know.

Issue #2663, opened 2026-05-14
ci: Oracle datasource broken across all engines — Invalid URL, missing DBMS_LOCK, constraint cleanup


bpamiri:
**Describe the bug**

The Oracle database test path in the compat-matrix is broken at multiple layers, affecting **every** engine that exercises it (Lucee 6/7, Adobe 2023/2025, BoxLang). The failures fall into three sub-clusters:

1. **Datasource registration failure (Lucee 6)** — Oracle test JSON for Lucee 6 contains `java.sql.SQLException: Invalid Oracle URL specified` and reports 0 specs run. The connection never establishes.
2. **DBMS_LOCK package not installed (any engine)** — `wheels.tests.specs.model.lockingSpec` errors with `Oracle advisory locks require DBMS_LOCK package setup.` (4 occurrences).
3. **Test cleanup leaves constraints behind** — `ORA-02264: name already used by an existing constraint` and `ORA-01003: no statement parsed` (~6 occurrences in `migratorSpec` and `bulkOperationsSpec`). Same root cause class as the Lucee/MySQL `Duplicate foreign key constraint name` issue tracked separately.

**Evidence**

[Compat-matrix run #25837380625](https://github.com/wheels-dev/wheels/actions/runs/25837380625) (2026-05-13). Oracle errors on BoxLang totaled **134** — the highest count per DB. On Lucee 6 the count was 0 because tests didn't run (datasource failure).

**Expected behavior**

Oracle tests should at minimum run to completion on every engine (currently fails on Lucee 6 before tests start). The advisory-lock and constraint-cleanup failures should be deterministic and either pass or skip gracefully.

**Likely fix areas**

This is **CI-infrastructure work**, not framework code. Three independent threads:

1. **Lucee 6 + Oracle JDBC URL** — Lucee 6's `CFConfig.json` for Oracle in [tools/docker/](tools/docker/) — verify the JDBC URL format. The matrix `compose.yml` shows the Oracle service definition; the per-engine Oracle datasource config is what needs the fix. The Adobe + BoxLang Oracle setups work, so the diff between their configs is the lead.
2. **DBMS_LOCK package** — needs to be granted to `wheelstestdb` in the Oracle container's initialization script. Probably an add-grant in `tools/docker/oracle/init.sql` or equivalent.
[...]

PR #2687, opened 2026-05-15
test(migrator): clean migrator-version tables to fix FK collision (#2663 sub-cluster 3, #2664)

### LINKED ISSUE #2664: test: migrator FK constraint cleanup is non-idempotent — fails on MySQL/H2/SQL Server reruns
**Describe the bug**

`wheels.tests.specs.migrator.migratorSpec` (the `migrateTo :: uses specified versions table name` suite) creates a foreign key constraint named `fk_wheels_level` but does not drop it on teardown. When the spec runs against the same database a second time (which happens inside one matrix job because the spec executes the migration up *and* re-executes it), the constraint creation errors:

- **MySQL**: `Duplicate foreign key constraint name 'fk_wheels_level'`
- **H2**: `Constraint "FK_WHEELS_LEVEL" already exists; SQL statement: ALTER TABLE c_o_r_e_migrator_versions ADD CONSTRAINT fk_wheels_level FOREIGN KEY ...`
- **SQL Server**: `There is already an object named 'fk_wheels_level' in the database.`
- **Oracle**: `ORA-02264: name already used by an existing constraint` (counted in #2663)

**Evidence**

[Compat-matrix run #25837380625](https://github.com/wheels-dev/wheels/actions/runs/25837380625) (2026-05-13).

- Lucee 6: 1 error each on `h2`, `mysql`, `sqlserver` (3 total).
- Same pattern appears in BoxLang results (Oracle path counts in #2663).
- Cockroach + Postgres: pass — they tolerate `IF NOT EXISTS`-style behavior or the constraint name handling differs.

**Expected behavior**

The migrator spec should run repeatedly within the same test session without leaving constraints behind. Either:

1. Drop the constraint in the suite's `afterEach` / `afterAll`, **or**
2. Use a uniquely-named constraint per spec run (timestamp suffix), **or**
3. Use the down-migration to clean up after the up.

**Root cause**

Pure test-cleanup bug. The migrator spec uses a hardcoded constraint name and re-runs without teardown. Cockroach/Postgres silently tolerate the redefinition; MySQL/H2/SQL Server/Oracle correctly reject it.

**Suggested fix shape**

Add a teardown step that drops `fk_wheels_level` (and any other named constraints created in the spec) before each run, **or** rewrite to use a per-test-run unique name. Low effort.

**Additional context**

- Part of the #2649 decomposition (Tier 3 — test cleanup, not framework).
- This is engine-agnostic and predates BoxLang's matrix participation — it would have shown up on the Lucee 6 + MySQL row since matrix runs started, but was masked by `continue-on-error: true` at the workflow level.
- Worth pairing with #2663's Oracle constraint cleanup since the underlying spec fix is the same.

Related: #2649, #2663

bpamiri:
## Summary

Closes the third and final sub-cluster of #2663 and resolves #2664. The migrator spec's `migrateTo` describe block was leaking `wheels_migrator_versions` / `c_o_r_e_migrator_versions` tables (plus their `fk_wheels_level` FK constraint) between tests. When the *"uses specified versions table name"* test swaps `migratorTableName` to `c_o_r_e_migrator_versions`, the bootstrap path in `Migrator.cfc::$getVersionsPreviouslyMigrated()` adds the same `fk_wheels_level` constraint to a different table — MySQL, H2, SQL Server, and Oracle scope FK names per-schema rather than per-table, so the second CREATE failed:

- MySQL: `Duplicate foreign key constraint name 'fk_wheels_level'`
- H2: `Constraint "FK_WHEELS_LEVEL" already exists`
- SQL Server: `There is already an object named 'fk_wheels_level' in the database.`
- Oracle: `ORA-02264: name already used by an existing constraint`

The Oracle cascade also produces `ORA-01003: no statement parsed` in `bulkOperationsSpec` because the partially-committed DDL leaves the JDBC session in a broken state.

## Fix

Mirror the cleanup pattern already used by the **F15 Phase 1** describe block (`migratorSpec.cfc:170`): in `beforeEach`, drop both table-name families inside try/catch so the bootstrap recreates them cleanly per test. `wheels_levels` is left in place — it is referenced by the FK, not the source of it, and the bootstrap is happy to reuse it.

## #2663 status

This completes the three-cluster fix for #2663:
- ✅ Sub-cluster 1 (Lucee 6 Oracle JDBC URL) — #2683
- ✅ Sub-cluster 2 (DBMS_LOCK / lockingSpec guard) — #2670
- ✅ Sub-cluster 3 (constraint cleanup) — **this PR**

Removing `oracle` from `SOFT_FAIL_DBS` in `compat-matrix.yml` is intentionally left for a follow-up once a green Oracle matrix run is observed.

## Test plan

- [ ] `compat-matrix` workflow passes on Lucee 6/7, Adobe 2023/2025, BoxLang
- [ ] Oracle errors drop to zero for `migratorSpec` and `bulkOperationsSpec`
- [ ] MySQL/H2/SQL Server `migratorSpec` errors drop from 1 → 0 each on Lucee 6 (per #2664 evidence)
- [ ] No regression in the existing F15 Phase 1 / Phase 2 / redomigration describe blocks (they have their own teardown)

Local SQLite / Docker matrix runs were not possible in this environment (no Docker daemon, no `wheels` CLI). Relying on CI.

https://claude.ai/code/session_01XWJ2m7SfxhuK7N6YCWv4E3

---
_Generated by [Claude Code](https://claude.ai/code/session_01XWJ2m7SfxhuK7N6YCWv4E3)_

### DIFF
--- a/vendor/wheels/tests/specs/migrator/migratorSpec.cfc
+++ b/vendor/wheels/tests/specs/migrator/migratorSpec.cfc
@@ -60,15 +60,7 @@ component extends="wheels.WheelsTest" {
 				for (local.table in ["c_o_r_e_bunyips", "c_o_r_e_dropbears", "c_o_r_e_hoopsnakes", "migrations"]) {
 					migration.dropTable(local.table)
 				}
-				// Drop both migrator-version table families so the bootstrap path
-				// in $getVersionsPreviouslyMigrated() can recreate them cleanly.
-				// The "uses specified versions table name" test swaps the configured
-				// table to `c_o_r_e_migrator_versions`, and its bootstrap adds an FK
-				// constraint named `fk_wheels_level`. Without this cleanup the
-				// constraint collides with the one already created on
-				// `wheels_migrator_versions` by earlier tests, since MySQL, H2,
-				// SQL Server, and Oracle scope FK names per-schema rather than
-				// per-table (ORA-02264 / "Duplicate foreign key constraint name").
+				// Drop both FK-bearing tables — MySQL/H2/SQLServer/Oracle scope FK names per-schema so fk_wheels_level collides otherwise.
 				for (local.table in ["c_o_r_e_migrator_versions", "wheels_migrator_versions"]) {
 					try { migration.dropTable(local.table); } catch (any e) {}
 				}
--- a/vendor/wheels/tests/specs/migrator/migratorSpec.cfc
+++ b/vendor/wheels/tests/specs/migrator/migratorSpec.cfc
@@ -60,6 +60,18 @@ component extends="wheels.WheelsTest" {
 				for (local.table in ["c_o_r_e_bunyips", "c_o_r_e_dropbears", "c_o_r_e_hoopsnakes", "migrations"]) {
 					migration.dropTable(local.table)
 				}
+				// Drop both migrator-version table families so the bootstrap path
+				// in $getVersionsPreviouslyMigrated() can recreate them cleanly.
+				// The "uses specified versions table name" test swaps the configured
+				// table to `c_o_r_e_migrator_versions`, and its bootstrap adds an FK
+				// constraint named `fk_wheels_level`. Without this cleanup the
+				// constraint collides with the one already created on
+				// `wheels_migrator_versions` by earlier tests, since MySQL, H2,
+				// SQL Server, and Oracle scope FK names per-schema rather than
+				// per-table (ORA-02264 / "Duplicate foreign key constraint name").
+				for (local.table in ["c_o_r_e_migrator_versions", "wheels_migrator_versions"]) {
+					try { migration.dropTable(local.table); } catch (any e) {}
+				}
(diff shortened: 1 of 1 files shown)


bpamiri:
Superseded by #2685 — merged to `develop` while this PR was in review. #2685 implements a strict superset of the fix here:

- Drops all four migrator system tables (`wheels_migrator_versions`, `c_o_r_e_migrator_versions`, `wheels_levels`, `c_o_r_e_levels`) instead of just the two version tables.
- Symmetric cleanup in `afterEach` as well as `beforeEach`.
- Adds a regression test for #2664.

Rebasing this branch onto develop produces an empty diff, so closing as superseded. Sub-cluster 3 of #2663 is now covered on `develop` via #2685; sub-clusters 1 and 2 were already addressed by #2683 and #2670. Removing `oracle` from `SOFT_FAIL_DBS` remains as the follow-up once a green Oracle matrix run is observed.

---

Issue #2677, opened 2026-05-14
`wheels deploy server bootstrap` is interpreted as LuCLI server command instead of deploy subcommand


zainforbjs:
**Describe the bug**
The documentation describes the command `wheels deploy server bootstrap` as the way to bootstrap Docker on deployment hosts.

However, running the command does not execute any deploy/bootstrap logic. Instead, the CLI falls through to the LuCLI server command help output.

Documentation URL:
[Deployment -> first-deploy -> #walk-through -> Part 5](https://guides.wheels.dev/v4-0-1-snapshot/deployment/first-deploy/#walk-through)

**To Reproduce**
1. Create a new Wheels application:
```
wheels new myapp
```
2. Run:
```
wheels deploy server bootstrap
```
3. Observe output.

**Expected behavior**
- Runs:
  ```
  which docker || curl -fsSL https://get.docker.com | sh
  ```
- Executes on remote hosts
- Safe/idempotent
- Validates SSH/sudo access

**Actual behavior**
- Instead of executing deployment bootstrap logic, the CLI displays the LuCLI server help output:
```
Usage: server [-hV] [COMMAND]
Manage Lucee server instances
```
followed by the standard LuCLI server command list.
- No deployment/bootstrap action occurs.

**Scrneeshot**
<img width="559" height="436" alt="Image" src="https://github.com/user-attachments/assets/165d55d1-8627-47cf-93a3-afe8d7c2f091" />

**Desktop:**
 - OS: macOS
 - Version: Wheels 4.0.0 (development)

PR #2699, opened 2026-05-15
fix(cli): add wheels deploy fetch-secrets/extract-secrets/print-secrets flat aliases (#2697)

### LINKED ISSUE #2697: `wheels deploy secrets` fetch invokes LuCLI global secrets command instead of Wheels deploy subcommand
**Describe the bug**
Running `wheels deploy secrets fetch` does not execute a Wheels deploy command.
Instead, the command falls through to LuCLI's built-in secrets command and displays the generic LuCLI secrets help output.

**To Reproduce**
1. Create a Wheels application
2. Run:
```
wheels deploy secrets fetch
```
3. Observe the output

**Actual behavior**
The command resolves to LuCLI's global secrets command:
```
Usage: secrets [-hV] [COMMAND]
Manage secrets for LuCLI projects
```
instead of executing a Wheels deploy workflow.

Full output:
testapp % wheels deploy secrets fetch
```
Usage: secrets [-hV] [COMMAND]
Manage secrets for LuCLI projects
  -h, --help      Show this help message and exit.
  -V, --version   Print version information and exit.
Commands:
  init      Initialize the local encrypted secret store
  set       Set or update a secret value
  list      List stored secrets (names and metadata only)
  rm        Remove a stored secret
  get       Retrieve a secret value
  provider  Manage secret providers
```

**Expected behavior**
`wheels deploy secrets fetch` should execute the intended Wheels deploy secrets workflow.
Example:
```
wheels deploy secrets fetch
Fetching deploy secrets...
```

**Screenshots**
Image

**Desktop (please complete the following information):**
 - OS: macOS
 - Version: Wheels 4.0.0 (development)

bpamiri:
## Summary

Fixes [#2697](https://github.com/wheels-dev/wheels/issues/2697) — `wheels deploy secrets <verb>` (fetch/extract/print) never reaches the deploy dispatcher because LuCLI's picocli root registers `secrets` as its own top-level subcommand (the local credential store: init/set/list/rm/get/provider) and intercepts the token before `Module.cfc::deploy()` can dispatch on it.

This is the same picocli command-shadowing pattern documented as gotcha #7 in `CLAUDE.md` (the `server` collision, tracked as [#2677](https://github.com/wheels-dev/wheels/issues/2677) and fixed in [#2690](https://github.com/wheels-dev/wheels/pull/2690)).

## Approach

Mirror the [#2690](https://github.com/wheels-dev/wheels/pull/2690) flat-alias pattern. Add three top-level cases to the deploy() switch:

- `wheels deploy fetch-secrets [KEY...]` → `DeploySecretsCli.fetch`
- `wheels deploy extract-secrets [KEY]` → `DeploySecretsCli.extract`
- `wheels deploy print-secrets` → `DeploySecretsCli.print`

The existing `case "secrets":` branch is retained unchanged for MCP and programmatic callers (same treatment as `case "server":`).

Verb-noun naming (vs. bare `fetch`/`extract`/`print`) for namespace safety — `fetch` and `print` are generic enough that future deploy subcommands could legitimately claim them.

## Changes

- **`cli/lucli/Module.cfc`** — three new `case` arms in `deploy()`, just before `case "secrets":`. ~20 lines including the explanatory docblock.
- **`cli/lucli/tests/specs/commands/DeployCommandSpec.cfc`** — extends the existing `#2677` dispatcher regression spec with 7 new `it` blocks covering all three flat aliases (positive routing, error propagation, and a regression guard for the retained `secrets <verb>` branch).
- **CLAUDE.md** — new gotcha #8 alongside the existing #7 (`server` collision); secrets quick-reference promotes flat aliases as canonical.
- **`v4-0-1-snapshot` guides** — three new doc pages (`fetch-secrets.mdx`, `extract-secrets.mdx`, `print-secrets.mdx`), `:::caution` blocks on the four nested-form pages pointing to the flat aliases, plus updated `deployment/secrets.mdx`, `deployment/hooks.mdx`, `deployment/migrating-from-kamal.mdx`, and `command-line-tools/commands/deploy/index.mdx`.
- **`v4-0-0` guides** — left untouched (frozen).
- **`CHANGELOG.md`** — `[Unreleased] § Fixed` entry alongside the existing `#2677` line.
- **`docs/superpowers/specs/2026-05-15-deploy-secrets-flat-aliases-design.md`** — design spec.

## Test plan

- [x] Static trace verification of dispatch paths (input arg array → `$deployStripFlags` output → opts struct → CLI method invocation). All seven dispatcher tests pass under static analysis.
- [ ] CI: `compat-matrix` + CLI test bundle. Local execution skipped due to a worktree-server conflict (another worktree is currently holding the `wheels` server registration on this machine; the `wheels start` CLI doesn't expose a name override to coexist). The test logic is mechanical (`toThrow(regex=...)` and `toBe(...)` assertions on a Module.deploy() dispatcher that previously throws `Unknown deploy subcommand: <verb>` and now routes correctly), so CI is the authoritative gate.
- [ ] Manual smoke: `wheels deploy fetch-secrets --adapter=op --from=op://Test/T DUMMY_KEY` on a workstation with `op` configured.

## Follow-up (out of scope)

While reviewing the secrets surface I noticed two pre-existing bugs that I deliberately left alone:

- `wheels deploy secrets extract --key=<name>` documents the `--key=` flag, but `DeployArgsParser` has no parser arm for it — the flag gets stripped silently and the key never makes it into opts. The dispatcher's positional-arg fallback covers the path my fix uses; the documented flag form remains broken.
- `wheels deploy secrets print --project-root=<path>` is similarly documented but unparsed.

These are tiny parser additions; I'll spin them off as their own issue if no one else gets there first.


bpamiri:
## Author response to Reviewer A & B

Both reviews agreed: approve and merge after CI passes. Three threads of feedback:

### 1. `##2697` in commit `6439e6440` subject (Reviewer A)
**Acknowledged, not fixed.** The `##` is a CFML escape that has no meaning in a git commit message and breaks GitHub's auto-link from that specific commit. Reviewer B independently noted this isn't worth a force-push to fix — I agree. The PR title, body, CHANGELOG entry, and every other in-history reference uses the correct `#2697` form. Affects only the auto-link from one historical commit subject; semantic content is intact.

### 2. Multi-key fetch-secrets coverage (Reviewer A)
**Addressed in [fe834a8ac](https://github.com/wheels-dev/wheels/commit/fe834a8ac).** Added a "boundary smoke" `it` block that drives three positional keys + an unknown adapter through the dispatcher. A's caveat is preserved in the docblock: the adapter check fires before the keys check, so this doesn't verify the slice landed in `opts.keys` end-to-end (would need a real adapter CLI for that). What the new test does cover: the `for fsi=2 to arrayLen(positional)` loop completes without crashing on a non-trivial positional list — any off-by-one would surface as an index-out-of-range rather than UnknownAdapter. Mirrors the existing `exec` smoke pattern at line 43.

### 3. Visual regression CI failure
**Pre-existing on develop, not introduced by this PR.** The [Deploy static sites run on develop after PR #2695](https://github.com/wheels-dev/wheels/actions/runs/25918326938) shows the same `[blog] FAIL — 43,560 pixels differ` against `/`. Introduced by [#2694](https://github.com/wheels-dev/wheels/pull/2694) (blog API rate-limiting), baselines weren't refreshed. This PR doesn't touch `web/blog`, doesn't change any other site's baseline-tracked surface, and adds three brand-new doc pages under `v4-0-1-snapshot/command-line-tools/commands/deploy/` (paths the visual-regression script doesn't screenshot). The blog baseline refresh is a separate-PR concern.

Ready for merge when CI is green ignoring the pre-existing visual regression.

bpamiri:
## Author response to Reviewer A round 2

> The \`##2699\` is inside a \`//\` comment, which CFML does not evaluate, so it causes no compilation error. But \`//\` comments are not CFML output — the \`##\` escape is unnecessary and slightly misleading.

Fixed in [c3fb9a8ed](https://github.com/wheels-dev/wheels/commit/c3fb9a8ed). Agreed — the test docstrings in this file got mixed because the same describe() *string literals* legitimately need the `##` escape (per the CLAUDE.md gotcha that bit #2690's first commit), but `//` line comments don't, so the `##` is just noise there.

PR #2734, opened 2026-05-16
feat(web/blog): anatomy of a Wheels package post + manifest doc fixes

bpamiri:
## Summary

Second post in the post-GA blog series after [Skip the Plugin](https://blog.wheels.dev/posts/skip-the-plugin-rate-limited-api). Worked walkthrough of the package system: loader, manifest, mixin allowlist (and why `controller` is the everyday target), per-method overrides, the `requires`/`replaces`/`suggests` graph, the mapping alias, service providers, error isolation, the CLI.

Drafting surfaced two doc-drift bugs in the package manifest reference. Both got fixed in the same PR so the article's "What changed while writing this post" section reflects shipped reality rather than aspirational truth.

## Doc fixes shipping alongside the article

**1. `dependencies` → `requires`** (`web/sites/guides/.../packages.mdx` ×2, `CLAUDE.md`)

Both copies of the public `Packages` guide and `CLAUDE.md` documented the manifest's inter-package dependency field as `dependencies`. The loader has always read `requires` (`vendor/wheels/ModuleGraph.cfc:171`). Test fixtures (`vendor/wheels/tests/_assets/packages/depA/package.json`) confirm. Anyone copying the example would have shipped a package that loaded but silently ignored its declared dependencies — no error, no warning, just broken at the first missing dep. All three docs now use `requires` and document `replaces` and `suggests` (also previously undocumented).

**2. `wheelsVersion` "logged" → hard skip** (`packages.mdx` ×2)

The guide described mismatches as "logged" — accurate but undersells the consequence. `PackageLoader.cfc:325-339` excludes incompatible packages from the load order before the CFC is instantiated and records them in `failedPackages`. Both guide copies now say so explicitly, naming the log content (constraint + running version) so a debugger reading "why isn't my mixin showing up" can connect the dots.

## Files

- `web/content/blog/posts/anatomy-of-a-wheels-package.md` — new (~2,700 words, dated 2026-05-22 to match the Friday cadence of the series)
- `web/sites/guides/src/content/docs/v4-0-1-snapshot/digging-deeper/packages.mdx` — `dependencies` → `requires` + add `replaces`/`suggests`; tighten `wheelsVersion`
- `web/sites/guides/src/content/docs/v4-0-0/digging-deeper/packages.mdx` — same fixes (stable docs)
- `CLAUDE.md` — manifest example + field reference updated

## Validation

Every concrete claim in the article was checked against source. Verified against:
- `vendor/wheels/PackageLoader.cfc` — discovery loop, manifest parsing, mixin collection, lifecycle-hook exclusion list (`init,onPluginLoad,onPluginActivate,register,boot`), mapping derivation regex, error types (`Wheels.PackageInvalidMixinTarget`, `Wheels.PackageNoCFC`, `Wheels.PackageInvalidManifest`), `wheelsVersion` enforcement (`$isCompatibleVersion`)
- `vendor/wheels/ModuleGraph.cfc` — `requires`/`replaces`/`suggests` semantics, topological sort, circular-dependency handling
- `vendor/wheels/ServiceProviderInterface.cfc` — `register(container)` / `boot(app)` two-phase contract
- `vendor/wheels/SemVer.cfc` — `satisfiesAll()` referenced in the dependency section
- `vendor/wheels/Plugins.cfc` — confirmed plugins still load with a deprecation warning
- `cli/lucli/services/packages/Registry.cfc` — default registry `wheels-dev/wheels-packages` + `WHEELS_PACKAGES_REGISTRY` env var override
- Test fixtures under `vendor/wheels/tests/_assets/packages*/` — for canonical manifest shapes

## Test plan

- [ ] Web preview renders the blog post correctly (frontmatter, code blocks, tables)
- [ ] `packages.mdx` guide pages render in both `v4-0-1-snapshot` and `v4-0-0` paths
- [ ] No broken internal links in the article (one rel link to the rate-limiter post)
- [ ] Series teaser at the end correctly names the next post ("Wheels + Claude: building a feature via the stdio MCP")

https://claude.ai/code/session_01RseAJ1xUfRc7zQv8NBwa8j

---
_Generated by [Claude Code](https://claude.ai/code/session_01RseAJ1xUfRc7zQv8NBwa8j)_


bpamiri:
Heads-up on the persistent `Visual regression` failure (three runs now, three commits, same root cause).

**The regression test itself is passing.** The job is failing on the subsequent `Commit pre-refreshed blog baseline` step, which only runs at all when `steps.regress.outcome == 'success'` (workflow line 277). What's actually happening:

1. `Pre-refresh blog baseline before regression test` runs (blog content changed).
2. `Run visual regression` → success.
3. `Commit pre-refreshed blog baseline` rebases against develop and tries to push the new `web/tests/visual-baselines/blog.png` directly. After 5 rejected pushes against a contended develop tip it falls through to `exit 1`.

Three runs, same step, same exit. Not a flake — looks structural: develop has been actively merging during the window this PR's CI has been running, so the auto-refresh loses the push race every time.

The workflow's own message in that branch reads "The auto-refresh will be retried on the next push that touches `web/content/blog/posts/`" — which is true, but the `exit 1` still fails the PR's check in the meantime. That mismatch (intent is best-effort, exit code is fail-loud) means a routine blog post can block on develop activity that has nothing to do with the post.

Two paths forward, neither in this PR's scope to land alone:

- **Immediate**: trigger `refresh-visual-baselines.yml` on this branch (Actions → Refresh visual baselines → "Use workflow from" = `claude/add-blog-series-articles-bXKcg`, "Which baseline" = `blog`). That commits the refreshed PNG to the branch directly, the next CI run sees a matching baseline, and the commit-and-push step short-circuits (`git diff --cached --quiet` returns success and exits 0).
- **Workflow fix** (separate PR): swap the final `exit 1` for `exit 0` with a `::warning::` so an auto-refresh push race doesn't fail PR checks, while still surfacing the unsuccessful push in the Annotations panel.

PR content is otherwise approved by both reviewers (Reviewer B converged-approve on `df30d9d`). Happy to file a follow-up issue for the workflow fix if useful.

---

bpamiri:
Quick follow-up on this PR's CI state — the only file added since the approved `df30d9d` is `docs/releases/blog-skeletons/social-post-anatomy-of-a-wheels-package.md` (pure Markdown, outside `web/`, not indexed by any site build). On the new run for `efcf61f`:

- **`Visual regression` failed (4th time)** — same root cause as before. Diagnosis covered in the earlier comment; nothing has changed.
- **`Deploy packages` failed (new)** — 24-second fast-fail, the other four site builds on the same run (`landing`, `blog`, `guides`, `api`) all succeeded. The packages site at `web/sites/packages/` doesn't reference `docs/` content anywhere, so the new file can't be affecting the Astro build. Looks like a transient Cloudflare Pages / wrangler hiccup on this site only. A plain re-run of the Deploy packages job should clear it.

Content side, both Reviewer A and Reviewer B converged-approve on the current head. Holding for maintainer action on the two infra-side failures.

---

Issue #2792, opened 2026-05-21
Reload: bare ?reload=true does NOT re-include app/global/*.cfm files — full re-init requires ?reload=true&password=<admin>


bpamiri:
## Summary

Hitting a URL with `?reload=true` does NOT re-evaluate `<cfinclude>` directives in `app/global/functions.cfm` (or any included file like `app/global/datapai/helpers.cfm`). New functions added to those included files are invisible until a full `applicationStop()` runs — which is only triggered by `?reload=true&password=<admin>` (where `<admin>` is the `admin.password` value from `lucee.json`).

Net effect: when developing new helper functions, the obvious developer reflex (`?reload=true`) lies. The page renders without error, but the new helper is still undefined. Looks like Lucee/Wheels just isn't picking it up, but actually the reload silently skipped the re-include step.

## Repro

1. Add a new function to a project's global helpers:

```cfml
// app/global/datapai/helpers.cfm (the file is included from app/global/functions.cfm)
<cfscript>
function newHelperJustAdded() {
    return "I am brand new";
}
</cfscript>
```

2. Hit the dev server with `?reload=true`:

```bash
curl -s "http://localhost:60050/?reload=true" -o /dev/null -w "HTTP %{http_code}\n"
# HTTP 302
```

3. Try to use the new helper in a controller or view:

```cfml
// app/controllers/Smoke.cfc
function index() {
    writeOutput(newHelperJustAdded());  // throws: function 'newHelperJustAdded' is not defined
    abort;
}
```

4. Now hit with `?reload=true&password=titan4dev` (or whatever your Lucee admin password is, from `lucee.json`):

```bash
curl -s "http://localhost:60050/?reload=true&password=titan4dev" -o /dev/null -w "HTTP %{http_code}\n"
# HTTP 302
```

5. Re-try step 3 — the helper now resolves correctly.

The difference: `?reload=true&password=...` triggers `applicationStop()`, forcing the application scope to be torn down and `onApplicationStart` to re-fire, which re-evaluates the `app/global/functions.cfm` include chain. Bare `?reload=true` runs a softer reload that re-loads SOME settings/routes but doesn't re-include global function files.

## Impact

This caused real iteration friction during the DataPAI Phase 0 build. Every time we added a new helper function and tried to test it, the first `?reload=true` request would appear successful but the function was missing. Spending several minutes debugging "why isn't my helper working" before remembering to use the password form is a common pattern.

[...]

Issue #2791, opened 2026-05-21
Routing: inside .namespace("foo"), redundant to="foo/controller##action" produces opaque Foocontroller lookup; should reject or strip the prefix


bpamiri:
## Summary

Inside a `.namespace("foo")` block, the `to=` argument of a route gets concatenated with the namespace name in a confusing way: writing the (intuitive but redundant) `to="foo/controller##action"` produces a lookup for a controller class literally named `Foocontroller` instead of `Controller` under the `foo` package. The correct form is just `to="controller##action"` — the namespace adds the package prefix automatically — but the failure mode when you get it wrong is opaque.

## Repro

```cfml
// config/routes.cfm — WRONG (matches the namespace prefix redundantly)
mapper()
    .namespace("datapai")
        .get(name="datapaiDashboard", pattern="dashboard", to="datapai/dashboard##index")
    .end()
.end();
```

Then `GET /datapai/dashboard` produces:

```
Wheels.ViewNotFound — Could not find the view page for the `index` action in the `datapai.Datapaidashboard` controller.

Looked for the view at: app/views/datapai/datapaidashboard/index.cfm
```

The error is misleading. Wheels concatenated the namespace (`datapai`) with the (already-prefixed) controller path (`datapai/dashboard`) to form a class name `Datapaidashboard` — lowercase concatenation with the first letter capitalized. The user, having written `datapai/dashboard` expecting an `app/controllers/datapai/Dashboard.cfc` file, can't find the source of the bizarre `Datapaidashboard` name in the error.

The fix is to drop the redundant prefix:

```cfml
// config/routes.cfm — CORRECT
mapper()
    .namespace("datapai")
        .get(name="datapaiDashboard", pattern="dashboard", to="dashboard##index")
    .end()
.end();
```

Now Wheels routes `/datapai/dashboard` → `app/controllers/datapai/Dashboard.cfc#index` as expected.

## Why this happens

In `vendor/wheels/mapper/scoping.cfc:42-49`, the `scope()` function combines the parent scope's `path` and `package` with the current scope's, which is correct. But when a route is later registered with `to="datapai/dashboard##index"`, the slash separator in the `to=` string is interpreted as a controller-path component and concatenated with the namespace's package, yielding `datapai/datapai/dashboard` → flattened to `Datapaidashboard` for the class name and `datapai/datapaidashboard` for the view directory.

There's no obvious place in the codebase where this concatenation is documented or guarded.

## Impact

[...]

Issue #2790, opened 2026-05-21
WheelsTest: custom global helpers added via include don't auto-bind into spec scope (getMetaData misses them)


bpamiri:
## Summary

The `WheelsTest` base class auto-binds Wheels globals from `application.wo` into the spec's `variables` and `this` scopes (so tests can call `model(...)`, `urlFor(...)`, etc. bare). The mechanism uses `getMetaData(application.wo).functions` to enumerate methods.

**That enumeration misses user-defined global helpers added via `include` from `app/global/functions.cfm`.** The functions ARE present on the `application.wo` object — `structKeyExists(application.wo, "myHelper")` returns true and `application.wo.myHelper(...)` works at runtime — but `getMetaData(application.wo).functions` doesn't list them because the metadata system only reports methods defined directly on the CFC, not symbols injected via `include`.

Net effect: every spec that exercises a user-defined global helper has to manually rebind in `beforeAll()`:

```cfml
variables.myHelper = application.wo.myHelper;
variables.anotherHelper = application.wo.anotherHelper;
// ... one line per helper, per spec
```

This boilerplate is a maintenance tax. Forget it and the spec fails with a confusing "function not defined" error even though the function clearly works in controllers/views/models.

## Repro

Set up a custom global helper via the `include` convention (this is how every Titan helper is added):

```cfml
// app/global/functions.cfm
<cfscript>
  include "cfml.cfm";
  include "datapai/helpers.cfm";  // includes our custom helpers
</cfscript>
```

```cfml
// app/global/datapai/helpers.cfm
<cfscript>
function can(required string permissionKey) {
    return StructKeyExists(session, "datapai")
        && StructKeyExists(session.datapai, "permissions")
        && StructKeyExists(session.datapai.permissions, arguments.permissionKey);
}
</cfscript>
```

Then in a controller it works:

```cfml
// app/controllers/Foo.cfc
component extends="Controller" {
    function index() {
        if (can("intranet.reports.view")) { /* ... */ }   // works fine
    }
}
```

But in a spec:

```cfml
// tests/specs/datapai/helpers/CanSpec.cfc
component extends="wheels.WheelsTest" {
    function run() {
        describe("can()", () => {
            it("returns true for a granted permission", () => {
                session.datapai = { permissions: { "intranet.reports.view": true } };
                expect(can("intranet.reports.view")).toBeTrue();
                //     ^^^ throws: "function 'can' not defined"
            });
        });
    }
}
```

Investigation:

```cfml
[...]

Issue #2779, opened 2026-05-21
BrowserTest: default base URL is 8080; should match Lucee server port or be auto-detected


bpamiri:
## Summary

\`wheels.wheelstest.BrowserTest\`'s default \`\$resolveBaseUrl()\` returns \`http://localhost:8080\`. Most CFWheels apps run on a port other than 8080 (Titan runs on 60050; new \`wheels new\` scaffolds default to 60080 IIRC). The env var \`WHEELS_BROWSER_TEST_BASE_URL\` is read by the JVM at server start, so setting it on the CLI doesn't help when the server is already running.

## Impact

Every BrowserTest spec needs a manual workaround like this:

\`\`\`cfml
var rawBase = this.browser.getBaseUrl();
var baseUrl = (rawBase == \"http://localhost:8080\") ? \"http://localhost:60050\" : rawBase;
this.browser.visitUrl(baseUrl & \"/\");
\`\`\`

That's friction on every spec and a footgun for newcomers (relative \`.visit(\"/some/path\")\` calls silently hit the wrong port).

## Suggested fixes (any of)

1. **Auto-detect:** read the running Lucee server's port from \`server.json\` / \`lucee.json\` / a runtime introspection and use that as the default.
2. **CLI-time override:** \`wheels test\` could accept \`--base-url=http://localhost:60050\` and inject it into the spec runner without requiring a server restart.
3. **Wheels config:** read \`get(\"baseUrl\")\` (or a dedicated \`get(\"testBaseUrl\")\`) from the app's config and use that instead of hardcoded 8080.
4. **Doc fix at minimum:** document the workaround prominently and explain why \`WHEELS_BROWSER_TEST_BASE_URL\` set at CLI time doesn't take effect when the server is already running.

## Repo / version

- CLI: \`wheels 4.0.1\`
- Wheels Core: \`4.0.0-SNAPSHOT+1779\`
- Files: \`vendor/wheels/wheelstest/BrowserTest.cfc\`, \`vendor/wheels/wheelstest/BrowserClient.cfc\`

## Where found

Surfaced while dogfooding BrowserTest in Titan (port 60050) for the DataPAI portal foundation.

🤖 Found while executing a Claude-Code-driven implementation plan

PR #2796, opened 2026-05-22
fix(test): keep test-local.sh from silently dying on missing ~/.lucli/express

bpamiri:
## Summary

- `tools/test-local.sh` silently aborts with `EXIT=1` (no `/tmp/wheels-test-server.log` produced) on every install whose `~/.lucli/express/` directory never existed — i.e. every install since the lucli→wheels rebrand window closed.
- Root cause is the interaction between `set -euo pipefail` (line 26) and `find ~/.wheels/express ~/.lucli/express -path "*/lib/ext" -type d 2>/dev/null | head -1` (line 81). `find` exits non-zero whenever any path argument is missing; `2>/dev/null` hides the stderr but not the exit status. `pipefail` propagates it through `head -1`, the assignment trips `set -e`, the cleanup trap fires with nothing to clean up, and the user sees the script just stop after "Starting Wheels CLI server on port 8080…".
- Drop the now-dead `~/.lucli/express` fallback (rename landed in 3.0; recent CLI releases extract Lucee Express to `~/.wheels/express/` only) and add `|| true` for defense in depth so a truly fresh install (before `wheels start` has ever run) leaves `LUCEE_LIB` empty and the downstream `[ -n "$LUCEE_LIB" ]` guard skips the JDBC pre-install cleanly.

## Test plan

- [x] Reproduced the silent fail locally on a system with only `~/.wheels/express/` and no `~/.lucli/express/` (post-rename setup): script exits 1 after "Starting Wheels CLI server on port 8080…" with no `/tmp/wheels-test-server.log`.
- [x] Confirmed `find /missing /existing 2>/dev/null | head -1` exits 1 under `set -euo pipefail`, killing the assignment via `set -e`.
- [x] After fix: `bash tools/test-local.sh wheels.tests.specs.wheelstest` runs end-to-end — server starts, `/tmp/wheels-test-server.log` produced (~69 KiB), 137 specs across 38 suites pass in ~17s.
- [ ] CI green across the compat matrix.

## Follow-up (not in this PR)

While verifying, noticed a separate pre-existing bug: even after the tests pass, the script exits 1 with `Tests returned HTTP 000000` because the `--write-out "%{http_code}"` capture (lines 129–132) appears to come back empty (likely a curl-output / response-handling quirk against the long-running test endpoint). This was previously masked by the silent-fail blocking the script from ever reaching the HTTP check. Out of scope here; will look at separately.


bpamiri:
Resolved the `CHANGELOG.md` merge conflict against `develop` so this PR can land cleanly.

**Conflict**

Both branches added new entries to the `[Unreleased]` / `### Fixed` list. The conflict was structural (overlapping appends to the same list region), not semantic — no overlapping content.

**Resolution**

- Kept the PR's `tools/test-local.sh` entry at the top of the list (newest-first ordering — this PR is #2796, develop's newest was #2795).
- Appended `(#2796)` to the entry to match the recent-entry suffix convention.
- Preserved every new entry from `develop` (#2791, #2790, model SELECT `$quoteIdentifier`, #2786, #2779, #2792) and every shared entry from the merge base (BrowserTest NotWired, Linux `.deb`/`.rpm`, `onError` guard).

**Commit**

`chore: merge develop into PR branch to resolve CHANGELOG conflict` — DCO sign-off present, no code changes outside `CHANGELOG.md`. New head SHA `c5efa62`. The previous PR commit `da93da1f5` is preserved as a merge parent so Reviewer A's and Reviewer B's approve marker from the prior SHA is still in the history.

Reviewer A will re-run on the new SHA. Once it and Reviewer B converge again, this is ready to merge.

PR #2802, opened 2026-05-23
feat(migrator): t.references() columnNames alias + underscore suffix flag (#2781)

### LINKED ISSUE #2781: Migrator: t.references() API quirks — 'referenceNames' (not 'columnNames'), and produces '<x>id' without underscore
## Summary

\`\`\`cfml

// Works:
\`\`\`

## Impact

## Suggested fixes

## Repro

\`\`\`cfml
t.timestamps();
t.create();
\`\`\`

## Repo / version

## Where found

bpamiri:
## Summary

`t.references()` has shipped since the 2017 dbmigrate merge with two quirks: parameter is `referenceNames` (every sibling helper uses `columnNames` via `\$combineArguments`), and the generated column suffix is `<x>id` (breaks Wheels `belongsTo` defaults which expect `<x>_id`). This PR resolves both without changing default behavior for existing apps.

- `t.references()` now accepts `columnNames` as an alias for `referenceNames`. Both keep working; `columnNames` is the modern form.
- New setting `useUnderscoreReferenceColumns` (boolean). Framework default `false` so existing schemas are unaffected. `wheels new` template pre-wires `set(useUnderscoreReferenceColumns=true)` so new apps get \`<x>_id\` / \`<x>_type\` out of the box.
- Conventions documented in new \`vendor/wheels/migrator/CLAUDE.md\`, plus anti-pattern #15 in top-level CLAUDE.md and an expanded doc-block on \`\$combineArguments\` positioning it as the canonical alias helper.

## Tests

5 new specs in \`vendor/wheels/tests/specs/migrator/referencesSpec.cfc\` covering alias acceptance, both flag states, and the polymorphic \`*_id\` / \`*_type\` pair. Full suite: **3761 pass / 0 fail** (was 3756 pre-PR).

## Scope (1 of 4)

This PR fixes the user-visible \`t.references()\` quirks at the \`TableDefinition\` layer. Three follow-ups already scoped:

- **PR2** — Migration.cfc command-version consistency: \`addReference\`, \`dropReference\`, \`removeColumn\`, \`addColumn\`, \`changeColumn\`, \`addForeignKey\` get \`columnName\`/\`columnNames\` aliases and respect the suffix flag at their own concat sites. Closes the gap noted in \`vendor/wheels/migrator/CLAUDE.md\` anti-pattern #2.
- **PR3** — \`wheels upgrade check\` gains a \"Recommended Improvements\" advisory tier alongside breaking changes. Pure scanner refactor; no advisories defined yet.
- **PR4** — Specific advisory entries: surface the underscore-references opt-in to existing apps that haven't enabled it, plus a mixed-convention warning for apps that flipped the flag after applying legacy migrations.

## Test plan

- [x] \`bash tools/test-local.sh migrator\` — 223 pass
- [x] \`bash tools/test-local.sh\` (full suite) — 3761 pass / 0 fail
- [ ] Cross-engine matrix (will run in CI)
- [ ] Smoke: generate a new app via \`wheels new\` and verify the scaffold template's \`t.references(columnNames="vacation")\` migrates cleanly under flag=true
- [ ] Verify legacy app with applied migrations is unaffected (flag=false default preserves \`<x>id\` everywhere)

Closes #2781


bpamiri:
Verified the bot's round-1 fixes locally:

- `bash tools/test-local.sh migrator` → **224 pass** (was 223; the new `still accepts the legacy referenceNames parameter` regression test brings it to 224)
- `bash tools/test-local.sh` (full suite) → **3762 pass / 0 fail**

All 6 reviewer findings addressed in `bea48ce1`. The `Migration.cfc` patches (lines 258, 278) match the same `$get("useUnderscoreReferenceColumns")` pattern used in `TableDefinition.cfc::references()` so behavior is uniform across all four reference-column concat sites. The new app template default of `true` is now safe — no silent column-name mismatch from `removeColumn(referenceName=...)` or `addReference()`.

Heads-up for the merge order: #2804 (PR2 — Migration.cfc command alias additions) was based on PR1 pre-fix and includes redundant suffix-flag changes that PR1 now ships. Once #2802 merges, #2804 will need a rebase to drop the duplicated portions; its remaining net diff is the `columnName`/`columnNames` arg aliases on `addColumn`, `changeColumn`, `addReference`, `dropReference`, `addForeignKey`, plus the `removeColumn` plural alias.

PR #2804, opened 2026-05-23
feat(migrator): Migration.cfc command consistency sweep (follow-up to #2781)

### LINKED ISSUE #2781: Migrator: t.references() API quirks — 'referenceNames' (not 'columnNames'), and produces '<x>id' without underscore

bpamiri:
## Summary

PR1 ([#2802](https://github.com/wheels-dev/wheels/pull/2802)) fixed \`t.references()\` at the \`TableDefinition\` layer. This PR brings the matching command-version helpers in \`Migration.cfc\` into the same convention.

**Argument-name consistency** — every column-accepting helper accepts the plural/singular alias shape via \`\$combineArguments\`:

| Function | Legacy | Modern alias(es) |
|---|---|---|
| \`addColumn\` / \`changeColumn\` | \`columnName\` | \`columnNames\` |
| \`removeColumn\` | \`columnName\` | \`columnNames\` |
| \`addReference\` / \`dropReference\` | \`referenceName\` | \`columnName\` / \`columnNames\` |
| \`addForeignKey\` | \`column\` | \`columnName\` |

**Suffix-flag correctness** — two hard-coded \`& \"id\"\` concats now route through \`useUnderscoreReferenceColumns\`, matching \`references()\` behavior from PR1:

- \`removeColumn\` line 258 — when \`referenceName=\` is used, the dropped column name follows the flag
- \`addReference\` line 278 — when the FK column is built, follows the flag

Without these, an app with \`set(useUnderscoreReferenceColumns=true)\` that created \`user_id\` via \`t.references()\` would silently fail to drop or constrain it via the command-version helpers.

## Tests

\`vendor/wheels/tests/specs/migrator/migrationCommandsSpec.cfc\` — 4 new integration tests:
- \`removeColumn\` drops \`user_id\` when flag is true
- \`removeColumn\` drops \`userid\` when flag is false (legacy preserved)
- \`addColumn\` with \`columnNames\` alias actually adds the column
- \`removeColumn\` with \`columnNames\` alias actually drops the column

Full suite: **3765 pass / 0 fail** (was 3761 pre-PR2).

## Note for reviewers

Base is [#2802](https://github.com/wheels-dev/wheels/pull/2802) — once PR1 merges to develop, I'll retarget this PR to develop and the diff narrows to just Migration.cfc + spec.

## Scope (2 of 4 PRs from the #2781 audit)

- ✅ [#2802](https://github.com/wheels-dev/wheels/pull/2802) — t.references() (closes #2781)
- 🟢 This PR — Migration.cfc command-version consistency
- 🔜 PR3 — \`wheels upgrade\` advisory tier (scanner refactor)
- 🔜 PR4 — specific advisories for the underscore-references opt-in

## Test plan

- [x] \`bash tools/test-local.sh migrator\` — 227 pass
- [x] \`bash tools/test-local.sh\` (full suite) — 3765 pass / 0 fail
- [ ] Cross-engine matrix (CI)

### DIFF
--- a/vendor/wheels/migrator/CLAUDE.md
+++ b/vendor/wheels/migrator/CLAUDE.md
@@ -17,7 +17,18 @@ public any function string(string columnNames, any limit, string default, boolea
 
 Callers can pass either `t.string(columnNames = "a,b,c")` or `t.string(columnName = "a")` — both resolve to `arguments.columnNames` for the function body. Drop the `required` keyword from the parameter declaration; `$combineArguments(required=true)` enforces it at runtime.
 
-**`references()` carries a back-compat exception.** Its legacy parameter is `referenceNames`, with `columnNames` accepted as a synonym since the [#2781](https://github.com/wheels-dev/wheels/issues/2781) fix:
+**`references()` and its command-version siblings carry back-compat aliases.** The legacy parameter names predate the `$combineArguments` convention; the modern ones are accepted as synonyms via the same helper. Each of these accepts the modern form going forward:
+
+| Function | Legacy param | Modern alias(es) |
+|---|---|---|
+| `TableDefinition::references()` | `referenceNames` | `columnNames` |
+| `Migration::addReference()` | `referenceName` | `columnName`, `columnNames` |
+| `Migration::dropReference()` | `referenceName` | `columnName`, `columnNames` |
+| `Migration::addColumn()` / `changeColumn()` | `columnName` | `columnNames` |
+| `Migration::removeColumn()` | `columnName` | `columnNames` |
+| `Migration::addForeignKey()` | `column` | `columnName` |
+
+Example (the `references()` form, [#2781](https://github.com/wheels-dev/wheels/issues/2781)):
 
 ```cfm
 $combineArguments(args = arguments, combine = "referenceNames,columnNames", required = true);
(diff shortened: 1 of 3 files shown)


bpamiri:
Rebased onto develop after #2802 merged. Force-pushed to \`fd18ae01\`.

## What changed in the rebase

- **Migration.cfc \`removeColumn\` and \`addReference\`**: the suffix-flag plumbing originally proposed in this PR is already on develop (PR1's review iterations added it directly). PR2's net contribution to those functions is now just the alias-parameter additions — \`columnName\` / \`columnNames\` as synonyms for \`referenceName\`. Same \`local.idSuffix\` naming convention as PR1.
- **\`migrationCommandsSpec.cfc\`**: removed the two \`removeColumn(referenceName=)\` suffix-routing tests since develop now has equivalents in \`migrationSpec.cfc::Tests removeColumn\` (added during PR1's R5 review). Kept the two \`columnNames\` plural-alias integration tests, which are PR2's unique value-add.
- **\`vendor/wheels/migrator/CLAUDE.md\`**: anti-pattern #1 updated to past-tense ("Both helpers now accept the modern aliases") since this PR closes the open consistency gap it previously flagged.

## Verification

\`bash tools/test-local.sh migrator\` → 228 pass / 0 fail (226 develop baseline + 2 PR2 alias tests). No regressions.

Convergence loop will rerun on the new SHA.

bpamiri:
Heads-up — the regression Reviewer B caught is already fixed in HEAD \`05cc4312\` (pushed before B's reviewer-A-response landed). The bot reviewers should re-evaluate on the latest SHA.

## On Reviewer A's prescription vs the fix in HEAD

A's literal prescription was:
- \`changeColumn\` → \`required=true\`
- \`addColumn\` → restore \`string columnName = \"\"\` default

There's a logic gap with that combination: \`\$combineArguments(required=true)\` checks both \`StructKeyExists\` AND \`Len() > 0\`. So if \`addColumn\` injects \`columnName=\"\"\` via the restored default and delegates to \`changeColumn(required=true)\`, the empty length check throws — breaking the \`addColumn(columnType=\"reference\", referenceName=\"user\")\` delegation that was valid pre-PR.

## What's in HEAD

Both \`addColumn\` and \`changeColumn\` use:
\`\`\`cfm
\$combineArguments(args = arguments, combine = \"columnName,columnNames\", required = (arguments.columnType != \"reference\"));
\`\`\`

- Non-reference column types: \`required=true\` (matches framework convention, restores missing-arg enforcement)
- Reference column types: \`required=false\` (allows the reference-type call pattern where \`columnName\` is unused)

Three regression-guard tests added in \`migrationCommandsSpec.cfc\` (5634abd9 → 05cc4312 range):
- \`addColumn\` without column args → throws ✓
- \`changeColumn\` without column args → throws ✓  
- \`addColumn(columnType=\"reference\", referenceName=\"user\")\` → does NOT throw ✓

The third test pins the reference-type exemption — a future \"just use \`required=true\` unconditionally\" simplification would fail this test.

## Verification

\`bash tools/test-local.sh migrator\` → 234 pass / 0 fail.
\`bash tools/test-local.sh\` (full) → 3769 pass / 0 fail.

Inline comments: zero.

PR #2806, opened 2026-05-23
feat(cli): underscore-references opt-in advisories in wheels upgrade check

### LINKED ISSUE #2781: Migrator: t.references() API quirks — 'referenceNames' (not 'columnNames'), and produces '<x>id' without underscore

bpamiri:
## Summary

Adds two advisory entries to the upgrade scanner, populating the \"Recommended Improvements\" section introduced in [#2805](https://github.com/wheels-dev/wheels/pull/2805) with the first concrete opt-in convention from the #2781 work.

### 1. \`t.references()\` opt-in suggestion

Fires when any migration file under \`app/migrator/migrations/\` uses \`t.references(\`. Suggests opting into \`<name>_id\` naming via \`set(useUnderscoreReferenceColumns=true)\` to match Wheels model \`belongsTo\` defaults. Explicitly notes that **already-applied migrations are unaffected** — only new migrations adopt the new suffix.

### 2. Mixed-convention warning

Fires when \`config/\` contains \`set(useUnderscoreReferenceColumns=true)\`. Reminds the user that if this flag was flipped after some migrations were already applied, the database still has \`<name>id\` columns alongside any new \`<name>_id\` columns. Informational — read once, ignore if not applicable.

Both use \`severity: \"advisory\"\` — surface in the cyan \"Recommended Improvements\" section, run regardless of major-version bump, never gate CI exit code.

## Tests

\`cli/lucli/tests/specs/commands/UpgradeAdvisorySpec.cfc\` — 5 source-level specs verifying both advisory entries are present and reference the correct patterns, paths, and fix-message phrasing.

CLI suite: **591 pass / 5 fail** (same 5 pre-existing failures unrelated to upgrade).

## Note for reviewers

Base is [#2805](https://github.com/wheels-dev/wheels/pull/2805) — once that merges to develop, this PR's diff narrows to just the advisory entries.

## Scope (4 of 4 PRs from the #2781 audit)

- ✅ [#2802](https://github.com/wheels-dev/wheels/pull/2802) — t.references() at TableDefinition (closes #2781)
- ✅ [#2804](https://github.com/wheels-dev/wheels/pull/2804) — Migration.cfc command consistency
- ✅ [#2805](https://github.com/wheels-dev/wheels/pull/2805) — wheels upgrade scanner advisory tier
- 🟢 This PR — specific advisory entries (closes the loop)

## Test plan

- [x] \`bash tools/test-cli-local.sh\` — 5 new advisory specs pass; pre-existing failures unchanged
- [ ] End-to-end smoke: actually run \`wheels upgrade check --to=4.0.1\` against a real app with t.references() migrations and verify advisory appears
- [ ] Cross-engine matrix (CI)


bpamiri:
Replaced by #2807 — this PR was auto-closed when its base branch `claude/upgrade-advisory-tier-pr3` was deleted on PR3 merge, and GitHub's reopen API rejected it ('Could not open the pull request'). Same commit content, rebased onto develop, opened fresh.

Issue #2808, opened 2026-05-24
wheels upgrade: underscore-flag pre-check scope-asymmetry with advisory #2


bpamiri:
Spun off from [PR #2807](https://github.com/wheels-dev/wheels/pull/2807) (Reviewer A round-3 non-blocking nit).

## Problem

In \`cli/lucli/Module.cfc::runUpgradeCheck()\`, the advisory-1 suppression guard reads only \`config/settings.cfm\`:

\`\`\`cfm
var settingsFile = variables.projectRoot & \"/config/settings.cfm\";
if (fileExists(settingsFile)) {
    underscoreFlagAlreadySet = reFindNoCase(
        \"useUnderscoreReferenceColumns\s*=\s*true\",
        stripCfmlComments(fileRead(settingsFile))
    ) > 0;
}
\`\`\`

But advisory #2 scans all of \`config/\` for the same pattern:

\`\`\`cfm
arrayAppend(checks, {
    description: \"useUnderscoreReferenceColumns=true is set — confirm legacy migrations don't conflict\",
    severity: \"advisory\",
    pattern: \"useUnderscoreReferenceColumns\s*=\s*true\",
    scanDir: \"config\",
    extensions: \"cfm,cfc\",
    ...
});
\`\`\`

## Impact

If a user sets the flag in an environment-specific override file (e.g., \`config/production/settings.cfm\`, \`config/development/settings.cfm\`), the pre-check on the root \`settings.cfm\` will return false, advisory #1 will fire (\"opt into the flag!\"), AND advisory #2 will also fire (\"the flag is set!\"). Same contradiction the round-1 fix was trying to prevent.

Low impact because:
1. Per-environment flag overrides for this setting are unusual
2. Both advisories are \`severity: \"advisory\"\` — they never gate CI

## Suggested fix

Expand the pre-check to scan all of \`config/\` recursively (matching advisory #2's scope), not just the root \`settings.cfm\`. Existing helper pattern at \`Module.cfc:3995\` (the grep loop) shows how to do recursive scanning.

Alternatively: simplify by letting the bucketing handle the contradiction — e.g., if advisory #2 fires, suppress advisory #1 even if its guard didn't catch the flag.

Good first issue for someone wanting to touch the CLI without a deep dive.

PR #2810, opened 2026-05-24
fix(model): bypass nested cftransaction when migrator owns the outer (#2789)

### LINKED ISSUE #2789: Migrator: model().create()/.update()/.deleteAll() inside up()/down() silently rolls back the row (nested-transaction conflict with the migrator's outer transaction)
## Summary

Calling `model().create()`, `.update()`, or `.deleteAll()` inside a migration's `up()` or `down()` causes the inserted/updated row to be **silently rolled back** while related rows inserted in the same closure (via other code paths, e.g., raw `queryExecute`, `addRecord()`, or operations on a different model) persist as orphans referencing the rolled-back row's nonexistent ID.

Root cause: the Wheels Migrator (`vendor/wheels/Migrator.cfc:98-118`) already wraps each `up()`/`down()` call in its own `transaction {}` block. Wheels Model's default `transaction = "commit"` mode opens a NESTED transaction that immediately commits — at the JDBC level, on SQL Server this nested commit appears to mark a savepoint but doesn't actually release the row, and then the outer transaction's eventual `transaction action="commit"` doesn't see the nested-committed row to keep. The net effect for the user is that `Model.create()` "succeeds" (returns the object, validations pass, no errors) but no row appears in the database. Related rows inserted via paths that don't open a nested transaction (raw `queryExecute`, `addRecord`) commit normally with the outer transaction.

## Repro

```cfml
// app/migrator/migrations/20260521160000_seed_datapai_system_roles.cfc
component extends="wheels.migrator.Migration" hint="seeds five system roles with their permission assignments" {

    function up() {
        // Inside the migrator-provided outer transaction.

        var Role = model("DataPaiRole");
        var RolePerm = model("DataPaiRolePermission");

        // This call APPEARS to succeed — no errors, no validation failures.
        var role = Role.create(name="DataPaiSuperAdmin", description="bootstrap role", isSystem=true);

        // The role's auto-increment ID is populated (e.g., role.id == 115).
        // We can read it here.

        // Insert role-permission assignments using that ID via raw SQL.
        queryExecute(
            "INSERT INTO datapai_role_permissions (datapai_role_id, datapai_permission_id, createdAt, updatedAt) VALUES (?, ?, GETDATE(), GETDATE())",
            [
                {value: role.id, cfsqltype: "cf_sql_integer"},
                {value: 1, cfsqltype: "cf_sql_integer"}
            ],
            {datasource: application.wheels.dataSourceName}
        );
    }
}
```

After `wheels migrate latest`:

```sql
SELECT * FROM datapai_roles WHERE name = 'DataPaiSuperAdmin';
-- 0 rows. The create() was silently rolled back.

SELECT * FROM datapai_role_permissions WHERE datapai_role_id = 115;
-- 1 row. The raw INSERT committed via the outer transaction.
-- It now references role ID 115 which does NOT exist — an orphan FK.
```

The IDENTITY value 115 IS consumed (the next inserted row gets 116), confirming the INSERT ran and the row was alive at some point during the migration before being rolled back.

## Impact

This is the most insidious failure mode we hit during Phase 0 of a real production application build. We didn't notice it during initial run of the migration — `wheels migrate latest` printed no errors, the migration was marked `migrated` in `wheels_migrator_versions`, and the migration's `up()` body completed without throwing. The bug only surfaced when downstream code that depended on those roles (a session-loader query joining `datapai_roles` against `datapai_user_roles`) returned zero results.

Every CFWheels engineer doing seed-style migrations is at risk. The plan documentation pattern (seed via Model.create() inside up()) is intuitive and the code APPEARS to work in isolation, so this lands silently and shows up later as data-integrity bugs.

We worked around it by rewriting the seed migration to use raw `queryExecute()` + the migrator's built-in `addRecord()` / `updateRecord()` / `removeRecord()` helpers exclusively. See [paiindustries/titan PR #3337 commit `ff15bb2ae`](https://github.com/paiindustries/titan/pull/3337/commits/ff15bb2ae) for the working pattern.

## Suggested fixes (any of)

1. **Detect the outer transaction and short-circuit Model's transaction handling.** When `model("X").create()` is invoked inside a context where `transactionStatus()` (or equivalent CFML reflection) shows an active transaction, skip the model's `transaction="commit"` wrapper and let the outer transaction handle commit/rollback. The model's transaction behavior is a feature for ad-hoc app-code use, not for batch contexts like migrations.

2. **Print a warning when `Model.create()`/`.update()`/`.deleteAll()` is called from inside a `wheels.migrator.Migration` subclass.** Use stack-frame inspection (`callStackGet()`) to detect the migrator context and emit a one-line warning recommending `addRecord()` or raw `queryExecute()` instead.

3. **Document the limitation prominently** in `guides.wheels.dev/v4-0-0/migrations/` — at the top of the "Writing migrations" page, with a comparison table:

   | Operation       | OK inside migration? | Alternative                                          |
   |-----------------|----------------------|------------------------------------------------------|
   | `model().create()`   | ❌ silently rolls back | `addRecord(table=..., col=val, ...)`                  |
   | `model().update()`   | ❌                    | `updateRecord(table=..., where=..., col=val, ...)`    |
   | `model().deleteAll()` | ❌                    | `removeRecord(table=..., where=...)`                   |
   | `queryExecute()`     | ✅                    | —                                                    |
   | `addRecord()` / `updateRecord()` / `removeRecord()` | ✅ | — |

4. **Long-term: align Model's transaction mode with CFML's nested-transaction semantics on each adapter.** SQL Server, MySQL, Oracle, and PostgreSQL all handle nested transactions differently. The current Wheels Model assumption that `transaction="commit"` is always safe doesn't hold inside an explicit outer transaction.

## Repo / version

- Wheels Core: `4.0.0-SNAPSHOT+1779`
- File: `vendor/wheels/Migrator.cfc:98-118` (outer transaction wrapper)
- Adapter: MSSQL (MicrosoftSQLServerMigrator.cfc) — but I'd expect the same on other adapters whose JDBC drivers don't implement savepoint-based nested transactions.

## Where found

Surfaced building the DataPAI portal seed migrations in Titan (Phase 0). The first seed migration (`20260521150000_seed_datapai_permissions.cfc`) hit it for the permissions table, and the second (`20260521160000_seed_datapai_system_roles.cfc`) hit it harder because it had a model-create-then-FK-insert pattern that left orphan join rows. Both migrations were rewritten to use raw SQL + migrator helpers; the rewrites are committed and tagged at `datapai-phase-0-complete`.

bpamiri:
## Summary

Fixes [#2789](https://github.com/wheels-dev/wheels/issues/2789) — `model().create()` / `.update()` / `.deleteAll()` inside a migration's `up()` or `down()` silently rolled back the row on MSSQL.

The Migrator already wraps every `up()` / `down()` invocation in its own outer `cftransaction`. Wheels Model's default `transaction="commit"` opened a nested `cftransaction` whose JDBC nested-transaction semantics differ per adapter — most acutely on MSSQL, where the inner commit didn't release the row and the eventual outer commit silently dropped it. The user saw `Role.create()` succeed (auto-increment fired, no errors), and a follow-up raw `queryExecute` insert of an FK row persist normally — but the role row was rolled back, leaving an orphan FK pointing at a nonexistent ID.

## Fix

Two-half change:

1. **`Migrator.cfc`** sets `request.$wheelsTransactionWrapper = true` around each `up()` / `down()` / `migrateIndividual()` invocation, and explicitly clears it in both the success and error paths. Request-scope so it never persists past the request; never leaks across migrations because each iteration re-sets/clears.

2. **`Model::invokeWithTransaction()`** reads the signal first. When set, the call switches to `"alreadyopen"` mode, skipping the nested `cftransaction` entirely. The existing per-connection tracking in `request.wheels.transactions` continues to work as a fallback for Wheels-internal nested calls — the new check is strictly additive.

Engine-agnostic: works the same way on MSSQL, MySQL, PostgreSQL, SQLite, Oracle, and H2 because the fix eliminates the nested transaction rather than working around per-engine JDBC behavior. Approach (1) from the issue body — detect outer transaction and short-circuit — picked over (2) warn-on-detection because it actually fixes the data-loss bug instead of merely surfacing it.

The signal name `$wheelsTransactionWrapper` is intentionally generic — app code that wraps work in its own `cftransaction` can opt in the same way:

```cfm
transaction {
    request.$wheelsTransactionWrapper = true;
    try {
        model("User").create(...);   // skips its own nested transaction
        model("Order").create(...);
    } finally {
        StructDelete(request, "$wheelsTransactionWrapper");
    }
}
```

(Docs follow-up will land in a separate PR via `bot-update-docs.yml`.)

## Tests

- **`vendor/wheels/tests/specs/model/OuterTransactionSignalSpec.cfc`** — four specs exercising `invokeWithTransaction`'s response to the flag for `create` / `save` / `deleteAll`, plus a control test proving rollback still works when the flag is unset. The control is critical — it establishes that the rollback path actually fires in this environment, so the inverse assertion ("with the flag set, rollback is bypassed and the row persists") is meaningful.

- **`vendor/wheels/tests/specs/migrator/MigratorOuterTransactionSpec.cfc`** — four specs running an end-to-end migration whose `up()` calls `model().create()`, asserting the row persists; that the flag is set during `up()`; and that it's cleared after both `up()` and `down()` return. Uses a separate test-migrations directory (`_assets/migrator/migrations_2789/`) so the synthetic `001_seed_tag_via_model_create.cfc` doesn't collide with the reconciliation spec's fixtures.

Local: `bash tools/test-local.sh` — 3780 passed (Lucee 7 + SQLite). Cross-engine matrix runs in CI.

## Test plan

- [x] Failing-then-passing specs (TDD)
- [x] `bash tools/test-local.sh model` — green
- [x] `bash tools/test-local.sh migrator` — green
- [x] `bash tools/test-local.sh` (full) — green
- [ ] Cross-engine matrix passes (CI)
- [ ] MSSQL reproduces issue #2789's exact scenario (CI's MSSQL leg)
- [ ] No regressions in the existing transaction specs

## Pull request checklist

- [x] Tests — failing-then-passing specs at `vendor/wheels/tests/specs/model/OuterTransactionSignalSpec.cfc` and `vendor/wheels/tests/specs/migrator/MigratorOuterTransactionSpec.cfc`
- [ ] Framework Docs — handled separately by `bot-update-docs.yml`
- [ ] AI Reference Docs — handled separately by `bot-update-docs.yml`
- [ ] CLAUDE.md — handled separately by `bot-update-docs.yml`
- [x] CHANGELOG.md
- [x] Test runner passes locally (Lucee 7 + SQLite)

Fixes #2789


bpamiri:
## Wheels Bot — Address Review (round 2)

Applied the consensus finding Reviewer B caught in round 1 that the round-1 address-review missed:

- **Multi-line comment blocks in test files trimmed to one-liners.** CLAUDE.md: "Never write multi-line comment blocks — one short line max." Reviewer A flagged the identical pattern in production code in round 1 and the address-review trimmed `Migrator.cfc` / `transactions.cfc`, but the same violation persisted in the new test files. Reviewer B's round-1 review caught the inconsistency; round-2 fixes it.
  - `vendor/wheels/tests/specs/model/OuterTransactionSignalSpec.cfc:32-39, 41-60, 62-82, 84-97` — trimmed the 9-line block in `treats invokeWithTransaction as 'alreadyopen'`, the 2-line block in `control:`, the 3-line block in the `save()` case, and the 3-line block in the `deleteAll` case to one-liners each, anchored on issue #2789.
  - `vendor/wheels/tests/_assets/migrator/migrations_2789/001_seed_tag_via_model_create.cfc:14-17` — trimmed the 2-line block in `down()` to one line.

Also merged `origin/develop` to resolve the dirty merge state (PR #2808 / commit `c21efb8cd` had added an adjacent `### Fixed` entry to `CHANGELOG.md`; kept both entries).

**Skipped (intentionally, per consensus scope):**

- **`migrateIndividual()` commit-after-rollback** — Reviewer A and Reviewer B both labelled this pre-existing in round 1 and explicitly recommended a follow-up issue, not action on this PR. The new `StructDelete` placement is correct in both paths.
- **`variables.migration` set in `beforeAll()` but referenced only via included helpers** — both reviewers labelled this a nit (Reviewer A's words), not a blocker. The bare-scope assignment is correct (closures inherit `variables` scope on every supported engine); the suggested explanatory one-liner is a future-readability nicety that can land separately.

**Test verification:**

- `bash tools/test-local.sh model` → 843 passed (Lucee 7 + SQLite)
- `bash tools/test-local.sh migrator` → 238 passed (Lucee 7 + SQLite)

Pure comment-trim changes with zero behavior delta. The cross-engine matrix will validate the broader fix on every supported engine × DB.

The new commit will trigger a fresh Reviewer A run on the updated SHA. Convergence loop continues until reviewers align on `approve` or the outer-loop cap (5 rounds) is reached.

bpamiri:
Follow-up to the round-2 address-review: while in the spec, also picked up the two outstanding round-1 nits both reviewers had labelled non-blocking.

- **`variables.migration` explanatory comment** — Reviewer A's nit (B confirmed reasonable, both labelled it not-blocker). Added a one-liner above the bare assignment in `vendor/wheels/tests/specs/migrator/MigratorOuterTransactionSpec.cfc:6` naming `helperFunctions.cfm`'s engine-detection helpers as the consumer. Also tightened the adjacent migrator-init two-liner to one line per CLAUDE.md.

- **`migrateIndividual()` commit-after-rollback** — both reviewers explicitly recommended a dedicated follow-up issue rather than expanding #2810's scope. Filed as #2811 with the full diagnosis from both reviews and a suggested fix pattern. The new `StructDelete` placement from #2810 stays correct in both paths; the follow-up is purely the pre-existing flow-control gap that's now visible in this code path.

Cross-engine matrix validation (MSSQL repro of the original #2789 scenario) still pending on CI for the new SHA `a5c2c7f99`.

Tests: `bash tools/test-local.sh migrator` → 238 passed (Lucee 7 + SQLite).

PR #2817, opened 2026-05-27
fix: green the compat matrix for BoxLang and Adobe CF 2023/2025

bpamiri:
## Summary

Gets the Wheels Compatibility Matrix green before v4.0.2. BoxLang, Adobe CF 2023, and Adobe CF 2025 were all red; Lucee 6/7 were green. Root-caused and fixed each, verified locally across all engines.

### BoxLang (89 failures → 0)
All 89 failures (17 fail / 72 error) traced to **one** root cause: `Global.cfc`'s pseudo-constructor seeded `local.varKey = ""`, which BoxLang materializes as `variables.local`, shadowing the function-local `local` scope of every mixed-in `$`-helper (`Migrator`/`Model` `local.appKey`, etc.) → `KeyNotFoundException: The key [APPKEY] was not found in the struct. Valid keys are ([VARKEY])`. Lucee/Adobe keep `local` reserved to function scope, so they were unaffected. Extracted the UDF-promotion loop into `$promoteIncludedGlobalsToThis()` so its iterator uses a real function-local scope.

### Adobe CF 2023 / 2025
- **Suite-crash (dominant cause of red):** `InvokeMethodSpec` invoked `Public.index()`, rendering the congratulations welcome page straight into the test-runner response buffer. Adobe committed the buffer (HTTP 404 + ~1MB HTML prefix), corrupting the entire suite's JSON output. Wrapped the call in `cfsavecontent`.
- **`RequestId` middleware:** wrote `request.wheels.requestId` through a parameter named `request`, which shadows the request scope on Adobe (anti-pattern #11). Moved the scope write into a helper with no `request` param.
- **`TestClient`:** emitted a `cfhttp` POST with zero `cfhttpparam` tags for an empty body; Adobe rejects that. Always emit a (possibly empty) body param for POST/PUT/PATCH.
- **`ParallelRunner.$collectFailures`:** relied on array-by-reference mutation, but Adobe passes arrays to functions by value, so the aggregated failure list stayed empty. Returns the array now.
- **`$reincludeGlobals`:** re-`include`d a file whose UDFs were already bound to `application.wo`, tripping Adobe's "Routines cannot be declared more than once". Evaluates the file in a throwaway `GlobalIncludeLoader` instance.
- **`RewriteConfigInstallerSpec` (2025 only):** compared file content to a literal string; Adobe 2025's `fileWrite`/`fileRead` round-trip appends a newline. Compares to a normalized post-write baseline now.

## Local verification (full suites)

| Engine | Result |
|--------|--------|
| BoxLang | sqlite 3790/0/0, mysql 3801/0/0 |
| Adobe 2023 | mysql 3801/0/0 (HTTP 200, clean JSON) |
| Adobe 2025 | mysql 3801/0/0 (HTTP 200, clean JSON) |
| Lucee 6 | mysql 3801/0/0 (no regression) |
| Lucee 7 | sqlite 3789/0/0 (no regression) |

## Test plan

- [ ] Compatibility Matrix CI green for lucee6, lucee7, adobe2023, adobe2025, boxlang across all DBs (oracle soft-fail)
- [ ] No regression on the standard test suite


bpamiri:
Thanks for the thorough review. Addressed the actionable nits in `aa8438523`:

- **`GlobalIncludeLoader` exclusion-list fragility (A + B):** switched from the by-name `loadFunctions` exclusion to a snapshot-diff — `variables` keys are captured *before* the include and only newly-appeared UDFs are returned, so a future sibling method can't silently leak. Behavior is unchanged today; verified the global subset (which runs `reloadGlobalsSpec`, the loader's integration path) at **114/0/0 on Lucee 7, BoxLang, and Adobe 2023**.
- **`TestClient` else-branch readability (A):** added the one-line comment clarifying the branch covers both JSON posts and empty-body form posts.

Deferred / noted:
- **Dedicated `GlobalIncludeLoader` unit spec:** leaving as the follow-up you both suggested — the loader is already exercised end-to-end by `reloadGlobalsSpec` (verified on Adobe), and the snapshot-diff now makes own-method exclusion structural rather than by-name.
- **Commit scope `(test)`:** fair point — the original commit touched production files. Not rewriting published history for it; later commits use accurate scopes (`fix(test)` for the cockroachdb test-only change, `refactor`/no-scope here).
- **CHANGELOG:** already added in `0be659982` (B confirmed).

On the **"Wheels Test Results" check:** the matrix's per-engine jobs all pass — that red is the JUnit aggregate counting **BoxLang + Oracle (2 fail / 5 error)**, which are pre-existing Oracle JDBC connection-lifecycle issues (`ORA-17009`, "Connection is closed") in the #2663 soft-fail bucket. These fixes actually improved BoxLang+Oracle from **22f/72e → 2f/5e**; Oracle now fully passes on Lucee 6/7 and Adobe 2023.

PR #2852, opened 2026-06-03
fix(model): preserve database column case for auto-derived properties

bpamiri:
## Problem

Migrating from CFWheels 2.5 → Wheels 3/4, auto-derived model property names come back lowercased: an `isHidden` column surfaces as the property `ishidden`, breaking case-sensitive consumers of serialized model output (`returnAs="structs"`, `renderWith()`, `serializeJSON()`). Reported on Slack; no prior issue. The same code worked in 2.5 on the same Lucee 7 + SQL Server, so it's a framework change — not a Lucee/driver change.

## Root cause

When a model declares no `property()` mappings, `Model.cfc` auto-derives its properties from the database column metadata. Commit `dbfbdda71` ("Oracle compatibility", 2025-07-15) replaced the raw column name with `lCase(local.columns["column_name"][local.i])`, force-lowercasing **every** auto-derived property name on **every** engine. Oracle reports unquoted identifiers in uppercase, so normalizing there made sense; applying it to all adapters regressed case preservation for SQL Server / MySQL / SQLite. Shipped in `v3.0.0` (build +33) through the `v4.0.x` line.

## Fix

Preserve the database's reported column casing by default; lowercase only on adapters whose database folds unquoted identifiers to a non-meaningful uppercase default.

- New `$lowerCaseColumnNames()` capability on `databaseAdapters/Base.cfc` (default `false`), mirroring the existing `$supportsAdvisoryLocks()` pattern.
- `OracleModel` and `H2Model` override it to `true` (both upper-fold unquoted identifiers).
- `Model.cfc` consults it instead of unconditionally calling `lCase()`.

Behavior by engine:

| Engine | Behavior |
|---|---|
| SQL Server, MySQL, SQLite | preserve declared case (`isHidden`) — restores 2.5 behavior |
| PostgreSQL, CockroachDB | DB folds to lowercase; reported name used as-is |
| Oracle, H2 | DB folds to uppercase; lowercased (unchanged from today) |

Models that explicitly declare `property(name="isHidden", column="isHidden")` were always unaffected and remain so.

## Tests

Adds `propertyCasePreservationSpec` and a `c_o_r_e_casepreservation` fixture table with an **undeclared** mixed-case `isHidden` column. The assertion is adapter-aware (preserve-case engines expect `isHidden`; fold engines expect `ishidden`) and uses a **case-sensitive** `ListFind` so the regression can't hide behind case-insensitive lookups.

Verified locally on **Lucee 7** across **SQLite, H2, MySQL, and PostgreSQL** — 0 failures, 0 errors on each (RED confirmed first: the spec fails pre-fix with `actual [0] is not greater than [0]`). Oracle + Adobe 2023/2025 + BoxLang + Lucee 6 run via the compat-matrix on this PR.

## Pre-release stopgap for affected users

`property(name="isHidden", column="isHidden")` on the affected models restores casing today without waiting for the release.


bpamiri:
## Addressed Reviewer A/B (round 2) — `27fddae53`

No inline review threads exist on this PR (the reviews are top-level), so summarizing here.

- **Condensed the inline comment blocks** Reviewer A flagged — `Model.cfc` (8-line `//` block → 1 line) and `propertyCasePreservationSpec.cfc` (the `/* */` + `//` blocks → single lines), per the one-line-comment guidance.
- **Kept `Base.$lowerCaseColumnNames()`'s docstring** — it matches the existing `$supportsAdvisoryLocks()` `/** */` convention, which both reviewers endorsed (Reviewer B explicitly called it a judgment call).
- **Spec adapter list kept as-is.** I evaluated deriving `preservesCase` from `$lowerCaseColumnNames()` directly (Reviewer A's suggestion), but there's no public model→adapter accessor in the spec context: `$getAdapter()` doesn't exist, and `getClass()` isn't a model method (it errors `The method getClass was not found in the CasePreservation model`). Both reviewers framed the hard-coded list as a non-blocking maintenance note, so it stays; a public accessor would be a reasonable separate follow-up.

Re-verified locally after the change: **Lucee 7 × SQLite (3795/0/0) and H2 (3793/0/0)**, both green.

### On the red `Lucee 7 + SQLite (LuCLI)` check

The failure was the **CLI module suite returning HTTP 500** after the app-reload phase — the **core suite passed `3795 / 0 / 0`** (`[Core Tests] All tests passed!`). It passed on the code commit (`1bdda1113`) and failed only on the **docs-only** commit (`12eccdb4c`), which can't change test behavior — so this is a flaky/pre-existing CLI-suite failure, not a regression from this change. The push above re-triggers the check.

PR #2857, opened 2026-06-04
fix(cli): source reloadPassword from .env in scaffold; reconcile on WHEELS_RELOAD_PASSWORD

bpamiri:
## Why

[Discussion #2854](https://github.com/wheels-dev/wheels/discussions/2854) reported that `wheels new` hard-codes the generated reload password as a literal in `config/settings.cfm` — a **committed** file — and echoes it again in a comment, so the secret lands in version control. The docs also disagreed on the variable name and accessor, so copy-pasting a documented snippet into a fresh app could resolve to `""` and trip the "reloadPassword is empty" warning.

Tracing the code confirmed:
- The scaffold wrote the random value to **three** places: `config/settings.cfm` `set()`, a `settings.cfm` comment, and `.env`.
- `.env` used `RELOAD_PASSWORD` while the deploy templates + upgrade guide used `WHEELS_RELOAD_PASSWORD`.
- The "Environments and Configuration" guide used `application.wo.env(...)`; the migration guide called bare `env()` the "canonical pattern"; `production-config.mdx` called `application.wo.env(...)` "the canonical accessor." Genuinely split.

## What changed

**Canonical form going forward:** `set(reloadPassword = env("WHEELS_RELOAD_PASSWORD", ""))`, with the secret living only in the git-ignored `.env`.

### Scaffold (`cli/lucli/templates/app/`)
- `config/settings.cfm` + both `app/snippets/*.txt` now read the password via `env("WHEELS_RELOAD_PASSWORD", "")`. The generator's random value is written **only** to `.env`; the literal is gone from the committed file and its comment.
- `_env` now emits `WHEELS_RELOAD_PASSWORD` — consistent with `WHEELS_ENV` / `WHEELS_DATASOURCE` and the `wheels deploy` secrets contract, so the **same** `env()` lookup works unchanged from a laptop `.env` to a production deploy.

### CLI / console
- `detectReloadPassword()` (and its e2e mirror) now match `(?:WHEELS_)?RELOAD_PASSWORD`, so apps generated before the rename keep working. Added a spec asserting `WHEELS_RELOAD_PASSWORD` detection.
- Console + `wheels reload` hint strings now name `WHEELS_RELOAD_PASSWORD`.

### Docs — standardized on bare `env()` + `WHEELS_RELOAD_PASSWORD`
- `core-concepts/environments-and-configuration` and all 5 deployment guides (`production-config`, `security-hardening`, `docker-deployment`, `vm-deployment`, `dev-server`).
- Replaced an **insecure** docker example — `Server.System.getEnv("RELOAD_PASSWORD") ?: "changeme"` (wrong accessor that bypasses `.env`, unprefixed name, guessable fallback) — with `env("WHEELS_RELOAD_PASSWORD", "")`.
- Refreshed the stale `security-hardening` note that claimed the template hard-codes `{{reloadPassword}}`, plus a stale `Global.cfc` line reference.

## Scope note

The audit found the drift was wider than the two docs cited in the discussion — hence the 5 deploy-guide edits. **Deferred to a follow-up:** 13 `application.wo.env` usages remain in 4 non-deploy guides (`database-and-multiple-datasources`, `sending-email`, `authentication-patterns`, `testing`) for a separate accessor sweep.

- [ ] Follow-up: convert remaining `application.wo.env` → `env()` in the 4 non-deploy guides

## Verification

- ✅ Regex logic verified: `(?:WHEELS_)?RELOAD_PASSWORD\s*=\s*(.+)` matches both `WHEELS_RELOAD_PASSWORD=` and legacy `RELOAD_PASSWORD=` and captures the value.
- ✅ Scaffold output grep'd clean — no `reloadPassword="..."` literal remains.
- ✅ All `{test:compile}` doc snippets use the bare `<cfscript> set(…=env(…))` form already proven to compile in the migration guide.
- ⏳ Full CFML/e2e/docs-harness/compat-matrix validation runs in CI against this branch (the e2e suite that holds the new spec runs in `deploy-ci.yml`; the docs harness needs `npm ci` + a CFML backend; the onboarding harness can't reliably test worktree templates locally due to the brew-wrapper `LUCLI_HOME` gotcha). Not claimed as locally green.


bpamiri:
Thanks @wheels-bot — all four addressed in a7b5cd722:

- **`examples/starter-app/config/settings.cfm` committed literal** — ✅ fixed. Now `set(reloadPassword=env("WHEELS_RELOAD_PASSWORD", ""))` (the starter-app's `Application.cfc` already loads `.env`), and `examples/starter-app/.env.example` gains a `WHEELS_RELOAD_PASSWORD=` entry. Good catch — a committed `"changeme"` is actually worse than the random literal the scaffold used.
- **`cli/src/.../config/check.cfc:374` `--fix` sample `.env`** — ✅ fixed (`RELOAD_PASSWORD=` → `WHEELS_RELOAD_PASSWORD=`). You're right that `env()` is an exact-key lookup at the CFML layer — the `(?:WHEELS_)?` alias only lives in the CLI's own `detectReloadPassword()` regex — so the legacy CommandBox `wheels config check --fix` would otherwise hand users a `.env` the framework can't read. The fix also makes that sample internally consistent with the `WHEELS_ENV` line two rows above it.
- **Stale test assertion message** — ✅ updated to `"Should check .env for WHEELS_RELOAD_PASSWORD (or legacy RELOAD_PASSWORD)"`.
- **CHANGELOG** — ✅ added an `[Unreleased] ### Fixed` entry covering the rename and the `RELOAD_PASSWORD` → `WHEELS_RELOAD_PASSWORD` `.env` migration path for existing apps.

On the non-blocking nits: left the docker `ENV …=change-me` default as-is (the compose `${WHEELS_RELOAD_PASSWORD:?required}` form is the authoritative pattern, as you noted) and the pre-existing snippet duplication. For completeness on the other in-repo apps I checked: `examples/tweet` already uses `set(reloadPassword="")` (no secret), and the root demo app + `tools/build/base` keep their known dev passwords as test/build infrastructure (the CLI test harness reloads against the root app), so I left those untouched.

teleotic:
`wheels new` also hard-codes the same password into lucee.json as the Lucee admin password.

bpamiri:
> `wheels new` also hard-codes the same password into lucee.json as the Lucee admin password

This one is probably fine since we are seeding the admin password for the Lucee server but it should probably then be .gitignored. Let me analyze this one and the ramifications.

bpamiri:
Fixed the blocking item in 78d4a94cc, @wheels-bot — genuinely good catch, this was a regression I introduced by renaming the key without sweeping all consumers.

- **`tools/test-onboarding.sh` (6 reload points)** — ✅ the anchored `grep -E '^RELOAD_PASSWORD='` can't match a line starting with `WHEELS_RELOAD_PASSWORD=`, so the harness fell back to `"wheels"` and every reload hit the app with the wrong password. Now `^(WHEELS_)?RELOAD_PASSWORD=` (matches both the new and legacy keys); verified the `grep | cut` extracts the value for each form.

I also did the full-repo sweep I should have run when renaming:
- `tools/test-cli-e2e.sh` — unaffected (unanchored `grep RELOAD_PASSWORD` matches the prefixed key as a substring), as you noted.
- `v3-0-0/*` guides, `docs/superpowers/*` plans, and `examples/tweet/.ai/*` reference docs — left as-is (prior-version docs / historical planning records / example-internal AI docs on a different accessor — out of this PR's scope).

On the two non-blocking test fixtures (`ServerCommandsTest.cfc:919`, `ConfigCommandsTest.cfc:190`): leaving both. Neither models real scaffold output — they use arbitrary keys (`APP_NAME`/`PORT`, `DATABASE_NAME`/`ENCRYPTION_SALT`), the back-compat regex reads their legacy `RELOAD_PASSWORD` fine, and they give incidental coverage of the legacy detection path. That matches your "non-blocking / follow-up" call.

bpamiri:
Analyzed — good catch, @teleotic, and you're right that it's the same class of leak (and arguably higher-impact, since the Lucee Server Admin password gates the admin UI). Confirmed `lucee.json` was committing the generated value, independent of the `config/settings.cfm` path.

Rather than git-ignore the whole `lucee.json` — which would drop its shareable config (ports, mappings, datasources) from version control and break `wheels start` on a fresh clone — LuCLI's native env interpolation gives a cleaner fix. `LuceeServerConfig` runs `admin.password` through `#env:VAR#` substitution at server start, resolved against the project `.env` (which LuCLI auto-loads). So as of 289f6d5c5 the scaffold's `lucee.json` reads:

```json
"admin": { "enabled": true, "password": "#env:WHEELS_RELOAD_PASSWORD#" }
```

The random value now lives only in the git-ignored `.env`, the committed `lucee.json` carries no secret, and its config stays version-controlled. (It reuses the existing `WHEELS_RELOAD_PASSWORD` since the scaffold already used one value for both; decoupling into a distinct `WHEELS_LUCEE_ADMIN_PASSWORD` would be an easy follow-up.) Thanks for flagging it.

bpamiri:
Thanks @wheels-bot — both non-blocking points confirmed, no code change needed.

**Fallback when `WHEELS_RELOAD_PASSWORD` is unset:** LuCLI leaves the literal `#env:WHEELS_RELOAD_PASSWORD#` as the admin password — it does **not** substitute empty. In `LuceeServerConfig.replaceLucliVars()`, the no-default branch is:

```java
replacement = getEnvVar(varExpression);
if (replacement == null) {
    replacement = matcher.group(0); // keep the literal "#env:WHEELS_RELOAD_PASSWORD#"
}
```

`getEnvVar()` checks the loaded `.env` first, then the system environment; if absent from both, the placeholder token is kept verbatim. So the admin password is never left blank/open — it falls back to a non-empty literal. On the `wheels new` happy path `.env` always carries `WHEELS_RELOAD_PASSWORD=<random>`, so it resolves to the random value. (We could switch to the `#env:VAR:-default#` form, but an empty default would be *worse* — passwordless admin — so the literal fallback is the safer default. Your second hypothesized outcome is the actual one.)

**Test coverage:** yes, exercised implicitly by `tools/test-onboarding.sh` — that harness runs `wheels start`, which is exactly where LuCLI loads `lucee.json` and resolves `admin.password` through this `#env:` path against the generated `.env`. A dedicated unit test for the interpolation itself belongs in the LuCLI repo (where the resolver lives), not here.

The two test-fixture nits (`ServerCommandsTest.cfc:919`, `ConfigCommandsTest.cfc:190`) and the docker `ENV` placeholder are unchanged from prior rounds — non-blocking, follow-up.

PR #2847, opened 2026-06-03
feat(bot): unblock release pipeline (title-lint, freshen, tiered conflict resolution)

bpamiri:
## Summary

Fixes the two recurring stalls in the wheels-bot release pipeline, each at the right layer.

**Commit-message CI failures** → `pr.yml` now lints the **PR title** (the squash subject) instead of every commit. The repo squash-merges, so intermediate commit headers never land in `develop` — yet the old per-commit check failed on throwaway bot commits (e.g. [#2845](https://github.com/wheels-dev/wheels/pull/2845)'s 105-char intermediate, while its valid title was ignored). Adds the `edited` trigger so a corrected title re-runs the check, and guards `fast-test` so title-only edits don't re-run the suite. Also corrects the `_shared-rails.md` rail that capped the *subject* when commitlint measures the whole *header* (`type(scope): ` prefix included).

**Merge conflicts from in-flight work** → two new workflows, cloning the proven `bot-address-review.yml` scaffold (App-token auth, skip-check idempotency):
- **`bot-freshen.yml`** — on push to `develop` (+30-min backstop), behind-but-clean bot PRs are brought current via non-destructive `update-branch`; `DIRTY` ones are dispatched to the resolver. (`freshen-decide.sh` maps merge state → action.)
- **`bot-resolve-conflicts.yml` + `/resolve-conflicts`** — a **deterministic** classifier (`classify-conflicts.sh`) gates resolution: auto-resolve content/docs conflicts (markdown/MDX anywhere, CHANGELOG, `.ai/`, `docs/`) then push; **escalate** anything touching code with a `conflict:needs-human` label + comment — never auto-resolved.

### Explicitly NOT automated
- **Merging** — PRs are brought to a green, conflict-free, ready state; the maintainer performs the final squash-merge.
- **Code-conflict resolution** — always escalated to a human (a green suite doesn't prove a code merge preserved intent).
- No force-push (freshen uses merge-from-base, not rebase).

## Required manual follow-up
- [ ] **Flip the repo to squash-only** (deferred — repo-settings change, intentionally not done by automation): `gh api -X PATCH repos/wheels-dev/wheels -F allow_merge_commit=false`. Makes "the PR title is what lands" a guarantee rather than a convention, closing the merge-commit path where an unlinted intermediate header could land verbatim.

## Test plan
- Local suites, all green: `tools/test-commit-title.sh`, `tools/test-classify-conflicts.sh`, `tools/test-freshen-decide.sh`.
[...]

PR #2865, opened 2026-06-05
fix(bot): thread captured head SHA into wheels-bot review markers

### LINKED ISSUE #2848: wheels-bot embeds a stale SHA in review idempotency markers, causing review re-fires
## Summary

Reviewer A/B comments carry an idempotency marker — `<!-- wheels-bot:review-a:<pr>:<SHA> -->` (and `review-b`). The embedded `<SHA>` can **lag the commit the review actually ran against** (`commit_id`). When that happens, the skip-check gate doesn't recognize the current head as already-reviewed, so Reviewer A re-fires on superseded commits and Reviewer B can emit contradictory verdicts.

## Evidence (from #2847, merged `70291faff`)

Across several rapid pushes, Reviewer B self-diagnosed this **twice**:

- > "review 4's idempotency marker contained the wrong SHA (`2474385d3` instead of `315b3c0a`)"
- > "the marker says `315b3c0a` but the review was submitted against head `e476c511f` (confirmed via the GitHub `commit_id` field)."

The result was a 5-round review thrash, including one Reviewer B `converged-approve` and one Reviewer B "not aligned" pointing at *different* SHAs.

## Impact

- Redundant Reviewer A/B runs on a PR that's already been reviewed → wasted Anthropic API spend.
- Confusing, self-contradictory review history; harder for a maintainer to know the real verdict.
- The skip-check idempotency can fail to suppress duplicate runs.

## Where to look

- Marker emission: `.claude/commands/review-pr.md` and `.claude/commands/review-the-review.md` (where the `<!-- wheels-bot:review-*:... -->` marker string is written).
- SHA source: how `.github/workflows/bot-review-a.yml` / `bot-review-b.yml` resolve and pass the head SHA into the prompt.

## Acceptance criteria

- The marker SHA always equals the commit the review ran against (checked-out head / review `commit_id`).
- Capture the head SHA once (at checkout) and thread it consistently into the marker rather than re-deriving it later.
- On a PR with several quick successive pushes, each review's marker SHA matches its `commit_id`, and the skip-check suppresses duplicate runs for an already-reviewed head.

_Surfaced during the #2847 review cycle._

bpamiri:
## Summary

Fixes #2848 — wheels-bot embedded a **stale SHA** in Reviewer A/B idempotency markers, so the skip-check gate failed to recognise an already-reviewed head and Reviewer A re-fired on superseded commits while Reviewer B emitted contradictory verdicts on different SHAs (observed across the #2847 review cycle, where B self-diagnosed the drift twice).

## Root cause

The marker (`<!-- wheels-bot:review-a:<pr>:<sha> -->` / `review-b`) is written by the skill prompts, but the prompts **re-derived `<sha>` at review time** via `gh pr view --json headRefOid`. That call races with pushes that land mid-session: between the workflow's checkout and the model's `gh pr view`, a new push can move the PR head, so:

- the marker SHA ≠ the commit the review actually ran against, and
- the skip-check (`marker-pattern`) keyed on a different SHA than the marker, so it stopped suppressing duplicate runs.

The workflows *already* capture the authoritative SHA (and check it out) — they just never handed it to the model.

## Fix: capture once, thread into the prompt as an argument

Capture the head SHA once at the workflow level and pass it into the prompt as an explicit `<head-sha>` argument. The prompt emits the marker from that value and never re-derives it.

| Surface | SHA source |
|---|---|
| `bot-review-a.yml` → `/review-pr`, `/respond-to-critique` | `${{ steps.pr.outputs.sha }}` (the SHA the Checkout step already pinned) |
| `bot-review-b.yml` → checkout `ref`, skip-check `marker-pattern`, `/review-the-review` | `${{ github.event.review.commit_id }}` (the commit Reviewer A's review — which B critiques — was attached to; immune to head drift) |

### Why a prompt argument and not the `REVIEW_HEAD_SHA` env var the issue sketched

The Reviewer A/B `--allowedTools` is `Bash(gh:*)` + read-only `git` (`git log/diff/show/grep/status`) — **no `echo`, `printenv`, or bare `Bash`**. The model therefore has *no way to read a step-level environment variable's value* to compare against existing markers or to emit a new one; it would silently fall back to `gh pr view` and the race would persist. So the SHA must arrive in the **prompt text** — the same channel the PR number already travels on (`cmd=/review-pr <pr_num> <sha>`). This is the one substantive design change from the triage's sketch, and it's why the fix actually works under the existing tool sandbox.

## Scope

- **`respond-to-critique.md` is included** even though the issue named only `review-pr.md` / `review-the-review.md`: it runs in the *same* "Run Reviewer A" step and has the identical race. Leaving it would let the response-path markers keep drifting.
- **Reviewer B's convergence/terminal markers** (`converged-approve`/`converged-changes`/`:terminal`) are fixed for free — review-the-review.md now resolves *every* `<sha>` it writes from the passed `<head-sha>`.
- **Out of scope (documented follow-up):** `address-review.md` and `advise-on-deadlock.md` also re-derive the SHA via `gh pr view`, but they live in the convergence/deadlock loop (a different trigger path) and are *not* broken by this change — all marker *triggers* are presence-based, and `address-review.md` reads B's comments without a SHA-equality gate. Worth a separate pass.

## Testing

New structural spec `vendor/wheels/tests/specs/cli/BotReviewMarkerShaThreadingSpec.cfc` (modeled on `OnErrorFallbackGuardSpec.cfc` / `ConfigRoutesStaleDocUrlSpec.cfc`) asserts the wiring across both workflow YAMLs and all three prompts.

Ran the core `cli` suite locally on Lucee 7 + SQLite (`/wheels/core/tests?...&directory=wheels.tests.specs.cli`):

- **Before fix:** spec bundle `0 pass / 9 fail` (true RED — each assertion fails for the right reason; an earlier weak `<head-sha>` substring check false-passed on a pre-existing `git log origin/develop..<head-sha>` line, so it was replaced with a behavioral `headRefOid`-absent assertion).
- **After fix:** spec bundle `9 pass / 0 fail`; whole `cli` suite `59 pass / 0 fail / 0 error`, no regressions.

`actionlint` passes on both workflows (only a pre-existing SC2129 *style* nit on the untouched "Resolve PR info" step).

## Notes for review

- **Security:** the only values threaded in are commit SHAs (`steps.pr.outputs.sha`, `github.event.review.commit_id` — GitHub-generated hex) and a PR number already validated `^[0-9]+$` upstream — not free-text fields (titles/bodies), so no command/ref-injection surface is added. Swapping one GitHub SHA (`head.sha`) for another (`review.commit_id`) in `ref:` keeps the same safety profile.
- **Cross-engine:** the spec uses only `fileRead` / `reFindNoCase` and the arrow-fn + IIFE loop-capture pattern already used by the sibling structural specs; `#` is escaped as `##` per the suite's convention.
- This PR's *own* Reviewer A/B run uses develop's current (pre-fix) workflows — for `pull_request` events GitHub runs the base-branch workflow — so the fix only takes effect for PRs opened after merge.

Closes #2848


bpamiri:
## Second commit (`94721fdf`) — fixes a self-inflicted review regression found while watching this PR

While watching this PR's own Reviewer A run, I noticed Reviewer A completed its session but **posted no review** — yet every recent PR (#2862, #2863, #2856, #2851, #2847) gets one on the first run.

**Root cause:** the first commit's prompt wording said *"do not re-derive the head with `gh pr view`"* / *"do not call `gh pr view`"*. An over-literal model reads that as "avoid `gh pr view` entirely" — but step 2 of each prompt uses `gh pr view` / `gh pr diff` to read the PR diff and context. The session floundered on denied alternatives (8 permission denials vs the baseline's ~4) and never posted a review.

**Fix (`94721fdf`):** reword all three prompts to forbid only *computing the marker SHA yourself*, and explicitly affirm `gh pr view` / `gh pr diff` remain how the model reads PR content. The SHA-threading behaviour and the structural spec are unchanged (still 9/9 locally; `headRefOid` stays absent so the racy derivation can't creep back).

The fresh Reviewer A run on `94721fdf` is the live test that posting works again.

bpamiri:
## `609f934c` addresses the converged-changes findings

Both findings from the Reviewer A/B consensus are fixed:

1. **Response-loop SHA divergence** — `bot-review-a.yml`'s "Resolve PR info" response branch now extracts the reviewed SHA from the triggering `review-b` comment marker (`wheels-bot:review-b:<pr>:<sha>:<round>`) instead of `gh pr view`'s current head, so A's response anchors to the same commit B critiqued. Read via a `COMMENT_BODY` env var (not spliced into the script) per B's injection note.
2. **Spec gap** — added an assertion that `bot-review-a.yml` derives no SHA via `headRefOid`, mirroring the prompt-file assertions, so the floating derivation can't creep back.

Spec is 10/10 locally and `actionlint` is clean.

_Note: `bot-address-review.yml` failed to auto-apply this — its App token lacks the `workflows` permission (the same constraint that put #2848 "on hold" for the bot originally), so I applied it by hand._

Issue #2861, opened 2026-06-05
CLI: commands flatten-and-reparse LuCLI args instead of reading argCollection directly (tech debt behind #2855)


bpamiri:
## Background

[#2856](https://github.com/wheels-dev/wheels/pull/2856) fixed the `--no-*` flag family (`--no-sqlite`, `--no-routes`, `--no-test-db`, `--no-open-browser`) by re-emitting `--no-<key>` for `false` values in `Module.cfc::argsFromCollection()`. That is the correct, minimal bug fix for #2855 and it is shipping. This issue tracks the underlying architectural smell it exposed, which was deliberately out of scope for a bug fix.

## The smell

LuCLI parses the command line into a **structured** `argCollection` map before dispatching to a module. For `wheels new myapp --no-sqlite`, the module's `new()` receives:

```
argumentCollection = { arg1: "myapp", sqlite: "false" }
```

(`--no-<key>` is normalized to `<key>=false`; see `LuceeScriptEngine.parseArguments()` in the LuCLI runtime, lines ~335–365.)

`cli/lucli/Module.cfc` then **discards that structure**: `getArgs()` → `argsFromCollection()` flattens the map back into a flat argv string array (`["myapp", "--no-sqlite"]`), and each command re-parses it with a hand-rolled token loop:

```cfm
} else if (arg == "--no-sqlite") {
    options.noSQLite = true;
}
```

So the data flows **structured → flattened → re-parsed**. `getArgs(arguments)` is used by ~18 commands.

## Why it matters

- **It is the root cause of #2855.** The flatten step (`argsFromCollection`) silently dropped every `false` value, so `--no-*` flags never survived the round-trip. #2856 re-adds them, but the round-trip itself is the fragile part.
- **It is lossy.** Re-emitting `--no-<key>` for any `false` value cannot distinguish a genuine `--no-flag` negation from an explicit `--flag=false` literal (both reach the module as `key=false`). Benign for today's boolean flags, but a latent trap for any future `--flag=value` option whose value can be `false`. (Flagged in the #2856 review.)
- **Double-maintenance.** Every command re-implements positional + flag parsing that LuCLI already performed.

## Proposed direction

[...]

Issue #2855, opened 2026-06-04
Wheels CLI ignores --no-sqlite argument


teleotic:
**Describe the bug**
Using `wheels new --no-sqlite` incorrectly results in SQLite database files still being created, and SQLite datasource code being generated. This was partially addressed in issue #2621 but the behavior remains. On investigation, the `--no-sqlite` argument is never passed through to Wheels CLI module.

**To Reproduce**
1. Create new app with `wheels new test_app --no-sqlite`
2. App is erroneously created with a SQLite database

**Expected behavior**
The SQLite database and related code should not be generated.

**Desktop (please complete the following information):**
 - OS: Ubuntu 26.04 LTS, Windows 11 Pro 25H2
 - Version: Wheels 4.0.2 (Linux),  4.0.3-snapshot.1966 (bleeding-edge)

**Additional context**
After testing several variations, it appears that no command line argument beginning with `--no-` is passed through to the Wheels CLI module (`/wheels/cli/lucli/Module.cfc`). Therefore `options.noSQLite` is never set to true and the default behavior continues. I have not determined whether this is a Wheels issue or a limitation with the upstream LuCLI tool. (LuCLI has a `--no-agents` option that may be causing a parsing problem, but this is just speculation.)

A possible quick fix is renaming the command line argument. For instance `--nosqlite` correctly passes through.

Issue #2313, opened 2026-04-26
Polish bucket: minor CLI / framework / doc inconsistencies surfaced by VM-onboarding journals


bpamiri:
## Summary

Tracker for the smaller-impact findings from the two fresh-VM tutorial-onboarding journals (April 2026). None of these block the cliff fixes already in flight (#2304, #2306, #2307, #2308, #2309, [homebrew-wheels#9](https://github.com/wheels-dev/homebrew-wheels/pull/9)) — they're polish items that affect first-time-user impression but not whether the tutorial completes.

Each item below is a separate small fix. Check off as completed.

---

## CLI

- [ ] **\`wheels stop\` is silently scoped to cwd; no helpful error from a non-Wheels directory.** Running \`wheels stop\` from a directory other than the app's root prints \"No running server found for this directory.\" without saying \"to stop a server in another directory, cd into it or pass --name\". Worse: \`wheels start\` from a non-Wheels dir creates a phantom server named after the cwd basename, polluting \`~/.wheels/servers/\`. **Fix:** \`wheels stop\` with no server in cwd should suggest the workaround, and \`wheels start\` from a directory with no \`app/\` or \`config/routes.cfm\` should refuse rather than create a phantom server.

- [ ] **\`wheels db status\` throws \`key [MESSAGE] doesn't exist\`.** The CLI reads \`message\` (singular) off the JSON response, but the framework's \`/wheels/cli\` endpoint returns \`messages\` (plural). **Fix:** guard the JSON parse with the actual response shape — defensively handle both keys, or fix the CLI to read \`messages\`.

- [ ] **\`--help\` per subcommand prints the top-level help instead of subcommand-specific help.** \`wheels migrate --help\`, \`wheels seed --help\`, \`wheels destroy --help\` all print the same generic top-level usage. Stuck users can't discover \`migrate latest|up|down|info\` syntax from the CLI itself. **Fix:** implement per-subcommand \`--help\` for every subcommand listed in \`wheels --help\`.

- [ ] **\`wheels destroy <type> <name>\` rejects the documented argument order.** The CLI is treating the second positional as the type, so \`wheels destroy controller Posts\` errors with \"Unknown type: posts. Valid types: resource, model, controller, view\". **Fix:** swap the parse order to match the docs, OR update the docs to match the parse order. (Tutorial currently falls back to manual \`rm\` so this is cosmetic.)

---

## Framework / SQLite adapter

[...]

Issue #2214, opened 2026-04-22
wheels new: remaining silent-exit-0 error paths in scaffoldNewApp


bpamiri:
Follow-up to #2211.

The fix in #2211 addressed the framework-not-found path specifically. `scaffoldNewApp()` and its caller `new()` in `cli/lucli/Module.cfc` have three remaining error paths that still `return ""` and silently exit 0 — same root cause, same fix pattern.

## Affected paths

1. **Directory already exists** — [cli/lucli/Module.cfc:3217-3220](cli/lucli/Module.cfc)
   ```cfm
   if (directoryExists(targetDir)) {
       out("Directory already exists: #appName#", "red");
       return "";
   }
   ```

2. **App name required** — [cli/lucli/Module.cfc:446-450](cli/lucli/Module.cfc)
   ```cfm
   if (!len(appName)) {
       out("Error: app name is required.", "red");
       out("Usage: wheels new <appname>");
       return "";
   }
   ```

3. **Project template not found** — [cli/lucli/Module.cfc:3237-3240](cli/lucli/Module.cfc)
   ```cfm
   if (!directoryExists(templateDir)) {
       out("Project template not found at: #templateDir#", "red");
       return "";
   }
   ```

All three print a user-facing error, then return an empty string that LuCLI prints as nothing with exit code 0 — fooling automation in the same way #2211 did.

## Proposed fix

Mirror the pattern from #2211: after the friendly `out()` block, throw a typed exception so Picocli's `ExecutionExceptionHandler` produces exit 1.

- `Wheels.TargetDirectoryExists` for case 1
- `Wheels.InvalidArguments` for case 2
- `Wheels.TemplateNotFound` for case 3

## Acceptance

- [ ] `wheels new <existing-dir-name>` exits non-zero
- [ ] `wheels new` (no args) exits non-zero
- [ ] `wheels new foo` with a broken install (missing `templates/app/`) exits non-zero
- [ ] Existing friendly error messages preserved

## Not in scope

Usage/help output when no args are passed is ambiguous — arguably the user wanted help, not an error. Current behavior prints usage and returns "". If we throw here, the "help me" experience degrades. Keep case 2 narrower: throw only when an arg was provided but parsing yielded no app name, not when zero args were passed.

PR #2872, opened 2026-06-06
refactor(cli): migrate 8 leaf commands to ArgSpec, dropping the argv round-trip (#2861)

bpamiri:
## What & why

Per-command migration tracked in #2861. The `ArgSpec` builder foundation shipped in #2862 (`d1cdbe3`) but **no command was migrated** — every subcommand still ran LuCLI's structured `argCollection` through the lossy `getArgs()` → `argsFromCollection()` flatten-and-reparse round trip.

This PR migrates the **8 clean leaf commands** to consume the structured collection directly via `ArgSpec.parse(structuredArgs(arguments))`, dropping each hand-rolled token loop:

`new` · `seed` · `notes` · `analyze` · `doctor` · `stats` · `upgrade` · `destroy`

## Fixes a latent bug the round trip masked

`getArgs()` only rebuilt argv when a positional `arg1` was present:

```cfm
if (structKeyExists(callerArgs, "arg1")) { return argsFromCollection(callerArgs); }
var raw = __arguments ?: []; ...   // <- named-only invocations fall through to []
```

LuCLI hands a command like `wheels seed --environment=production` the collection `{environment:"production"}` — **no `arg1`** — so `getArgs()` fell through to the empty `__arguments` fallback and the flag was silently dropped. Confirmed empirically (regression test, see below): `getArgs({verbose:"true"})` → `[]`.

Affected commands whose **named-only** flags were silently ignored before this PR:
- `wheels seed --environment=… --mode=… --generate`
- `wheels doctor --verbose`
- `wheels stats --verbose`
- `wheels notes --annotations=… --custom=…`

Consuming the structured map directly means the named keys (and `--no-X` negations) survive — no `arg1` gate, no flatten, no re-parse.

> The user-facing #2855 symptom (`--no-*` dropped on commands **with** a positional, e.g. `new --no-sqlite`) was already mitigated by #2856's re-emit and stays fixed; this PR removes the round trip those commands relied on.

## How

A small sourcing layer feeds `ArgSpec`:

- **`structuredArgs(callerArgs)`** — returns LuCLI's live `argCollection` when present; otherwise reconstructs it from the instance-level `__arguments` array (internal delegation like `create` → `new`, and unit tests).
- **`argvToCollection(argv)`** — faithful inverse of LuCLI's `parseArguments()` (`--no-X`→`X=false`, `--X`→`X=true`, `--k=v`→`k=v`, positionals→`arg<n>` by global index).

Each command extracts a private, side-effect-free `parse<Command>Args(coll)` helper (testable) and keeps its execution path unchanged.

## Behavior preserved (the parsing changed, the logic didn't)

- `new`: usage on no args, the #2214 "app name required" throw, datasource/password defaults.
- `destroy`: `<type> <name>` / `<name> <type>` smart reorder (#2313) — **now gap-tolerant**: positionals are gathered by sorted `arg<n>` index, so `--force` works before *or* after the pair (LuCLI's global token numbering leaves an `arg1` gap when a flag leads).
- `upgrade`: `check`-gate, read-only help, and the `--dry-run` / `--to` "did you mean" nudge.
- `doctor` / `stats`: `-v` shorthand still works — LuCLI delivers short flags as positionals (it only normalizes `--x` / `--no-x`), so the parser checks for a `-v` positional in addition to the `--verbose` flag.
- `analyze`: the "not in a Wheels project" guard still fires only for the bare `wheels analyze`.

## Scope (why 8, not all ~18)

Deliberately limited to leaf commands that map cleanly to `ArgSpec` as shipped and delegate **parsed values** (not a flat args array). Deferred to the follow-up (they need `ArgSpec` extended and/or carry regression risk):

- **Dispatchers** `generate` / `create` / `db` / `browser` — pass a flat `args` array to sub-handlers.
- **Parser-delegating** `deploy` / `packages` — own parsers, nested sub-verbs, variadic positionals.
- **`migrate`** — dispatches on the action and passes `args` to `runForgetOrPretend`.
- **`test` / `console`** — undocumented space-separated option forms (`--filter value`, `--password value`) that `ArgSpec` doesn't model.

`getArgs()` / `argsFromCollection()` stay as the deprecated shim until those convert; shim removal is the final step of #2861.

## Test plan

- New server-free specs: `cli/lucli/tests/specs/commands/CommandArgParsingSpec.cfc` (37 specs) via an extended `ModuleArgvProbe`, covering `argvToCollection`, `structuredArgs`, the `getArgs` named-only-drop **regression characterization**, and every `parse<Command>Args` helper (named-only fix, `--no-X`, `-v`, the destroy `--force`-first gap, the #2214 empty/throw split).
- Ran the full CLI suite locally against a worktree server (`/wheels/cli/tests`): **`CommandArgParsingSpec` 37 pass / 0 fail / 0 error**; suite **711 pass / 0 fail**. The 7 remaining errors are pre-existing/environmental (`DbCommandsSpec`, `IntrospectSpec` — HTTP integration specs that need a migrated demo DB) and are present on `develop` unchanged.
- `Module.cfc` compiles cleanly (711 specs ran; a syntax error would crash the bundle).

## Follow-up (#2861 remaining)

- [ ] Migrate the dispatchers (`generate`/`create`/`db`/`browser`) and `migrate`.
- [ ] Extend `ArgSpec` (variadic positionals / passthrough) and migrate `deploy`/`packages`, and the space-form `test`/`console`.
- [ ] Remove the `getArgs()` / `argsFromCollection()` shim once every call site is converted.
- [ ] Add the `tools/test-onboarding.sh` e2e assertion for `wheels new <app> --no-sqlite` (issue's test-robustness bullet).

Refs #2861, #2862, #2855.


bpamiri:
Addressed Reviewer A/B feedback in `1a85ae7e3`:

- **`wheels new --flags-only` delta** — CHANGELOG now calls it out explicitly (usage → `Wheels.InvalidArguments` throw) and the inaccurate "behavior otherwise preserved" wording is corrected. Added a spec pinning it: `parseNewArgs({sqlite:"false"}).isEmpty == false` (the coverage gap Reviewer B noted).
- **Dead `hasArgs`** — removed from `parseUpgradeArgs`'s return and its spec assertion; `upgrade()` only reads `isCheck`.
- **Test isolation** — `ModuleArgvProbe.$structuredArgs` / `$getArgs` are now self-isolating: `underscoreArguments` defaults to `[]` and is always assigned to `variables.__arguments`, so the shared `beforeAll` probe can't carry a stale fallback into a later spec (stronger than a teardown — a future spec author can't reintroduce the trap).
- **Regex consistency** — `^arg[0-9]+$` → `^arg\d+$` in `parseDestroyArgs`, matching the `ArgSpec`/`parseVerboseFlag` precedent.

Full CLI suite green locally: `CommandArgParsingSpec` 38 pass / 0 fail / 0 error; suite 712 pass / 0 fail (the 7 remaining errors are pre-existing `DbCommandsSpec`/`IntrospectSpec` environment failures on a fresh, un-migrated demo DB, unchanged on `develop`).

PR #2873, opened 2026-06-06
fix(cli): resolve CLI services via module-relative path so packaged module loads

bpamiri:
## Problem

The **Wheels Snapshots** workflow has failed on every `develop` push since [#2861](https://github.com/wheels-dev/wheels/pull/2861). The `build / Smoke Test Installed Distribution` job dies at Phase 3 (`wheels new`):

```
=== Phase 3: Scaffold a scratch app ===
[ERROR] Error: lucee.runtime.exp.ApplicationException: could not find component or class with name [cli.lucli.services.ArgSpec]
FAIL: wheels new smoke failed
```

Affected runs: [27063480471](https://github.com/wheels-dev/wheels/actions/runs/27063480471), [27055040302](https://github.com/wheels-dev/wheels/actions/runs/27055040302) (both identical).

## Root cause

#2861 (ArgSpec migration) and #2363 (`TestRunner`) instantiated CLI services with the **absolute FQN** `new cli.lucli.services.X()`. That path only resolves against the source-tree layout `cli/lucli/services/`.

The module tarball is built with `tar -C cli/lucli .` ([`release.yml`](.github/workflows/release.yml)), which flattens the module root so services live at `<module-root>/services/` — there is **no** `cli/lucli/` tree and **no** `cli.lucli` mapping in the package. The 17 sibling services already use the **module-relative** `new services.X()`, which resolves relative to `Module.cfc`'s own directory in *both* the source tree and the package.

`fast-test` passes because it runs from source (both forms resolve); only the smoke test, which runs the installed distribution, caught the regression.

## Fix

Convert all 8 absolute references to the module-relative form:
- `cli/lucli/Module.cfc` — 7× `ArgSpec` (the snapshot blocker) + 1× `TestRunner` (a latent twin; no CI path exercises `wheels test` against the installed module yet)
- `cli/lucli/services/ArgSpec.cfc` — the docblock usage example, so the canonical pattern can't re-seed the bug

## Verification

Reproduced the smoke test locally — built the module tarball from this branch **and** from pre-fix `HEAD` (same `tar -C cli/lucli .` packaging, Java 21, isolated `LUCLI_HOME`), then ran `wheels new` against each:

| Build | Result |
|---|---|
| pre-fix `HEAD` | `exit=1` — reproduced `could not find component or class with name [cli.lucli.services.ArgSpec]` |
| this branch | `exit=0` — **"Application created!"** |

CI on this PR: `fast-test` (core) + Deploy Subsystem CI `cli-tests` (full CLI suite, 365+/0/0 — `Module.cfc` is in its path filter). The snapshot smoke test itself only runs post-merge on `develop`.


bpamiri:
Addressing Reviewer A's two nits:

**Docs / CHANGELOG** — Added under `[Unreleased] ### Fixed` in `fff5ccba9`.

**Tests** — I looked at the suggested guard and it wouldn't actually catch this regression. A spec that instantiates `new services.ArgSpec()` runs from the **source tree**, where *both* the relative path and the absolute FQN resolve. So such a test passes regardless of which form `Module.cfc` uses — it can't detect a revert to `new cli.lucli.services.X()`. (This is the same source-vs-package blindness that let `fast-test` stay green while only the packaged smoke test caught the original break.)

The guards that *do* catch it:
1. The existing `tools/ci/smoke-test-module.sh` (post-merge) — already covers it; this PR turns it green.
2. A **static source-scan** of production `Module.cfc` for `new cli.lucli.` — this would gate pre-merge (the Deploy Subsystem CI runs the CLI suite on `Module.cfc` changes). It must use a plain `findNoCase` substring check, **not** a global comment-stripping regex, because a non-greedy `/* ... */` strip over a 4,500-line file hangs Lucee 7 (prior art: the line-anchored-matching note in our cross-engine docs).

Rather than bundle a hard-to-locally-verify spec into this regression hotfix, I'm tracking the static-scan guard as a follow-up so it can be written and verified properly on its own.

Verification on this PR: built the module tarball from this branch and from pre-fix `HEAD`, installed each into an isolated `LUCLI_HOME`, and ran `wheels new` — pre-fix reproduces `could not find component or class with name [cli.lucli.services.ArgSpec]`; this branch scaffolds the app (`exit 0`). The Deploy Subsystem CI (full CLI suite) is green, confirming no regression in the 8 migrated commands.

(Minor: the sign-off is `Signed-off-by: Peter Amiri <peter@alurium.com>` — the review quoted `amurium`.)

Issue #2878, opened 2026-06-07
Wheels migration script runs against wrong Lucee instance


teleotic:
**Describe the bug**
As discovered in #2876, when working on multiple apps migration scripts may run against the wrong Lucee instance. This could result in writes going to the wrong database.

**To Reproduce**
1. `wheels new test_app_a`
2. `wheels new test_app_b`
3. `cd test_app_a`
4. `wheels start`
5. `cd ../test_app_b`
6. `wheels generate scaffold Post title body:text publishedAt:datetime --belongsTo=author`
7. `wheels migrate latest`
8. See that test_app_b has now run its migration against test_app_a's Lucee instance

**Expected behavior**
`wheels migrate` should refuse to run, and provide an appropriate error, if the project's associated server is not active. It should not run against another project's server.

**Desktop**
 - OS: Ubuntu 26.04 LTS
 - Version 4.0.2

**Additional context**
In my case, I was working on an app with a PostgreSQL backend. Without stopping that app's server, I created a second app to test some Wheels behavior. The second app was created with default settings,  which created a SQLite database. Running a `wheels migrate latest` in the second app's directory without starting that app's server resulted in an error mentioning PostgreSQL.

Issue #1944, opened 2026-03-14
LuCLI: release.yml drops codegen templates — all template-driven generators emit empty output in installed module


bpamiri:
## Scope change

This issue originally tracked adding three stretch generators (`api-resource`, `helper`, `snippets`). **All three are implemented in source** ([Module.cfc:190-205](cli/lucli/Module.cfc), [CodeGen.cfc](cli/lucli/services/CodeGen.cfc), [Scaffold.cfc](cli/lucli/services/Scaffold.cfc), + 12 snippet templates). However, e2e testing against an installed `wheels` binary surfaced a broader packaging bug that makes most generators unusable — reframing this issue to track the fix.

## Bug

In a freshly installed Wheels project:

```
$ wheels generate model Post title:string body:text
  create  app/models/Post.cfc

$ cat app/models/Post.cfc
component extends="Model" {
    function config() {
        
        
        
        
    }
}
```

No validations. No relationships. Empty `config()`. Same stripped output for `controller`, `view`, `scaffold`, `api-resource`. `helper` fails loudly: `Template not found: HelperContent.txt`.

**Only `wheels generate snippets` works end-to-end** because its templates ship under `cli/lucli/templates/snippets/`.

## Root cause

[`Templates.cfc::resolveTemplateDir()`](cli/lucli/services/Templates.cfc#L177) expects code-generation templates at `moduleRoot/templates/codegen/` for installed modules. Per [PLAN.md](cli/lucli/PLAN.md), the release workflow is supposed to copy `cli/src/templates/` → `cli/lucli/templates/codegen/` before tarring the module.

[`.github/workflows/release.yml:201`](.github/workflows/release.yml#L201) has no such copy step:

```yaml
tar czf ... wheels-module-${{ env.WHEELS_VERSION }}.tar.gz -C cli/lucli .
```

Result: the distributed tar ships only `admin/`, `app/`, `deploy/`, `migrations/`, `snippets/` under `templates/` — no `codegen/`. All `generateFromTemplate()` calls targeting `ModelContent.txt`, `ControllerContent.txt`, `ApiControllerContent.txt`, `ViewContent.txt`, `HelperContent.txt`, `CRUDContent.txt`, `ActionContent.txt`, `DBMigrate.txt`, `tests/*.txt`, `crud/*.txt`, `admin/*.txt` silently fall through to the empty/error paths.

[...]

PR #2882, opened 2026-06-09
fix(cli): repair broken commands and correct command docs (CLI audit)

bpamiri:
## What & why

A full audit of **every Wheels CLI command** — running the worktree's CLI source in an isolated harness and cross-checking each command's real behavior against its source, `--help`, the root `CLAUDE.md`, `cli/CLAUDE.md`, and the v4-0-0 guides. 24 command groups were audited (fan-out + adversarial verification), surfacing **69 unique confirmed issues**. This PR lands the **verified, low-risk subset** that makes broken commands work and corrects the most misleading agent-facing docs. Larger/behavioral and systemic-LuCLI items are listed below for follow-up.

Every fix here was reproduced and re-verified against the worktree CLI before and after the change.

## Code fixes (commands that did not work as documented)

| Fix | Was | Now |
|---|---|---|
| **`g` alias** | `wheels g model X` → raw `has no function with name [g]` | `g()` → `generate()`, hidden from MCP |
| **`console` / `reload` auto-password** | `detectReloadPassword()` `(.+)` captured across newlines → wrong password on every default scaffold | anchored `([^\r\n]+)` (matches the sibling helper) |
| **`generate migration`** | `dbmigrate/blank.txt` template left `\|DBMigrateExtends\|` tokens → uncompilable migration on packaged installs | always build inline via `buildEmptyMigration()` |
| **`generate api-resource`** | wrote singular route `.resources(name="product")` → never reached `api/Products` | writes plural, matching controller/table |
| **`validate`** | mixed-args / routes / view checks scanned raw source (comment-blind, false +/−) | scan comment-stripped content (anti-pattern #14) |
| **`db status`** | rendered a literal `%s` in the Applied column | `%-19s` |
| **`info`** | counted commented-out `.resources()` | strips comments first |
| **`migrate` help/hint** | listed only `latest, up, down, info, rename-system-tables` | lists all 8 real subcommands (adds `doctor/forget/pretend`) |
| **`wheels new` help + snippet hint** | "reload-password default: app name"; `wheels db:seed` | "default: random"; `wheels seed` |

## Doc fixes (agent-facing)

- **Root `CLAUDE.md`** — removed claims for commands that error or don't exist: `wheels dbmigrate diff` (CFC-only), `wheels generate seed`/`--all`, `wheels db:seed`, `wheels mcp setup`, `wheels status` + the `wheels_server` MCP tool; `wheels test run` → `wheels test` (`run` is a reserved LuCLI builtin); added `g` to the hidden-tools list.
- **`cli/CLAUDE.md`** — rewritten wholesale. It described the **defunct CommandBox CLI** (`box install wheels-cli`, `/cli/src/`, `wheels init`, `wheels g app` via ForgeBox); now documents the current LuCLI architecture (`cli/lucli/Module.cfc` subcommands, `services/`, ArgSpec parsing, brew/scoop install, MCP, testing).

## Verification

- Each command fix reproduced + re-verified in an isolated LUCLI_HOME running the worktree source.
- Relevant CLI specs reviewed (`AnalysisSpec`, `ScaffoldSpec`, `GenerateCommandSpec`) — no assertion regressions.
- The full `tools/test-cli-local.sh` suite couldn't be run from the `.claude/worktrees` worktree (the framework demo app won't stay up there — a known worktree limitation), so **CI is the backstop for the full CLI suite** on this PR.

## Deferred (follow-up — larger / behavioral / systemic, not in this PR)

- [ ] **`wheels test` exit codes / `--ci` / `--reporter=tap`** — `--ci` inert, exit always 0 on failure, TAP reporter crashes on a failing spec (`failOrigin` is an array). Needs a failing-spec test.
- [ ] **`deploy` flags** — `--version=<tag>` is shadowed by picocli's root `--version` (use `--release`); `--config` vs `--configPath`; nested `deploy server`/`secrets` verbs are picocli-shadowed (flat aliases work); `app` role/host filters parsed-but-dropped. Mostly docs + a couple of `DeployArgsParser` arms.
- [ ] **`browser install` / `test run`** verb errors across the testing/CI guides (`install` hits LuCLI's extension installer; `run` is reserved) — guide sweep.
- [ ] **Systemic LuCLI dispatch (highest leverage, likely upstream):** per-command `--help`/`-h`/`help` all print the global banner; `--verbose`/`-v` swallowed (breaks `doctor`/`stats`/`test` verbose); `wheels help` + several error paths leak the raw `lucli`/picocli banner; per-command help is hard-coded in the launcher and drifts.
- [ ] **`ReleaseChannel`** — shipped stable 4.0.2 reports `(development)` (build token clobbered the dev sentinel); `wheels version` vs `wheels --version` format mismatch.
- [ ] **Misc:** `generate enum` parsed-but-not-emitted; silent view-gen failures; `generate test` overwrites without `--force`; `notes` scanner matches strings/identifiers not comments; a command rewrote tracked `vendor/wheels/wheels.json` as a side effect (observed during the audit).
- [ ] LOW guide sample-output drift across the v4-0-0 command-line-tools pages.

Full per-command findings + a prioritized fix plan are available (157 confirmed findings; 69 unique after de-dup).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -614,7 +614,7 @@ wheels seed --environment=production
 wheels seed --generate                 # legacy: random test data
 ```
 
-To scaffold the seed files themselves, use the snippet generator: `wheels generate snippets seed-data` (writes `app/snippets/seeds*.cfm`). There is no `wheels generate seed` generator.
+To scaffold seed templates, use: `wheels generate snippets seed-data` (writes `app/snippets/seeds*.cfm` — copy or move to `app/db/` to activate them). There is no `wheels generate seed` generator.
 
 `seedOnce()`: idempotent — checks `uniqueProperties` via `findOne()`, creates only if not found. Execution: `seeds.cfm` → `seeds/<environment>.cfm`, wrapped in a transaction. Programmatic: `application.wheels.seeder.runSeeds()`. (Note: `wheels db:seed` is NOT a valid command — it errors. Use `wheels seed`.)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,7 +25,7 @@ plugins/                        DEPRECATED — legacy plugin system; modern pack
 | If you touched | Run | Required? |
 |---|---|---|
 | `vendor/wheels/**` | `bash tools/test-local.sh` (full) or `bash tools/test-local.sh <area>` | Always |
-| `app/**` only | Demo-app specs via `wheels test run` | Always |
+| `app/**` only | Demo-app specs via `wheels test` | Always |
 | `cli/lucli/**` | `bash tools/test-cli-local.sh` | Always |
 | Anything cross-engine-risky (closures, `obj.map()`, reserved scopes, struct literals, mixins) | `tools/test-matrix.sh adobe2023 mysql` AND `tools/test-matrix.sh lucee7 mysql` | If touched code matches any anti-pattern below |
 | Added/changed a migration | `wheels migrate latest && wheels migrate down && wheels migrate up` | Always |
@@ -588,13 +588,7 @@ var all = am.diffAll({hints: {"User": {renames: {"full_name": "fullName"}}}, heu
 am.writeMigration(d, "rename_name_field");
 ```
 
-```bash
-wheels dbmigrate diff User
-wheels dbmigrate diff User --rename=full_name:fullName
-wheels dbmigrate diff User --write --name=rename_name
-wheels dbmigrate diff --threshold=0.85
-wheels dbmigrate diff --rename=User.full_name:fullName
-```
+_Auto-migration is currently CFC-only (`wheels.migrator.AutoMigrator`, shown above). There is no `wheels dbmigrate diff` CLI command — invoking it errors._
 
 Result struct: `{modelName, tableName, addColumns, removeColumns, changeColumns, renameColumns, suggestedRenames}`. Limits: PK renames not detected; rename + type change requires separate migrations; calculated properties excluded.
 
@@ -618,11 +612,11 @@ seedOnce(modelName="User", uniqueProperties="email", properties={
 wheels seed                            # auto-detect env (canonical)
 wheels seed --environment=production
 wheels seed --generate                 # legacy: random test data
-wheels generate seed                   # create app/db/seeds.cfm
-wheels generate seed --all             # create seeds.cfm + dev/prod stubs
 ```
 
-`seedOnce()`: idempotent — checks `uniqueProperties` via `findOne()`, creates only if not found. Execution: `seeds.cfm` → `seeds/<environment>.cfm`, wrapped in a transaction. Programmatic: `application.wheels.seeder.runSeeds()`. The legacy `wheels db:seed` is a CommandBox alias — prefer `wheels seed`.
+To scaffold the seed files themselves, use the snippet generator: `wheels generate snippets seed-data` (writes `app/snippets/seeds*.cfm`). There is no `wheels generate seed` generator.
+
+`seedOnce()`: idempotent — checks `uniqueProperties` via `findOne()`, creates only if not found. Execution: `seeds.cfm` → `seeds/<environment>.cfm`, wrapped in a transaction. Programmatic: `application.wheels.seeder.runSeeds()`. (Note: `wheels db:seed` is NOT a valid command — it errors. Use `wheels seed`.)
 
 ## Background Jobs Quick Reference
 
@@ -708,9 +702,9 @@ Notes:
 {"mcpServers":{"wheels":{"command":"wheels","args":["mcp","wheels"]}}}
 ```
 
-Or run `wheels mcp setup` to generate `.mcp.json` + `.opencode.json`.
+There is no `wheels mcp setup` command — copy the JSON above into `.mcp.json` manually (see the MCP integration guide for OpenCode/Cursor variants).
 
-Tools are auto-discovered from `cli/lucli/Module.cfc` public functions, prefixed with the module name (`wheels_generate`, `wheels_migrate`, `wheels_test`, `wheels_reload`, `wheels_seed`, `wheels_analyze`, `wheels_validate`, `wheels_routes`, `wheels_info`, `wheels_destroy`, `wheels_doctor`, `wheels_stats`, `wheels_notes`, `wheels_db`, `wheels_upgrade`, `wheels_create`, `wheels_deploy`). CLI-only tools (`mcp`, `d`, `new`, `console`, `start`, `stop`, `browser`) are hidden via `mcpHiddenTools()`.
+Tools are auto-discovered from `cli/lucli/Module.cfc` public functions, prefixed with the module name (`wheels_generate`, `wheels_migrate`, `wheels_test`, `wheels_reload`, `wheels_seed`, `wheels_analyze`, `wheels_validate`, `wheels_routes`, `wheels_info`, `wheels_destroy`, `wheels_doctor`, `wheels_stats`, `wheels_notes`, `wheels_db`, `wheels_upgrade`, `wheels_create`, `wheels_deploy`). CLI-only tools (`mcp`, `d`, `g`, `new`, `console`, `start`, `stop`, `browser`) are hidden via `mcpHiddenTools()`.
 
 **Deprecated:** the in-dev-server HTTP endpoint at `/wheels/mcp`. Emits a deprecation notice on first request. Migrate to the stdio surface.
 
@@ -725,12 +719,12 @@ Prefer MCP tools when the Wheels MCP server is available. Fall back to CLI other
 | Generate | `wheels_generate(type, name, attributes)` | `wheels g model/controller/scaffold Name attrs` |
 | Migrate | `wheels_migrate(action="latest\|up\|down\|info\|doctor")` | `wheels migrate latest\|up\|down\|info\|doctor` |
 | Migrator reconciliation | — | `wheels migrate forget\|pretend <version> --yes` (shared dev DB orphan cleanup; see #2780) |
-| Test | `wheels_test()` | `wheels test run` |
+| Test | `wheels_test()` | `wheels test` |
 | Reload | `wheels_reload()` | `?reload=true&password=...` |
-| Server | `wheels_server(action="status")` | `wheels start\|stop\|status` |
+| Server | — | `wheels start\|stop` |
 | Analyze | `wheels_analyze(target="all")` | — |
 | Admin | — | `wheels g admin ModelName` |
-| Seed | — | `wheels seed` (legacy alias: `wheels db:seed`) |
+| Seed | — | `wheels seed` |
 
 ## Reference Docs (verified to exist)
 
--- a/cli/CLAUDE.md
+++ b/cli/CLAUDE.md
@@ -1,68 +1,85 @@
-# CLAUDE.md
+# CLAUDE.md — Wheels CLI (`cli/`)
 
-This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+This file guides Claude Code when working in the Wheels CLI directory.
 
-## Project Overview
-This is the Wheels CLI, a CommandBox module that provides command-line tools for Wheels framework applications.
+## What This Is
 
-## Commands
-- **Test**: `wheels test [type] [servername] [reload] [debug]`
-  - Types: core, app, plugin
-- **Run Application**: Use CommandBox's server start functionality
-- **Initialize**: `wheels init [name] [path] [reload] [version] [createFolders]`
+The Wheels CLI is the **LuCLI runtime rebranded as `wheels`** — it is NOT a CommandBox
+module. There is no separate `lucli` binary on a normal install; `libexec/wheels` IS the
+LuCLI binary, name-routed to the `wheels` module via its `$0` basename (`-Dlucli.binary.name=wheels`).
+Install is via Homebrew / Scoop / apt / yum (see `tools/distribution-drafts/` and the
+sibling distribution repos) — **not** `box install`.
 
-## Code Style Guidelines
-- **Language**: CFML (ColdFusion Markup Language)
-- **Architecture**: Component-based with CFC files inheriting from base.cfc
-- **Naming**: CamelCase for component names, functions, and variables
-- **Organization**:
-  - Commands in /commands directory, grouped by functionality
-  - Templates for code generation in /templates
-- **Error Handling**: Use try/catch blocks with appropriate error messages
-- **Code Generation**: Follow existing template patterns when modifying or creating new templates
+> Anything in older docs mentioning `box`, `CommandBox`, `box install wheels-cli`, `/cli/src/`,
+> `wheels init`, or `wheels g app` predates the LuCLI rebrand and is wrong. Scaffold with
+> `wheels new <name>`.
 
-## Development Notes
-- This is a CommandBox module that integrates with Wheels framework
-- Follow MVC pattern when creating new features
-- Maintain backward compatibility when possible
-- Refer to CLI-IMPROVEMENTS.md for planned enhancements and architectural goals
-- To run the CLI commands we need to launch CommandBox with the `box` command
-- The CLI is installed with `box install wheels-cli` command
-- Then use the `wheels` commands to run a particular CLI command
-- First create an app with the `wheels g app` command.
-- Then start the web server with `server start` commandbox command
+## Code Map
 
-## Monorepo Integration
+```
+cli/lucli/
+  Module.cfc          The CLI itself. Each PUBLIC function is a subcommand
+                      (generate, g, migrate, seed, test, reload, start, stop, new,
+                      create, routes, info, mcp, console, analyze, validate, destroy, d,
+                      doctor, deploy, packages, stats, notes, db, upgrade, browser,
+                      version, showHelp). Private functions are internal helpers.
+  module.json         Module manifest (name=wheels, main=Module.cfc).
+  ARCHITECTURE.md     Deep architectural reference — read this first.
+  services/           Service objects the subcommands delegate to:
+                        ArgSpec.cfc        typed, declarative argument/flag parser
+                        Scaffold.cfc       generate/scaffold/api-resource
+                        CodeGen.cfc        model/controller/view/test generation
+                        Templates.cfc      template rendering
+                        MigrationRunner.cfc, Analysis.cfc, Doctor.cfc, Stats.cfc,
+                        Destroy.cfc, ServerRegistry.cfc, ReleaseChannel.cfc,
+                        UpdateChecker.cfc, Helpers.cfc, SemVer.cfc, …
+                        deploy/            Kamal-compatible deploy port (DeployMainCli +
+                                           App/Proxy/Build/Registry/Lock/Prune/Secrets/Server CLIs,
+                                           config/, secrets/ adapters)
+                        packages/          package registry/install (Installer, Registry, …)
+  templates/          Code-generation templates (app scaffold, codegen, app tests).
+  tests/specs/        CLI test suite (commands/, services/, integration/, deploy/, packages/).
+```
 
-The Wheels CLI is part of a larger monorepo ecosystem:
+## Argument Parsing
 
-### CLI's Role in the Ecosystem
-- **Source**: `/cli/src/` contains CLI source code (commands, models, templates)
-- **Build**: `tools/build/scripts/build-cli.sh` packages CLI for distribution
-- **Distribution**: Published to ForgeBox as `wheels-cli` CommandBox module
-- **Integration**: Works with base templates and generates code following framework patterns
+Subcommands parse args via the **ArgSpec service** (`services/ArgSpec.cfc`) — a typed,
+declarative spec builder (`.option()`, `.flag()`, `.parse()`). Build the command's ArgSpec
+to enumerate its real flags/options/defaults. (Historical note: the legacy `getArgs()`
+argv round-trip was removed in the #2861/#2875 ArgSpec migration.)
 
-### Key Dependencies
-- **ForgeBox Integration**: Downloads `wheels-base-template` package from ForgeBox during `wheels g app`
-- **Template Snippets**: Uses base template snippets from `/app/snippets/`
-- **Core Patterns**: Generates code that follows core framework conventions (`$` prefix, `config()` methods)
-- **Version Sync**: Shares version numbers with other monorepo components
+LuCLI reserves some tokens before the module sees them: `--help`/`-h`, bare `help`,
+`--verbose`/`-v`, `--version`, and verbs like `run`/`install`/`mcp` and nested `server`/`secrets`.
+Be aware these can be intercepted by the runtime/launcher rather than reaching `Module.cfc`.
 
-### Development Workflow
-1. Modify CLI source in `/cli/src/`
-2. Test in monorepo `/workspace/` directory
-3. Reload CommandBox: `box reload` after changes
-4. Build process handles packaging and ForgeBox distribution via GitHub Actions
+## Running & Testing
 
-### Package Structure
-- `ModuleConfig.cfc` - CommandBox module configuration
-- `commands/wheels/` - Hierarchical command structure extending `base.cfc`
-- `models/` - Business logic with WireBox dependency injection
-- `templates/` - Code generation templates using `{{variable}}` syntax
-- `box.json` - Package metadata with `type: "commandbox-modules"`
+- Run a command against a checkout: the CLI loads the module from `$LUCLI_HOME/modules/wheels`.
+- CLI test suite: `bash tools/test-cli-local.sh` (boots a server, hits `/wheels/cli/tests`),
+  or the in-server endpoint `/wheels/cli/tests?format=json`. End-to-end lifecycle smoke:
+  `tools/test-cli-e2e.sh`. Deploy verb smoke (dry-run): `tools/deploy-verb-smoke.sh`.
+- Build/distribution: `tools/build/scripts/build-cli.sh` packages the module; the Homebrew/
+  Scoop/apt launchers stage `libexec/wheels` + the module zip and generate the `wheels` wrapper.
 
-For complete monorepo architecture details, see the main repository's `CLAUDE.md` file.
+## Code Style
+
+- **Language**: CFML. Components, CamelCase names. Internal helpers use a `$` prefix and
+  must be `public` only when they need to be reachable as mixins/specs — otherwise keep them
+  `private` so they don't surface as MCP tools.
+- **Cross-engine**: the CLI runs on the bundled Lucee; still avoid the engine traps in the
+  root `CLAUDE.md` (e.g. `(.+)` matching newlines in `reFind` — use `[^\r\n]+`).
+- **Error handling**: try/catch with actionable messages; throw typed errors for usage
+  failures so the runtime maps them to a non-zero exit.
+- **Code generation**: follow existing template patterns; strip CFML comments before any
+  source-scanning (anti-pattern #14).
+
+## MCP
+
+`wheels mcp wheels` launches the stdio MCP server. Tools are auto-discovered from `Module.cfc`
+public functions; stateful/interactive commands are hidden via `mcpHiddenTools()`. There is
+no `wheels mcp setup` command — write `.mcp.json` manually. See the root `CLAUDE.md`
+"CLI / MCP" section and `web/.../command-line-tools/mcp-integration` for details.
 
 ## Things to remember
-- Don't add the Claude signature to commit messages
-- Don't add the Claude signature to PR reviews
+- Don't add the Claude signature to commit messages.
+- Don't add the Claude signature to PR reviews.
--- a/cli/lucli/Module.cfc
+++ b/cli/lucli/Module.cfc
@@ -161,6 +161,7 @@ component extends="modules.BaseModule" {
 			"main",     // bare `wheels` no-args dispatch target — not an MCP tool
 			"mcp",      // meta command — prints MCP setup instructions
 			"d",        // alias for destroy
+			"g",        // alias for generate
 			"new",      // scaffolds a whole new Wheels project
 			"console",  // interactive CFML REPL — not usable over stdio
 			"start",    // dev server lifecycle (stateful)
@@ -266,7 +267,7 @@ component extends="modules.BaseModule" {
 		help &= "  generate            Generate model, controller, scaffold, migration, etc." & nl;
 		help &= "  destroy (or d)      Remove generated files" & nl & nl;
 		help &= "Database:" & nl;
-		help &= "  migrate             Run database migrations (latest, up, down, info, rename-system-tables)" & nl;
+		help &= "  migrate             Run database migrations (latest, up, down, info, doctor, forget, pretend, rename-system-tables)" & nl;
 		help &= "  seed                Run database seeds" & nl;
 		help &= "  db                  Database management (reset, status, version)" & nl & nl;
 		help &= "Testing & Inspection:" & nl;
@@ -391,7 +392,7 @@ component extends="modules.BaseModule" {
 	// ─────────────────────────────────────────────────
 
 	/**
-	 * hint: Run database migrations (latest, up, down, info)
+	 * hint: Run database migrations (latest, up, down, info, doctor, forget, pretend, rename-system-tables)
 	 */
 	public string function migrate() {
 		var args = new services.ArgSpec().toArgv(structuredArgs(arguments));
@@ -871,7 +872,7 @@ component extends="modules.BaseModule" {
 			out("Options:", "bold");
 			out("  --port=<number>           Server port (default: 8080)");
 			out("  --datasource=<name>       Datasource name (default: app name)");
-			out("  --reload-password=<pw>    Reload password (default: app name)");
+			out("  --reload-password=<pw>    Reload password (default: random)");
 			out("  --no-sqlite               Skip default SQLite database setup");
 			out("  --setup-h2                Use H2 embedded database instead of SQLite");
 			out("  --no-open-browser         Don't open browser on server start");
@@ -1086,7 +1087,10 @@ component extends="modules.BaseModule" {
 			// Count routes
 			var routesFile = variables.projectRoot & "/config/routes.cfm";
 			if (fileExists(routesFile)) {
-				var routeContent = fileRead(routesFile);
+				// Strip comments first so a commented-out .resources(...) isn't
+				// counted (anti-pattern #14 — commented code must not satisfy
+				// substring scans). Mirrors the datasource block above.
+				var routeContent = stripCfmlComments(fileRead(routesFile));
 				var resourceCount = 0;
 				var pos = 1;
 				while (pos > 0) {
@@ -1792,7 +1796,14 @@ component extends="modules.BaseModule" {
 	 * hint: Alias for destroy
 	 */
 	public string function d() {
-		return destroy();
+		return destroy(argumentCollection = arguments);
+	}
+
+	/**
+	 * hint: Alias for generate
+	 */
+	public string function g() {
+		return generate(argumentCollection = arguments);
 	}
 
 	// ─────────────────────────────────────────────────
@@ -2825,19 +2836,14 @@ component extends="modules.BaseModule" {
 
 		ensureDirectory(migrationDir);
 
-		// Use the DBMigrate template if available, otherwise inline
-		var templates = getService("templates");
-		var result = templates.generateFromTemplate(
-			template = "dbmigrate/blank.txt",
-			destination = "app/migrator/migrations/#fileName#",
-			context = {migrationName: migrationName}
-		);
-
-		if (!result.success) {
-			// Fallback to inline empty migration
-			var content = buildEmptyMigration(migrationName);
-			fileWrite(filePath, content);
-		}
+		// Always build the migration inline. The shipped codegen template
+		// dbmigrate/blank.txt carries |DBMigrateExtends|/|DBMigrateDescription|
+		// tokens that Templates.cfc never substitutes, so on packaged installs
+		// (where that template resolves) `generate migration` produced an
+		// uncompilable file (literal extends="|DBMigrateExtends|"). The inline
+		// builder emits a correct extends="wheels.migrator.Migration" body and
+		// was already the path every dev-checkout install used. See CLI audit H4.
+		fileWrite(filePath, buildEmptyMigration(migrationName));
 
 		printCreated("app/migrator/migrations/#fileName#");
 		return "";
@@ -3371,7 +3377,7 @@ component extends="modules.BaseModule" {
 			"seed-data": {
 				name: "Seed Data",
 				description: "Database seeding template with seedOnce() examples",
-				hint: "Run seeds with: wheels db:seed.",
+				hint: "Run seeds with: wheels seed.",
 				generate: function(string projectRoot, boolean force) {
 					var created = [];
 
@@ -3798,7 +3804,7 @@ component extends="modules.BaseModule" {
 			out("Migration Status", "bold");
 			out(repeatString("=", 70));
 
-			var fmt = "%-16s %-30s %-10s %s";
+			var fmt = "%-16s %-30s %-10s %-19s";
 			out(sprintf(fmt, "Version", "Description", "Status", "Applied"));
 			out(repeatString("-", 70));
 
@@ -5755,7 +5761,7 @@ component extends="modules.BaseModule" {
 		var envFile = variables.projectRoot & "/.env";
 		if (fileExists(envFile)) {
 			var envContent = fileRead(envFile);
-			var pwMatch = reFindNoCase("(?:WHEELS_)?RELOAD_PASSWORD\s*=\s*(.+)", envContent, 1, true);
+			var pwMatch = reFindNoCase("(?:WHEELS_)?RELOAD_PASSWORD\s*=\s*([^\r\n]+)", envContent, 1, true);
 			if (arrayLen(pwMatch.match) > 1 && len(trim(pwMatch.match[2]))) {
 				return trim(pwMatch.match[2]);
 			}
--- a/cli/lucli/services/Analysis.cfc
+++ b/cli/lucli/services/Analysis.cfc
@@ -354,8 +354,10 @@ component {
 			});
 		}
 
-		// Check mixed argument styles (uses same pattern as analyze)
-		if (reFindNoCase(variables.MIXED_ARGS_PATTERN, content)) {
+		// Check mixed argument styles (uses same pattern as analyze).
+		// Scan comment-stripped content so a commented-out association with
+		// mixed args doesn't false-positive (anti-pattern #14).
+		if (reFindNoCase(variables.MIXED_ARGS_PATTERN, activeContent)) {
 			arrayAppend(issues, {
 				file: arguments.path,
 				severity: "error",
@@ -393,7 +395,9 @@ component {
 
 	private array function validateRoutes(required string path) {
 		var issues = [];
-		var content = fileRead(arguments.path);
+		// Strip comments first so commented-out mapper()/.end()/.wildcard()/
+		// .resources() lines don't skew the balance/order checks (anti-pattern #14).
+		var content = $stripCfmlComments(fileRead(arguments.path));
 
 		// Check for balanced mapper/end
 		var mapperCount = 0;
@@ -443,10 +447,13 @@ component {
 		// Skip layouts and partials
 		if (fileName == "layout.cfm" || left(fileName, 1) == "_") return issues;
 
-		// Check for variable usage without cfparam
+		// Check for variable usage without cfparam. Scan comment-stripped
+		// content so a view whose only # / cfparam / cfset lives inside a
+		// CFML comment doesn't warn (or falsely appear protected) — #14.
+		var activeContent = $stripCfmlComments(content);
 		var hashChar = chr(35);
 		var cfsetTag = chr(60) & "cfset";
-		if (findNoCase(hashChar, content) && !findNoCase("cfparam", content) && !findNoCase(cfsetTag, content)) {
+		if (findNoCase(hashChar, activeContent) && !findNoCase("cfparam", activeContent) && !findNoCase(cfsetTag, activeContent)) {
 			arrayAppend(issues, {
 				file: arguments.path,
 				severity: "warning",
--- a/cli/lucli/services/Scaffold.cfc
+++ b/cli/lucli/services/Scaffold.cfc
@@ -437,8 +437,12 @@ component {
 				}
 			}
 
-			// 5. Update routes with API namespace
-			updateApiRoutes(arguments.name);
+			// 5. Update routes with API namespace.
+			// Use the PLURAL name so the route matches the plural controller
+			// (api/Products.cfc) and table (products) — mirrors updateRoutes(pluralName)
+			// on the non-api scaffold path. Passing singular arguments.name mapped
+			// /api/product and never reached the plural controller.
+			updateApiRoutes(pluralName);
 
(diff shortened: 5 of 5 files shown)


bpamiri:
## Full CLI Audit Report (for the record)

This PR lands the verified low-risk subset. Below is the **complete synthesis** from the audit (24 command groups, 157 confirmed findings → 69 unique after de-dup; adversarially verified). The Phase 3/4 items are tracked as follow-ups (see the PR description checklist).

<details><summary>Full report — findings &amp; prioritized fix plan</summary>


# Wheels CLI Audit — Findings & Fix Plan

## 1. Executive Summary

This audit covered the worktree Wheels CLI (LuCLI-rebranded) across 17 command groups, cross-checking each command's real behavior against its source (`cli/lucli/Module.cfc` + `cli/lucli/services/*.cfc`), its in-CLI help, the root `CLAUDE.md`, `cli/CLAUDE.md`, and the user-facing v4-0-0 guides. Findings were adversarially verified against worktree source and, where relevant, the shipped brew 4.0.2 install.

**Overall health:** The CLI's core read/write commands work, but a meaningful set of *documented* commands, flags, and aliases are non-functional, and many docs (especially the user-facing guides and the agent-facing root `CLAUDE.md`) describe behavior the CLI does not exhibit. Two systemic root causes dominate:

1. **LuCLI runtime interception** — `--help`/`-h`, the bare `help` token, `--verbose`/`-v`, `--version`, and reserved verbs (`run`, `install`, `mcp`, nested `server`/`secrets`) are consumed by the LuCLI runtime *before* the Wheels module sees them. This makes per-command help unreachable everywhere, breaks `--verbose` on `doctor`/`stats`/`test`, breaks `wheels test run`/`wheels browser install`, leaks the `lucli`/picocli brand to users, and crashes documented deploy flags at the picocli root.
2. **Argument round-trip / parsing gaps** — singular vs plural route names, dropped enum/`hasOne` attributes, space-form `--tag`, and several deploy flags that the service code reads but the parser never populates.

**Counts (CONFIRMED findings only; de-duplicated cross-cutting items counted once):**

| Severity | Bug | Doc | Total |
|---|---|---|---|
| HIGH | 7 | 6 | 13 |
| MED | 11 | 14 | 25 |
| LOW | 13 | 18 | 31 |
| **Total** | **31** | **38** | **69** |

Plus 4 pre-existing orchestrator findings (F-001…F-005) folded in below. Two raw findings were **refuted** during verification and are excluded (`migrate-4` reproduction mechanism wrong; `doctor-7` "dead code" claim false — the `-v` fallback fires after a `--` separator).

**Do all commands work as documented? No.** The most damaging classes: the `g` generate alias and the `d` destroy alias are advertised everywhere but one throws a raw exception and the other silently no-ops; `wheels reload`/`wheels console` auto-password is broken on every default scaffold; `generate migration` produces an uncompilable file on packaged installs; `--reporter=tap` crashes and `--ci`/exit codes never signal test failure; and the entire documented `wheels deploy` flag surface (`--version`, `--config`, nested `server`/`secrets` verbs, `app` filters) does not work as written.

---

## 2. Command-Group Status

| Group | Status | Worst issue |
|---|---|---|
| new / create | doc-drift + minor bug | In-CLI help wrong default; `create` unknown-type exits 0 |
| generate (core) | **bug** | `g` alias throws raw exception; enum dropped; silent view-gen failures |
| generate (db) | **bug** | `generate migration` leaks `\|DBMigrate*\|` tokens → uncompilable on packaged installs |
| generate (advanced) | **bug** | `api-resource` writes singular route mismatching plural controller |
| destroy | **bug** | `d` alias is a total no-op (drops all args) |
| migrate | doc-drift | `dbmigrate diff` & 4 real subcommands mis-documented; `migrate help` leaks lucli |
| db | bug + doc-drift | `db status` renders `%s`; `db:seed` documented but errors |
| seed | doc-drift | `generate seed` & `db:seed` documented, neither exists |
| test | **bug** | `--reporter=tap` crashes; `--ci`/exit always 0; `test run` & `browser install` are wrong verbs |
| routes | doc-drift | wrong endpoint + "verbatim" claim; silent exit-0 failures |
| info | bug + doc-drift | framework-version line never renders (stale path); route count counts comments |
| doctor | **bug** | `--verbose`/`-v` swallowed by LuCLI; sample output fabricated; over-claims scope |
| stats | **bug** | `--verbose`/`-v` dead (Top-10 never renders) |
| notes | bug + doc-drift | scanner matches in strings/identifiers, not comments; HACK default mismatch |
| analyze | doc-drift | `CLAUDE.md` says CLI is "—"; MCP `target` arg silently ignored |
| validate | **bug** | 3–4 sub-checks are comment-blind (false +/− against anti-pattern #14) |
| server-lifecycle | bug + doc-drift | `reload` broken on default scaffold; `wheels status` doesn't exist; foreground claim wrong |
| console | **bug** | auto-password broken (greedy `(.+)`); `/routes` always errors |
| mcp | bug + doc-drift | `mcp()` dead code; `mcp setup` documented but errors; leaks `$`-helpers + lucli brand |
| packages | bug + doc-drift | `--tag <v>` space-form dropped; docs say `reload` activates (cold restart needed) |
| upgrade | doc-drift | "no scan runs" false; advisory section + 2 checks undocumented |
| deploy | **bug** + doc-drift | `--version`/`--config`/nested verbs/`app` filters all broken as documented |
| version/help | **bug** | stable build reports `(development)`; `wheels help` leaks lucli; two `--version` formats |

---

## 3. CONFIRMED Findings by Severity

### HIGH

---

**H1 — `--verbose`/`-v`/`--version`/`--help`/bare `help`/reserved verbs are swallowed by the LuCLI runtime before the Wheels module runs (systemic).**
*type: bug · de-dups: doctor-1, stats-1, test-3, version-help-3, server-lifecycle-5, mcp-1/6, packages-1/3, browser-4/5, upgrade-4/5, routes-4, db-3, console-5, migrate-extra, F-003, and every per-command "--help prints the global banner" finding.*

Evidence (representative, all reproduced):
- `wheels doctor --verbose` / `-v` / `--verbose=true` on a WARNING app → no `Passed (` section; probe shows `arguments={}`. `--show-passed`/`--all`/`--bananas` DO arrive — proving `--verbose`/`-v`/`--debug` are LuCLI-reserved. (`parseVerboseFlag` Module.cfc:1808 always sees an empty struct.) Same defect breaks `wheels stats --verbose` (Top-10 never renders, Module.cfc:2430/2468) and `wheels test --verbose` (per-spec tree at Module.cfc:4567 never fires).
- `wheels <cmd> --help` and `wheels <cmd> -h` print the global banner for EVERY command (doctor/generate/migrate/seed/packages/upgrade/db/routes/console/browser/server/stats/validate/info verified byte-identical to `wheels --help`). The banner footer (Module.cfc:291) literally promises "For command-specific help: wheels <command> --help" — false for all.
- `wheels <cmd> help` (positional) leaks the raw picocli banner: `Usage: lucli [-dhtvw] ... 🚀 LuCLI - A terminal application with Lucee CFML integration` (verified for migrate, packages, browser, upgrade, analyze, test, routes). **The packages() `opts.help`/`sub=="help"` branch (Module.cfc:2245) and the in-source comment at Module.cfc:2238-2240 are dead/false** — none of `packages --help`/`-h`/`help` reaches `$packagesHelp()`.

proposedFix: **This must be fixed in the LuCLI dispatch/help layer, not per-command** (the per-command "mirror packages()" approach is proven non-functional). Two coordinated changes:
- (a) Route `<cmd> --help`/`-h`/`help` into the target module (or to a Wheels per-command help string) before picocli/builtin interception; rebrand any surfaced `lucli`/`🚀 LuCLI` strings to `wheels`. Until then, soften the global-help footer at `cli/lucli/Module.cfc:291` so it stops promising help it cannot deliver.
- (b) Expose LuCLI's global `--verbose`/`-v` state to modules (or rename the user-facing verbose triggers to non-reserved flags — `--show-passed` for doctor, `--top`/`--largest` for stats, `--details` for test). Fix once in `parseVerboseFlag` (Module.cfc:1808) and the test verbose path (Module.cfc:4507). Then update the guides (`app-inspection.mdx` doctor/stats `--verbose` rows, `testing.mdx:60`) to match.

---

**H2 — `g` alias for `generate` is unimplemented; `wheels g <type>` throws a raw Lucee exception leaking internals.**
*type: bug · de-dups generate-core/gen-1, generate-advanced/gen-2, gen-5.*

Evidence: `wheels g model Tester` / `g admin User` / `g scaffold Post` → `[ERROR] Error: lucee.runtime.exp.ExpressionException: Component [modules.wheels.Module] has no  function with name [g]` (exit 1). No `function g(` exists (grep = 0); `d()` exists at Module.cfc:1794. The alias is documented and *taught in the getting-started tutorial*: code-generation.mdx:26, quick-start.mdx:51, tutorial 02-first-model.mdx:100, 05-comments-streams.mdx:158, CLAUDE.md:725 & :732.

proposedFix: Add to `cli/lucli/Module.cfc` (mirror `d()` at 1794): `public string function g() { return generate(argumentCollection = arguments); }` and add `"g"` to `mcpHiddenTools()` (Module.cfc:159-170). This makes all the doc claims true. (Implementing is strongly preferred over deleting the doc claims, since the tutorial teaches it.)

---

**H3 — `d` alias for `destroy` is a total no-op: `d()` calls `destroy()` without forwarding `argumentCollection`, so every `d …` drops all args and deletes nothing.**
*type: bug · destroy-1; doc mirror destroy-3.*

Evidence: `d Sprocket`, `d model Sprocket --force`, `d controller Sprockets` all print the no-args usage banner (exit 0) and delete nothing — even `--force`. `destroy Sprocket` correctly previews. Root cause Module.cfc:1794-1796: `public string function d() { return destroy(); }` — no params, no forward; `destroy()` then reads its own empty `arguments` and falls to the usage banner. scaffold-cleanup.mdx:23 claims "Every example on this page works with either destroy or d."

proposedFix: `cli/lucli/Module.cfc:1795` — change `return destroy();` to `return destroy(argumentCollection = arguments);`. Verified in-worktree: preview AND `--force` deletion then work. This single-line fix also makes scaffold-cleanup.mdx:11/:23/:33 true.

---

**H4 — `generate migration <Name>` leaks unsubstituted `|DBMigrateExtends|`/`|DBMigrateDescription|` tokens on packaged installs, producing an uncompilable migration.**
*type: bug · generate-db/gen-1.*

Evidence: Shipped template `…/templates/codegen/dbmigrate/blank.txt` line 1 is `component extends="|DBMigrateExtends|" hint="|DBMigrateDescription|" {` (byte-identical to `cli/src/templates/dbmigrate/blank.txt`). `Templates.cfc::processTemplate` (66-187) substitutes only `{{…}}` keys and a fixed pipe-token set — never these two. Reproduced by mirroring the packaged codegen layout into the worktree: the generated `.cfc` started literally `component extends="|DBMigrateExtends|" …`, which cannot resolve as a component mapping → `wheels migrate latest` errors on instantiation. (In the symlinked harness the template doesn't resolve, so the clean inline fallback at Module.cfc:5522 runs and masks it — only the *packaged* path is broken.) **Blast radius is exactly `generate migration <Name>`**: `dbmigrate/blank.txt` has one caller (Module.cfc:2831); scaffold (Scaffold.cfc:615) and `generate property` (Module.cfc:2997-3017) build content inline and are unaffected.

proposedFix: Simplest — in `Module.cfc` generateMigration (2830-2834) drop the template path and call `buildEmptyMigration()` (Module.cfc:5522), which already emits a correct `extends="wheels.migrator.Migration"` body. (Or pass `DBMigrateExtends`/`DBMigrateDescription` into the context AND add their substitution to `Templates.cfc::processTemplate`.) Add a regression test asserting the generated file contains `extends="wheels.migrator.Migration"` and no literal `|` tokens.

---

**H5 — `generate api-resource <Model>` writes a SINGULAR route name, mismatching the plural controller, table, and its own documented/printed URL.**
*type: bug · generate-core/gen-2 = generate-advanced/gen-1; doc symptom generate-advanced/gen-3.*

Evidence: `generate api-resource Product …` → controller `app/controllers/api/Products.cfc` (plural), table `products`, success message `curl …/api/products.json` (plural), but `config/routes.cfm` gets `.resources(name="product", except="new,edit")` (SINGULAR). `.resources(name="product")` maps `/api/product`, never reaching the plural controller. Regular scaffold/route correctly emit plural. Root cause `Scaffold.cfc:441` passes raw `arguments.name`; `updateApiRoutes` (468) only `lCase`s. The codebase comment at Scaffold.cfc:148-153 documents this exact mistake (onboarding F4) for the non-api path.

proposedFix: `cli/lucli/services/Scaffold.cfc:441` — change `updateApiRoutes(arguments.name)` to `updateApiRoutes(pluralName)` (`pluralName` already computed at Scaffold.cfc:371), mirroring `updateRoutes(pluralName)` at Scaffold.cfc:155. Once plural, code-generation.mdx:210/218 and Module.cfc:3037/3069 become correct with no doc change. (Related MED, generate-advanced-extra: the *append* branch of `updateApiRoutes` (Scaffold.cfc:487-492) mangles indentation when a second api-resource is added to an existing `.namespace("api")` block — fix the insert-offset math and reuse the namespace indent + one tab; add a 2-sequential-generation spec.)

---

**H6 — `--reporter=tap` crashes mid-emission on any failing spec; `--ci` is inert and `wheels test` always exits 0 even on failures.**
*type: bug · test-1 + test-2.*

Evidence (tap): `wheels test --filter=models --reporter=tap` with a failing spec → emits `not ok 2`, then `[Test execution failed: Invalid call of the function [tapEscapeYaml] … Cannot cast Object type [Array] to a value of type [string]]`. `failOrigin` is an array of structs; Module.cfc:4442 keeps the array, and `if (len(t.failOrigin))` (4483, truthy for non-empty arrays) passes it to `tapEscapeYaml(required string value)` (4496). Evidence (ci/exit): `wheels test --filter=models; echo $?` → 0 with a failing spec; `--ci` output byte-identical to non-ci. `ciMode` is parsed (529/546/569) and declared (4315) but never read; `runTests` returns `""` unconditionally (4412). testing.mdx:61/:91 promise CI-mode exit tightening and non-zero exit on failure.

proposedFix:
- TAP: coerce `failOrigin` to a string in `emitTapResults` (Module.cfc:4442) — extract `sp.failOrigin[1].Raw_Trace` when it's an array of structs; belt-and-suspenders, guard line 4483 with `isSimpleValue(t.failOrigin)`. Add a TAP-reporter spec exercising a failing test with a stack origin.
- CI/exit: in `runTests` before the final `return ""` (~4407/4412), compute `totalFail + totalError` (with the `displayTestResults` fallbacks at 4516-4517) and `throw type="Wheels.TestsFailed"` when >0, mirroring the "Throw so LuCLI exits non-zero" pattern at Module.cfc:4685/4732. Document whether the throw is unconditional (matches :91) or gated on `ciMode`, and wire `ciMode` accordingly.

---

**H7 — `wheels test run` and `wheels browser install` are documented as canonical but are reserved-verb collisions that do the wrong thing.**
*type: doc · test-4 + test-5 + browser-1 + browser-2.*

Evidence: `wheels test run` → `Usage: lucli run SCRIPT …` (the LuCLI `run` builtin); `wheels test models` (non-reserved) correctly filters. `wheels browser install` → LuCLI's extension installer (`No git or extension dependencies to install`, downloads nothing); `wheels browser setup` is the real verb (Module.cfc:2688). A CI step `run: wheels browser install` installs nothing → browser specs stay `this.browserTestSkipped` → green-but-untested.

proposedFix:
- `wheels test run` → `wheels test`: CLAUDE.md:28, CLAUDE.md:728, and the genbase template comment `cli/lucli/templates/app/tests/runner.cfm:3`.
- `wheels browser install` → `wheels browser setup` (global replace) in `testing.mdx` (frontmatter line 3, intro line 11, heading 115, synopsis 122, examples 137/159, notes 125/143), `browser-tests.mdx` (27, 36), `ci-integration.mdx` (129, 249), `running-tests-locally.mdx` (201), and the single stale `07-testing-deploying.mdx:226` (do NOT blanket-replace that file — 165/171/294/301 are already correct). Add a one-line note that bare `install` hits LuCLI's extension installer and does nothing.

---

**H8 — `wheels reload` and `wheels console` auto-password detection is broken on every default scaffold (greedy `(.+)` captures across newlines on Lucee).**
*type: bug · server-lifecycle-2 = console-1; F-004 confirmed.*

Evidence: `detectReloadPassword` (Module.cfc:5758) uses `reFindNoCase("(?:WHEELS_)?RELOAD_PASSWORD\s*=\s*(.+)", envContent, 1, true)`. Verified on the actual Lucee engine: input with `WHEELS_RELOAD_PASSWORD=secret123\n…` captured `secret123\nWHEELS_LUCEE_ADMIN_PASSWORD=…` (the password PLUS the rest of the file); the `([^\r\n]+)` form returned exactly `secret123`. The scaffold `_env` (`cli/lucli/templates/app/_env`) puts `WHEELS_RELOAD_PASSWORD` on line 6 ABOVE `WHEELS_LUCEE_ADMIN_PASSWORD` (line 7), so `wheels reload` fails (`Failed to reload: Stream handler unavailable …`) and `wheels console` auto-connect fails (`Console connection failed: Invalid reload password …`) on a stock `wheels new` app.

proposedFix: `cli/lucli/Module.cfc:5758` — change the capture group `(.+)` to `([^\r\n]+)`, mirroring `$resolveAppTestDataSource` at Module.cfc:5807 which already uses `([^\r\n]+)`. This is the real fix (reordering `_env` would only mask it).

---

**H9 — `wheels deploy` documented flag surface is broken: nested `server`/`secrets` verbs are picocli-shadowed, `--version`/`--config` don't work, and `app` filters are dropped.**
*type: doc + bug · deploy-1, deploy-2, deploy-3, deploy-4, deploy-extra.*

Evidence (all reproduced):
- `deploy server bootstrap` / `deploy secrets print` → LuCLI's own picocli help (never the deploy module). Flat aliases work: `deploy bootstrap`, `deploy exec uptime`, `deploy print-secrets`. Module.cfc:2077-2084/2141-2151 confirm the nested form is retained only for MCP/programmatic callers. `.ai/wheels/deploy.md` already calls the flat aliases canonical — the user guides do not.
- `deploy --version=v1.2.3` / `deploy rollback --version=v1.2.3` → `Invalid value for option '--version': 'v1.2.3' is not a boolean` + leaks `lucli` banner (picocli root absorbs `--version`). Working forms: `deploy --release=v1.2.3` and positional `deploy rollback v1.2.3` (Module.cfc:1942 `opts.version = positional[2]`). `--release` is documented in ZERO guide pages.
- `deploy config --config=<path>` → ignored (falls back to default, errors); only `--configPath` is parsed (`DeployArgsParser.cfc:45-49`).
- The entire `deploy app <verb>` surface: `--version=<v>` crashes at picocli root, and `--role`/`--hosts`/`--follow`/`--container` are read by `DeployAppCli.cfc` (108/116) but never parsed by `DeployArgsParser.cfc` (grep count 0) → silently dropped. `deploy app boot --release=v1 --role=job` boots BOTH roles (no filtering).
- `deploy exec --cmd=`, `deploy extract-secrets --key=`, `deploy print-secrets --project-root=` document flags never parsed (exec/extract take positionals; print never reads project-root). The `secrets.mdx:118` pipe example (`echo … | … extract`) is doubly wrong — extract reads `--from`, not stdin.

proposedFix:
- Docs: replace nested forms with flat aliases throughout the deployment guides (first-deploy.mdx, secrets.mdx, migrating-from-kamal.mdx, deploy/index.mdx CardGrid, server/secrets index pages); document `--release=<v>` (note `--version` only works on packaged brew/scoop launchers that rewrite it); change rollback to positional `rollback <version>`; document exec/extract as positionals; drop `--project-root`/`--cmd`/`--key` rows.
- Code (preferred for `--config` and `app` filters): add a `--config` alias arm to `DeployArgsParser.cfc` (after line 49, set `opts.configPath`), and add `--role`/`--hosts`/`--follow`/`--container` arms (the service already reads them). Until parsed, remove the unparsed filter rows from the `app/*.mdx` guides.

---

**H10 — Shipped stable 4.0.2 reports `(development)`; `wheels version` and `wheels --version` emit two different formats; `wheels help` leaks the full LuCLI builtin help.**
*type: bug · version-help-1, version-help-2, version-help-3, version-help-extra.*

Evidence (shipped brew, same install): `wheels version` → `Wheels 4.0.2 (development)\nJava 21.0.11`; `wheels --version` → `Wheels Version: 4.0.2` + ASCII art + URL. The wrapper comment mandates parity (`/opt/homebrew/Cellar/wheels/4.0.2/bin/wheels:14-15`); they diverge. **Root cause of `(development)`**: the shipped `services/ReleaseChannel.cfc:48` reads literally `|| v == "4.0.2"` — the build's `@build.version@` token substitution clobbered the dev-sentinel string literal, so the release classifies its own version as a dev checkout BEFORE reaching the stable regex (line 68). `wheels help` (worktree AND shipped) → `Usage: lucli [-dhtvw] … 🚀 LuCLI … Repository: github.com/cybersonic/lucli` (no `function help(` exists; the bare token hits the LuCLI builtin). installation.mdx:216 falsely promises `wheels help` prints the Wheels banner.

proposedFix:
- ReleaseChannel: `cli/lucli/services/ReleaseChannel.cfc:48` — assemble the dev sentinel at runtime so the token replacer can't clobber it: `var devToken = "@" & "build.version" & "@"; … || v == devToken`. Verify a packaged MAJOR.MINOR.PATCH build classifies as `stable`.
- version parity: make the brew/scoop/apt wrapper `--version` block emit the same lines as `Module.cfc` `version()` (or route `--version` through the module). Then align installation.mdx:202-211 and release-channels.mdx:34/40 to the single format.
- `wheels help`: add a public `help()` → `showHelp()` to `Module.cfc` and `"help"` to `mcpHiddenTools()`; if the LuCLI builtin still shadows it, intercept bare `help` in the wrappers (the brew wrapper's `help)` case is gated on `$# -ge 2`, so single-arg `wheels help` falls through). Fix installation.mdx:216.

---

### MED

---

**M1 — `db status` renders the Applied/applied-at column as literal `%s`.**
*type: bug · db-1.* `Version … Status     %s` / `… applied    %s`. `fmt = "%-16s %-30s %-10s %s"` (Module.cfc:3801) ends in a width-less `%s`; the sprintf helper (Module.cfc:6420-6421) requires `\d+` width, so the bare `%s` arg is silently dropped. proposedFix: change the trailing `%s` to `%-19s` (fits ISO timestamps and the `-` placeholder). Do NOT loosen the regex — the stats table shares the helper. This also surfaces the existing empty-`appliedAt`→`-` guard (Module.cfc:3809) that is currently dead.

**M2 — `generate enum` property type is parsed but never emitted.**
*type: bug · generate-core/gen-4.* `generate model Pricing status:enum:draft,published,archived` → model has only `validatesPresenceOf` (no `enum()`), migration is a plain `t.string`. `parseGeneratorArgs` (Module.cfc:5577-5588) parses `prop.values` but `CodeGen.cfc`/`ModelContent.txt` never reference enum/values. proposedFix: implement — emit `enum(property=…, values=…)` in `CodeGen.generateModel`; or drop the parse branch; at minimum document the limitation.

**M3 — Controller/scaffold view-generation failures are silently swallowed.**
*type: bug · generate-core/gen-3.* `generate controller Buyers index show …` → exit 0, `app/views/buyers/` empty, no warning; scaffold reports "complete" with zero views. `Scaffold.cfc:117-131` and `Module.cfc:2783-2787` act only on `viewResult.success` (no `else`). Bites on the #1944 unbundled-templates failure mode. proposedFix: add an `else` branch appending `viewResult.error` to `results.skipped` (Scaffold) and printing a warning (Module.cfc controller loop).

**M4 — `generate test` silently overwrites an existing spec (no `--force`, no `fileExists` guard, prints "create").**
*type: bug · generate-db/gen-2.* Verified clobber via sentinel. `CodeGen.cfc generateTest` (214-283) has no force param/guard; contrast `generateHelper` (300-302). proposedFix: add `fileExists` guard + `boolean force=false` param mirroring `generateHelper`; thread `--force` through `Module.cfc generateTest` (2947-2974); print "overwrite" when forcing.

**M5 — `generate route` prints misleading "Could not find insertion point" when the route already exists in named-arg form.**
*type: bug · generate-db/gen-3.* With `.resources(name="orders", …)` and the marker present, `generate route orders` → "Could not find insertion point … Add manually: .resources("orders")". `generateRoute` dup-checks only the positional form (Module.cfc:2928); `Scaffold.updateRoutes` returns bare `false` for both the named-arg dup (Scaffold.cfc:317) and the genuine no-marker case. proposedFix: have `updateRoutes` return `{inserted, reason}` and print "Route already exists" for the duplicate case. (Also fix code-generation.mdx:232 which describes dup detection as positional-only.)

**M6 — Root `CLAUDE.md` documents non-existent `wheels generate seed` / `--all` and `wheels db:seed` (both error/no-op).**
*type: doc · generate-db/gen-4 = seed-1; db-2 = seed-2.* `generate seed` → "Unknown generator type: seed" (exit 0, no file); `db:seed` → raw exception (exit 1). No `seed` case in the generate dispatcher (Module.cfc:343-381); `db()` (2542-2576) only routes reset/status/version; no colon dispatch exists. The `generate snippets seed-data` snippet hint *also* prints the dead `wheels db:seed` (Module.cfc:3374). proposedFix: in CLAUDE.md, remove lines 621-622 (`generate seed`) and the `db:seed` clauses (625, 733). Fix the snippet hint Module.cfc:3374 → `wheels seed`. (Note the snippet writes `app/snippets/seeds*.cfm`, not the seeder-convention `app/db/seeds.cfm` — document the relocation or add a real `seed` generator targeting `app/db/`.)

**M7 — `migrate` docs: `dbmigrate diff` doesn't exist; database.mdx omits 4 real subcommands; 3 in-code descriptions disagree; `migrate help` leaks lucli.**
*type: doc · migrate-1, migrate-2, migrate-3, migrate-extra; F-005.* `wheels dbmigrate diff Widget` → raw exception (CLAUDE.md L591-597). database.mdx lists only latest/up/down/info and claims doctor/forget/pretend/rename-system-tables "exit without touching the schema" — all four work (Module.cfc:400-440). Module.cfc:269/394/438 give three different action lists; `wheels --help` surfaces the incomplete L269 form. proposedFix: delete the `dbmigrate diff` bash block in CLAUDE.md (keep the CFC `AutoMigrator` examples); add the 4 subcommands to database.mdx and soften the "unknown action" claim; normalize Module.cfc:269/394 to the 8-action list at :438. The `migrate help` lucli leak is covered by H1.

**M8 — `wheels info` framework-version line never renders (stale path) + route count includes commented `.resources()`.**
*type: bug · info-1 + info-2; doc info-extra.* `wheels info` jumps from "Project:" to "Engine:" — `versionFile = …/vendor/wheels/events/onapplicationstart/settings.cfm` (Module.cfc:1045) no longer exists; authoritative source is `vendor/wheels/BuildInfo.cfc`. Separately, the route counter reads raw `fileRead` (Module.cfc:1089) and counts commented `.resources(` (verified 3 vs 2), violating anti-pattern #14 — while the datasource block 23 lines up (1066) strips comments. proposedFix: replace the stale read with `CreateObject("component","wheels.BuildInfo").version()` (Module.cfc:1044-1054); change Module.cfc:1089 to `stripCfmlComments(fileRead(routesFile))`; fix app-inspection.mdx:64/74/98.

**M9 — `wheels info` doc says output goes to stderr (it's stdout); `2>&1` advice is backwards.**
*type: doc · info-3.* `info 1>/dev/null` → empty; `info 2>/dev/null` → full summary. `out()` uses `System.out.println`. proposedFix: rewrite app-inspection.mdx:82-84 to say STDOUT; delete the `2>&1`/stderr instruction.

**M10 — `doctor` docs: sample output fabricated; scope over-claimed; standalone command absent from CLAUDE.md.**
*type: doc · doctor-2 + doctor-3 + doctor-4.* The app-inspection.mdx:212-232 sample (`.env file missing`, `vendor/wheels/ installed`, `lucee.json present`, `_env template`) matches no real `Doctor.cfc` string; real HEALTHY output is a `Passed (N)` list (`Required directory exists: app/`, `Datasource configured in config/settings.cfm`, etc.). Doctor does NO DB connection and NO `wheels_migrator_versions` read (that's the separate `wheels migrate doctor`), contradicting "datasource reachability, migrator table state" (app-inspection.mdx:190). The standalone `wheels doctor` is missing from the CLAUDE.md Development Tools table (723-733). proposedFix: replace the sample with verbatim real output; correct line 190 to "config presence (string match), migration file count … all local file ops"; add a `| Health check | wheels_doctor() | wheels doctor |` row to CLAUDE.md.

**M11 — `notes` HACK mismatch + scanner matches in strings/identifiers, not comments.**
*type: doc + bug · notes-1 + notes-2 (+ notes-extra).* CLI help (Module.cfc:282) and mcp-integration.mdx:75 list HACK, but the default is `TODO,FIXME,OPTIMIZE` (Module.cfc:2490, Stats.cfc:116); the canonical app-inspection.mdx omits HACK. Separately, `Stats.cfc:237` (`aType & "[\s:]+(.*)"`) is applied to raw lines with no comment detection (its own docstring at :236 claims comment-context) and no leading word boundary — verified false positives on string literals (`"…TODO list…"`) and identifiers (`methodTODO`), with trailing `";` leaking in. proposedFix: pick one canonical HACK default and align all three surfaces; restrict the scanner to comment context (borrow the block-comment state machine at Stats.cfc:166-222) and anchor the token with `\b`; fix the docstring at Stats.cfc:236.

**M12 — `analyze` is mis-documented in CLAUDE.md (CLI "—" is false; MCP `target` arg silently ignored).**
*type: doc · analyze-1.* `wheels analyze models` works (Files: 1); `analyze --target=models` ignores the named target (Files: 6). `parseAnalyzeArgs` binds only `arg1` (Module.cfc:1533); ArgSpec ignores undeclared named keys. proposedFix: CLAUDE.md:731 → `| Analyze | wheels_analyze() (whole project) | wheels analyze [target] (all\|models\|controllers\|views) |`; drop the `target="all"` named form. (If a named target is wanted for MCP, declare `.option(name="target")` and OR it into `hasTarget`.)

**M13 — `validate` ships comment-blind sub-checks (false +/−), contradicting anti-pattern #14 which holds it up as the exemplar.**
*type: bug · validate-1 + validate-2 + validate-3 + validate-extra; doc validate-4, validate-5.* Verified: a commented `// hasMany("comments", dependent="delete");` triggers a false ERROR (`Analysis.cfc:358` scans raw `content`, while the sibling extends check at 349 uses stripped `activeContent`); a tag-comment-only `<!--- #posts# --->` triggers a false cfparam WARNING (validateView, 449); a commented `// .end()` masks a real unclosed block AND a commented `mapper()` inflates the count (validateRoutes, 402-410); a commented `.resources()` after a live `.wildcard()` false-positives (426). proposedFix: `Analysis.cfc:358` → use `activeContent`; in `validateView` (438-455) compute `activeContent = $stripCfmlComments(content)` and test against it; in `validateRoutes`, read `content = $stripCfmlComments(fileRead(...))` at line 396 (covers mapper/.end()/.wildcard()/RESOURCE_AFTER_WILDCARD). Docs: code-quality.mdx:95 drop "public filter functions" (that's analyze-only); after the fixes, reconcile CLAUDE.md:219.

**M14 — `wheels start` doc says foreground/blocking; it's background. `wheels status` and `wheels_server` MCP tool don't exist.**
*type: doc · server-lifecycle-1 + server-lifecycle-4.* `wheels start` returns immediately (background); dev-server.mdx:58/60-62 say it blocks the terminal. `wheels status` → raw exception (no `status()`/`server()` function); CLAUDE.md:730 documents `wheels start|stop|status` + `wheels_server(action="status")`. proposedFix: rewrite dev-server.mdx:58/60-62/82/125-134 to background semantics (point to `wheels stop` / `wheels server run` for foreground); CLAUDE.md:730 → drop `status` from CLI column, replace the MCP cell with `—`.

**M15 — `reload` silently ignores `--password` and has no project/identity guard (can reload an unrelated server on a common port).**
*type: bug · server-lifecycle-3.* `reload()` (Module.cfc:621-654) never reads `--password` and calls `$requireRunningServer()` with the default `requireProjectConfig=false`. From a non-project dir, `wheels reload` reloaded an unrelated server on :8080 via the common-port fallback. proposedFix: honor a `--password` override (or stop implying one exists), add a `$isWheelsProjectDir` guard mirroring start()/stop(), and call `$requireRunningServer(requireProjectConfig=true)` (reload mutates app state via `applicationStop()`, so it belongs in the #2878 write-side identity class).

**M16 — `packages` docs say `wheels reload` activates an install; the CLI itself says `wheels stop && wheels start`.**
*type: doc · packages-4.* `packages add wheels-basecoat` prints "Run `wheels stop && wheels start` to activate it." (PackagesMainCli.cfc:211-217, cites onboarding F14 — reload doesn't re-fire `onApplicationStart`). Five doc spots claim reload activates: CLAUDE.md:495; install.mdx:10/54/58; update.mdx:35; index.mdx:9. proposedFix: replace each reload-activation reference with the cold-restart form (verbatim CLI string for the two example-output blocks).

**M17 — `mcp setup` documented in CLAUDE.md but errors; `mcp()` is dead code; tools/list leaks two `$`-prefixed internal helpers.**
*type: doc + bug · mcp-2 (=F-002), mcp-1, mcp-3.* `wheels mcp setup` → `mcp: module not found: 'setup'`; the only `mcp setup` artifact is the dead CommandBox `cli/src/commands/wheels/mcp/setup.cfc`. `Module.cfc::mcp()` (1138) has zero callers (the LuCLI builtin shadows the verb). `wheels mcp wheels --once tools/list` exposes `$normalizetestfilter` and `$resolveapptestdatasource` as callable tools — both declared `public` (Module.cfc:586, 5799), violating the in-repo warning at Module.cfc:5659-5664. proposedFix: CLAUDE.md:711 — rewrite to "manually create `.mcp.json`/`.opencode.json`; no generated-file command exists" (matches mcp-integration.mdx:40). Make `$normalizeTestFilter`/`$resolveAppTestDataSource` `private` (use TestBox `makePublic()` in specs if needed). Route bare `wheels mcp`/`mcp --help` to `Module.cfc::mcp()` or delete it (covered partly by H1).

**M18 — `wheels packages help` leaks lucli; `$packagesHelp()` unreachable.**
*type: bug · packages-1 + packages-3.* Covered by H1 (de-dup); the false in-source comment at Module.cfc:2238-2240 should be corrected or the dead `sub=="help"` branch (2245) removed.

**M19 — `upgrade` docs: "no scan runs" on same-major is false; the advisory "Recommended Improvements" section + 2 checks are undocumented; RateLimiter mis-bucketed.**
*type: doc · upgrade-1 + upgrade-2 + upgrade-extra.* `upgrade check --to=4.0.5` (from 4.0.0) prints "Same major … no known breaking changes" THEN "Scanning for opt-in recommendations…" + a cyan "Recommended Improvements (N found):" section (the same-major branch at Module.cfc:3927-3931 doesn't return; advisory checks run unconditionally). upgrade.mdx:33 says "No scan runs"; grep finds 0 hits for Recommended/advisory/paginationLinks/t.references; the RateLimiter check (doc line 106) is `severity:'advisory'` (Module.cfc:4001), not breaking. proposedFix: rewrite upgrade.mdx:33; add a "Recommended Improvements (advisory)" subsection covering paginationLinks→paginationNav and the t.references underscore opt-in; update the same-major sample (83-89); annotate RateLimiter as advisory.

**M20 — `deploy --help`/`<sub> --help` print global help; `server exec`/`secrets` flags mis-documented.**
*type: bug + doc · deploy-5 + deploy-6.* Covered by H1 (help) and H9 (flags) — listed for completeness.

**M21 — `browser` help leaks/global-banner; `browser test` subcommand+flags undocumented; `--browser=firefox|webkit` presented as usable but DSL is chromium-only.**
*type: bug + doc · browser-3 + browser-4 + browser-5 + browser-6.* `browser --help`/`help` → global banner / lucli leak (H1). `browser test` (`--verbose`/`-v`/`--format=text|json`/`--directory=`, Module.cfc:6200-6222) is entirely undocumented (testing.mdx:125 mentions only that it exists). browser-tests.mdx:39 presents firefox/webkit as installable engines, but `this.browserEngine = "chromium"` is the only wired engine (browser-testing.md:11, testing.mdx:132). proposedFix: add a `### wheels browser test` subsection (note it requires a running server and only pre-flight-checks JARs, never auto-installs); reword the `--browser` line to "only chromium is wired into the DSL today."

**M22 — Brew wrapper drift + wrong `create --help`; `create` missing from `--help` listing.**
*type: bug + doc · version-help-4 + version-help-5 + version-help-7.* Wrapper top-level `--help` drifted from `showHelp()` on migrate/packages/upgrade descriptions (wrapper 66/83/84 vs Module.cfc 269/284/285). `wheels create --help` describes "deployment scaffolding" (wrapper 433-439) but `create()` only forwards to `wheels new`. `create` is a live MCP tool (`wheels_create`) and a working CLI command but is absent from `showHelp()`/wrapper top-level `--help`. proposedFix: regenerate the wrapper help block from `showHelp()` (or have the wrapper shell out to the module); fix the `create --help` stub; add a `create` line to `showHelp()` (Module.cfc:253-293) — or, if `create` should be hidden, add it to `mcpHiddenTools()`.

---

### LOW (grouped — fix opportunistically alongside the above)

- **In-CLI / wrapper text fixes** *(new-create-2, console-6, version-help-8, version-help-9)*: `wheels new` usage says reload-password default is "app name" — actually random (Module.cfc:874). Console connect-error says "Wheels 3.x" — framework is 4.x (Module.cfc:1205). Wrapper `version --help` claims "Same as wheels --version" — false (resolved by H10). installation.mdx:209 channel label `(snapshot)` should be `(bleeding-edge)` (ReleaseChannel.cfc:63-64).
- **In-product help/text mismatches** *(console-3, console-4)*: `/env` returns a struct (environment/version/urlRewriting/datasource), not just the environment string (console-and-repl.mdx:62). In-product `/help` omits `/datasource` and `/q` aliases (printConsoleHelp, Module.cfc:1501/1504).
- **Doc sample-output drift** *(routes-1, routes-2, analyze-3, validate-6, stats-2, stats-3, gen-6, upgrade-3, deploy-7, deploy-8, new-create-extra)*: `wheels routes` doc names the wrong endpoint (`/wheels/ai?context=routing`) and falsely claims "verbatim" output, with a wrong sample layout (real: uppercase 3-col + inline `(name)`, Module.cfc:1009-1023). code-quality.mdx sample messages are paraphrases. mcp wheels_stats description says "routes" (Stats counts none) and omits views/helpers/config; stats sample shows wrong category set/order. code-generation.mdx shows named-arg `belongsTo(name="author")` but the generator emits positional `belongsTo('author')` (mdx:67/189/398). upgrade.mdx:31 says box.json only — source prefers wheels.json. deploy init.mdx omits Dockerfile/.dockerignore; config.mdx claims Mustache interpolation (runtime uses `${VAR}`). creating-a-project.mdx layout-comment omits `db/test.sqlite`.
- **Silent exit-0 on error paths** *(new-create-4, generate-core-extra, routes-3, doctor-6, browser-8, migrate-5, db-5)*: `create <unknown-type>`, `generate <unknown-type>`, `generate view` missing-template, `routes` parse/fetch failures, `doctor` CRITICAL, all `browser` error paths, `migrate` user-errors, and `db reset`'s un-wrapped seed phase all return `""` → exit 0 (or surface a raw error). Scripts/CI cannot detect failure. Consider routing user-error/failure paths through a typed throw that LuCLI maps to a non-zero exit (verify whether this is a systemic LuCLI return-string convention first).
- **Real-but-undocumented / discoverability** *(generate-core/gen-8, test-extra, server-lifecycle-6, packages-5, packages-extra)*: `--hasOne` works for `generate model` but is dropped by scaffold/api-resource. `--directory` is a real test alias for `--filter` (Module.cfc:491) but undocumented. `wheels start -f/--force` is wheels-side but absent from dev-server.mdx. CLAUDE.md package block omits the real `registry info` verb. packages index.mdx registry links (`./registry-refresh`, `./registry-info`) are broken (files live under `registry/`).
- **Stale `cli/CLAUDE.md`** *(F-001, new-create-5, test-7, destroy-extra)*: the entire file describes the defunct CommandBox CLI (`box install wheels-cli`, `/cli/src/`, `wheels init`, `wheels g app via ForgeBox`, `wheels test [type] [servername] …`). Rewrite or remove wholesale; fold the per-command stale slices into that rewrite.
- **Upstream / packaging** *(server-lifecycle-7, packages-6, mcp-4, mcp-6)*: LuCLI `server restart` help is a copy-paste of `status` (upstream LuCLI, not Module.cfc). `mcp()`/`McpServer.cfc:11` reference a non-existent doc path (`mcp-configuration-guide.md`); repoint to `command-line-tools/mcp-integration`. The packages "add" guide lives at `install.mdx` (slug `/install`) — rename to `add.mdx`. mcp-integration.mdx:21/58 + CLAUDE.md:713 say MCP tools are `wheels_<function>` but tools/list emits bare names. The `lucli`/serverInfo brand leaks are part of H1.

---

## 4. Prioritized Fix Plan

Docs first (low-risk), then code. `[code]`/`[doc]` and `[low-risk]`/`[needs-confirmation]` marked per item.

### Phase 1 — Doc fixes (low-risk, high-value; correct what users/agents are told)

1. `[doc][low-risk]` **Root `CLAUDE.md` corrections** — remove `dbmigrate diff` block (M7), `generate seed`/`db:seed` (M6), `wheels status`/`wheels_server` (M14); fix `analyze` row (M12); add `wheels doctor` row (M10); fix `g`-alias claims (H2 — leave true if H2 code lands); reload-activation (M16); `mcp setup` (M17); `wheels test run`→`wheels test` (H7). One PR.
2. `[doc][low-risk]` **`wheels test run`→`wheels test` + `wheels browser install`→`wheels browser setup`** across testing/CI guides and the genbase template comment (H7).
3. `[doc][low-risk]` **`deploy` guides** — flat aliases, `--release`, positional rollback, `--configPath`/`--config`, exec/extract positionals, drop unparsed `app` filter rows, init Dockerfile/.dockerignore, Mustache→`${VAR}` (H9, deploy-7/8).
4. `[doc][low-risk]` **`app-inspection.mdx`** — doctor sample + scope (M10), info stderr→stdout (M9), info framework-version intro (M8), routes endpoint/sample (routes-1/2), stats sample/order (stats-3), notes HACK (M11 doc half).
5. `[doc][low-risk]` **`upgrade.mdx`** — same-major scan, advisory section, RateLimiter bucket, wheels.json source (M19, upgrade-3).
6. `[doc][low-risk]` **`testing.mdx` / `browser-tests.mdx`** — document `browser test`, chromium-only engine, `--directory` alias, per-user install dir + `WHEELS_BROWSER_HOME` (M21, test-extra, browser-7).
7. `[doc][low-risk]` **`dev-server.mdx`** — background semantics, `-f/--force` flag (M14, server-lifecycle-6).
8. `[doc][low-risk]` **`code-generation.mdx`** — `belongsTo` form, dup-detection named-arg note, test/migration template descriptions (gen-6, generate-db-extra, gen-5). **`code-quality.mdx`** — drop "public filter functions" from validate, fix sample messages (M13 doc, validate-6). **`console-and-repl.mdx`** — `/env` struct (console-3). **`creating-a-project.mdx`** — drop `helpers/`, add `db/test.sqlite` (new-create-1, new-create-extra). **`mcp-integration.mdx`** — wheels_stats wording, tool-name prefix note (stats-2, mcp-5). **packages guides** — fix broken registry links, rename install.mdx→add.mdx (packages-extra, packages-6). **CLAUDE.md** — add `registry info` (packages-5).
9. `[doc][low-risk]` **Rewrite/remove `cli/CLAUDE.md`** wholesale (F-001 et al.).
10. `[doc][low-risk]` **`installation.mdx`/`release-channels.mdx`** — version format + `wheels help` claim, channel labels (H10 doc half, version-help-9).

### Phase 2 — Low-risk one-line code fixes (clear root cause, verified)

11. `[code][low-risk]` **`d` alias** — Module.cfc:1795 `return destroy(argumentCollection = arguments);` (H3, verified).
12. `[code][low-risk]` **reload/console password regex** — Module.cfc:5758 `(.+)`→`([^\r\n]+)` (H8, verified on Lucee).
13. `[code][low-risk]` **`db status` format** — Module.cfc:3801 trailing `%s`→`%-19s` (M1).
14. `[code][low-risk]` **info route-count comment strip** — Module.cfc:1089 wrap in `stripCfmlComments()` (M8).
15. `[code][low-risk]` **validate comment-blindness** — `Analysis.cfc:358` use `activeContent`; `validateView` strip; `validateRoutes` strip at :396 (M13, verified).
16. `[code][low-risk]` **`g` alias** — add `g()`→`generate(argumentCollection=arguments)` + mcpHiddenTools (H2, verified parallel `d()` works).
17. `[code][low-risk]` **MCP `$`-helper leak** — make `$normalizeTestFilter`/`$resolveAppTestDataSource` private (M17).
18. `[code][low-risk]` **snippet hint** — Module.cfc:3374 `db:seed`→`seed` (M6).
19. `[code][low-risk]` **migrate description sync** — Module.cfc:269/394 → 8-action list (M7).
20. `[code][low-risk]` **console "3.x" text** — Module.cfc:1205; **new usage default text** — Module.cfc:874 (LOW).

### Phase 3 — Code fixes needing care/regression tests

21. `[code][needs-confirmation]` **`generate migration` token leak** — route through `buildEmptyMigration()` (or fix Templates substitution); add regression test asserting `extends="wheels.migrator.Migration"` + no `|` tokens (H4 — only manifests on packaged installs, so verify against a packaged build).
22. `[code][needs-confirmation]` **`api-resource` plural route** — Scaffold.cfc:441 `updateApiRoutes(pluralName)`; also fix the append-branch indentation (Scaffold.cfc:487-492); add 2-sequential-generation spec (H5).
23. `[code][needs-confirmation]` **TAP reporter + `--ci`/exit codes** — coerce `failOrigin`; throw on failures; wire `ciMode`; add a failing-spec TAP test (H6).
24. `[code][needs-confirmation]` **ReleaseChannel dev-sentinel** — assemble `@build.version@` at runtime so the build token-replacer can't clobber it; verify a packaged stable build classifies as `stable` (H10).
25. `[code][needs-confirmation]` **DeployArgsParser** — add `--config` alias + `--role`/`--hosts`/`--follow`/`--container` arms (the service already reads them) (H9 code half).
26. `[code][needs-confirmation]` **silently-swallowed view-gen** + **generate test overwrite guard** + **route dup messaging** + **enum emission** + **scaffold/api-resource `--hasOne`** + **notes comment-context scanner** (M2/M3/M4/M5/M11 code half, gen-8).
27. `[code][needs-confirmation]` **`reload` project/identity guard** — `$isWheelsProjectDir` + `requireProjectConfig=true` + honor `--password` (M15; touches the #2878 write-side identity model — confirm threat-model alignment).

### Phase 4 — Systemic LuCLI dispatch work (highest leverage, highest risk; likely upstream LuCLI changes)

28. `[code][needs-confirmation]` **Help dispatch** — route `<cmd> --help`/`-h`/`help` into the module (or a Wheels per-command help string) before picocli/builtin interception; rebrand surfaced `lucli`/`🚀 LuCLI` strings; or, minimally, soften the global-help footer (Module.cfc:291). Resolves the per-command-help half of H1 and all `<cmd> help` lucli leaks (M7/M17/M18/M20/M21, F-003).
29. `[code][needs-confirmation]` **Verbose forwarding** — expose LuCLI global `--verbose`/`-v` to modules, or rename the doctor/stats/test verbose triggers to non-reserved flags; fix `parseVerboseFlag` + `test()` once. Resolves the verbose half of H1 (doctor-1/stats-1/test-3) and the self-referential `Doctor.cfc:375` "run --verbose for details" instruction.
30. `[code][needs-confirmation]` **`--version` / `wheels help` / unknown-command** — make the launcher not absorb `--version` before the deploy parser; add a `help()` function and/or wrapper interception for bare `help`; add `onMissingMethod()` so unknown commands print a branded "Unknown command" instead of leaking `modules.wheels.Module` (H9 version flag, H10 help, version-help-6).
31. `[code][low-risk]` **`packages --tag <value>` space-form** — give packages a real ArgSpec typing `tag` as a value-bearing option so the launcher captures `--tag ui` (packages-2).

Phase 4 items are cross-cutting and should be scoped as a single LuCLI-dispatch workstream; many Phase 1 doc fixes (help promises, lucli leaks) become unnecessary if Phase 4 lands, but the docs should be corrected now regardless since the dispatch fix is non-trivial.
</details>

PR #2883, opened 2026-06-09
fix(cli): CLI audit follow-ups (test exit codes, deploy flags, release channel, guide fixes)

bpamiri:
## What & why

Follow-up to #2882 — the next tier of CLI-audit fixes (the deferred Phase 3 items you asked me to tackle), plus a user-facing guide sweep. Every code fix was reproduced and re-verified against the CLI in an isolated harness; every doc edit was reviewed against verified behavior and for MDX/JSX safety.

## Code fixes (`cli/lucli/**`)

| Fix | Was | Now |
|---|---|---|
| **`wheels test` exit code** | always exited `0`, even with failing specs (CI green-but-failing) | exits **non-zero** when specs fail/error; a fully-passing run still exits 0 (verified both ways) |
| **`--reporter=tap`** | crashed on the first failing spec (`failOrigin` is an array → "Cannot cast Array to string") | `$tapOriginString()` coerces it; emits well-formed TAP (verified) |
| **`deploy --config`** | only `--configPath` parsed; guides document `--config` | `--config` accepted as an alias (verified) |
| **`deploy app --role/--container/--follow`** | parsed nowhere → `app boot --role=web` booted **all** roles | parser populates them; role filtering works (verified: `--role=web` → web only) |
| **ReleaseChannel** | the build's `@build.version@` token-replacer clobbered the dev-sentinel string literal → stable builds reported `(development)` | sentinel assembled at runtime so the replacer can't see a literal token |

## Doc sweep (`web/sites/guides/v4-0-0/**`, 23 files)

Corrected guides that documented commands/flags/verbs that error or behave differently:
- **deploy**: `--version=<v>` → `--release=<v>` (picocli root absorbs `--version`); `--configPath` alias noted; `rollback <version>` positional; nested `deploy server`/`secrets <verb>` → flat aliases (`bootstrap`, `exec`, `*-secrets`).
- **testing/CI**: `wheels browser install` → `wheels browser setup`; `wheels test run` → `wheels test`; non-zero-exit-on-failure noted; chromium-only engine; documented `wheels browser test`.
- **wheels-commands**: `wheels info` → STDOUT (not stderr); `wheels start` runs in the **background** + `-f/--force`; all 8 `migrate` subcommands documented (incl. doctor/forget/pretend/rename-system-tables); reload-password is random; no `app/helpers/` dir; `belongsTo` generator emits positional form; `/env` returns a struct; upgrade same-major still runs the advisory scan.
- **packages**: a newly-installed package needs a cold restart (`wheels stop && wheels start`), not `reload`; fixed broken `registry/` links; `add` (not `install`) wording.
- **installation**: `wheels help` falls through to LuCLI's generic help, not the Wheels banner.

## Verification
- Code fixes reproduced + re-verified in an isolated worktree-source harness (exit codes via direct `$?`, TAP output, `--role` dry-run filtering, `--config` load).
- MDX reviewed for JSX safety — fixed one `LinkCard` description that had a JSX-invalid `\"` escape; all other angle-brackets are inside code fences / inline-code / valid `<Aside>` components.
- The full `tools/test-cli-local.sh` suite can't run from a `.claude/worktrees` worktree (framework demo app won't stay up there); **CI is the backstop** (CLI suite + deploy verb-smoke + docs build).

## Still deferred (upstream / out of this repo)
The remaining systemic items live in the **LuCLI runtime** (`cybersonic/LuCLI`) + the per-launcher wrapper, not in `cli/lucli/`, so they need upstream + Homebrew-formula PRs:
- [ ] Per-command `--help`/`-h`/bare `help` all print the global/LuCLI banner (preprocessModuleHelp interception).
- [ ] `--verbose`/`-v` swallowed by the runtime (breaks `doctor`/`stats`/`test` verbose).
- [ ] `--version` picocli root shadowing (the launcher should rewrite `--version`→`--release` for deploy); `wheels help`/`mcp`/error paths leak the `lucli` brand.
- [ ] Per-command help text is hard-coded in each launcher wrapper and drifts.

(Lower-priority in-repo nits — `generate enum` not emitted, silent view-gen failures, `generate test` overwrite guard, `notes` comment-context scanner — remain as smaller follow-ups.)

### DIFF
--- a/cli/lucli/Module.cfc
+++ b/cli/lucli/Module.cfc
@@ -4364,6 +4364,8 @@ component extends="modules.BaseModule" {
 			out("Scope: #filter#", "cyan");
 		}
 
+		var testsFailed = false;
+
 		try {
 			var testUrl = "http://localhost:#serverPort##testPath#?format=#format#&db=#db#";
 			// App tests default to running against the <appname>_test
@@ -4401,6 +4403,11 @@ component extends="modules.BaseModule" {
 					default:
 						displayTestResults(result, verboseOutput, resolvedDir);
 				}
+
+				// Record failure so the command can exit non-zero AFTER the output
+				// is flushed. Throwing here would be swallowed by the catch below.
+				// testing.mdx documents a non-zero exit on failure. CLI audit H6.
+				testsFailed = ((result.totalFail ?: 0) + (result.totalError ?: 0)) > 0;
 			} else {
 				// Could be an HTML error page
 				if (reFindNoCase("<html", httpResult)) {
@@ -4413,6 +4420,13 @@ component extends="modules.BaseModule" {
 			}
 		} catch (any e) {
 			out("Test execution failed: #e.message#", "red");
+		}
+
+		// Exit non-zero when specs failed/errored so CI and shells can detect it.
+		// Previously runTests always returned "" → `wheels test` exited 0 even when
+		// tests failed, silently green-lighting broken builds. CLI audit H6.
+		if (testsFailed) {
+			throw(type = "Wheels.TestsFailed", message = "Tests failed — see the report above.");
 		}
 
 		return "";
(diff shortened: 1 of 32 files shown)


bpamiri:
Added two more verified follow-ups to this PR:

- **`generate test --force` guard** — `generate test` refused to overwrite an existing spec unless `--force` (mirrors `generateHelper`); previously it silently clobbered and still printed `create`.
- **`wheels notes` comment-context scanner** — now scans only the comment portion of each line and anchors the token with `\b`, so `TODO`/`FIXME` inside string literals (`x = "TODO: ..."`) and identifier suffixes (`methodTODO`) are no longer reported as real notes. (Known limitation: multi-line `/* ... */` blocks opened on an earlier line aren't tracked.)

Both reproduced + verified in the isolated harness. The prior commits were fully green (CLI suite, Deploy guides docs build, Reviewer A, visual regression).

Issue #2233, opened 2026-04-22
security(dispatch): verify /wheels/public/* is prod-gated or disable in production


bpamiri:
## GA blocker (Bucket A8 in audit)

\`vendor/wheels/Public.cfc\` is instantiated on every app start and exposes functions like \`index\` (congratulations page), \`info\` (env dump), \`routes\` (route table), \`routetester\`, \`testbox\` (test runner). Internal dispatch appears to route \`/wheels/public/*\` to these. No environment gate is visible on Public.cfc itself.

Shipping 4.0 without an audit of what \`/wheels/public/*\` exposes in production risks leaking routes, env info, and possibly test-runner endpoints to unauthenticated users on prod deploys.

### Evidence
- [vendor/wheels/events/onapplicationstart.cfc:316](https://github.com/wheels-dev/wheels/blob/claude/infallible-davinci-58997d/vendor/wheels/events/onapplicationstart.cfc#L316): \`application.\$wheels.public = \$createObjectFromRoot(path="wheels", fileName="Public", method="\$init")\` — instantiated unconditionally.
- [vendor/wheels/Public.cfc:14-47](https://github.com/wheels-dev/wheels/blob/claude/infallible-davinci-58997d/vendor/wheels/Public.cfc#L14-L47): functions \`index\`, \`info\`, \`routes\`, \`routetester\`, \`routetesterprocess\`, \`api\`, \`runner\`, \`testbox\` — mix of debug, introspection, and test-runner endpoints.
- The only gating found: \`setting showDebugOutput=false\` in one function (line 139). Not an access gate.

### Fix
- [ ] **Investigate first.** Trace the dispatch path for \`/wheels/public/*\` URLs — is it a framework wildcard route, a convention lookup, or explicit mapping? Document the entry point.
- [ ] If \`/wheels/public/*\` is reachable in production, add an environment gate: \`if (get("environment") == "production") abort;\` at the top of each handler, or block at the dispatch layer.
- [ ] Confirm \`testbox()\`, \`runner()\`, \`tests_testbox()\` are behind the gate — these are the highest-risk endpoints.
- [ ] Leave the congratulations page (\`index()\`) discoverable in dev/testing envs; gate everything else tightly.
- [ ] Add a spec: simulate a production request to \`/wheels/public/info\` → assert 404 or 403.

### Acceptance
- Clear documentation (in code or CLAUDE.md) of what \`/wheels/public/*\` serves and in which environments.
[...]

PR #2900, opened 2026-06-10
fix(events): validate url.format to prevent error-template path traversal

bpamiri:
## Summary

`$getRequestFormat()` (`vendor/wheels/events/EventMethods.cfc:327`) returned `url.format` verbatim, and `$runOnError` interpolates that value into the on-disk error-template include path — `"#application.wheels.eventPath#/onerror.#local.format#.cfm"` (`vendor/wheels/events/EventMethods.cfc:129`) — guarded only by `FileExists`. An attacker-supplied value such as `?format=../../somefile` therefore enabled path traversal / local file inclusion of any `.cfm` on disk during error rendering. Both call sites are affected via the same helper: the development `showErrorInformation` branch (`EventMethods.cfc:47`) and the production template-include branch (`EventMethods.cfc:128-134`).

## Source

Internal multi-agent framework review, 2026-06-09, finding T4 (format-lfi).

## Fix

Gate the `url.format` branch in `$getRequestFormat()` behind `ReFind("^[A-Za-z0-9]+$", url.format)`, falling through to the existing `"html"` default for anything else. Applied at the source so both consumers are covered with one change. A plain-alnum token covers every configured format key (`html`, `xml`, `json`, `csv`, `pdf`, `xls`) and any `addFormat()` extension, while rejecting traversal sequences, dots, slashes, backslashes, null bytes, encoded variants, and duplicate-param comma-joins.

Deliberately out of scope:
- The `http_accept` else-branch only emits `$get("formats")` keys (developer-controlled config), so it was left alone.
- `$requestContentType` (`vendor/wheels/controller/rendering.cfc:738`) passes `params.format` through verbatim, but both consumers validate against `$acceptableFormats` before building any template path — not vulnerable, no change.
- `vendor/wheels/public/docs/core.cfm:77` does an unvalidated `include "layouts/#request.wheels.params.format#.cfm"`, but that surface is production-blocked by `Public.cfc` `$blockInProduction()` (#2233) — dev-mode-only exposure, recommended as a separate low-severity follow-up finding rather than folding into this diff.

## Tests

Added five regression cases to `vendor/wheels/tests/specs/events/onerrorSpec.cfc` (`$getRequestFormat rejects unsafe format tokens (T4 LFI)`): `../` traversal, slash/dot tokens, two valid alphanumeric formats preserved, and empty-string fallback. A private `$requestFormatFor` helper (mirroring the existing `$expectedStatusFor` pattern in the same spec) saves/restores `url.format` with try/finally so the harness's own `?format=json` output detection is undisturbed.

Local verification on Lucee 7 + SQLite: red pre-fix (3 of the 5 cases fail — the pre-fix code returns `url.format` verbatim, so the traversal/empty cases come back unchanged), green post-fix (13 pass / 0 fail for the bundle). Per-PR CI runs the full engine × DB matrix.

## Cross-engine notes

- No closures, no catch-body `local` assignments (the `finally` only mutates the shared `url` struct), no `attributeCollection`, no `Left(str, 0)`, no reserved-scope parameter names (`formatValue`).
- `$getRequestFormat` keeps `public` access with the `$` prefix (mixin invariant — `$integrateComponents` only copies public methods) and its `EventHandlerInterface` signature is unchanged.
- Regex nano-quirk noted during review: both ORO and `java.util.regex` allow `$` to match before a single trailing newline, so `"json\n"` passes the gate — but the accepted value is then alnum + one `\n` (no slash/dot/backslash possible), `FileExists` misses, and execution lands on the safe `onerror.cfm` fallback. Harmless, not a bypass; tightening to `\A…\z` is an optional cosmetic follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ All historical references to "CFWheels" in this changelog have been preserved fo
 
 ----
 
+## [Unreleased]
+
+### Security
+
+- `$getRequestFormat` now rejects non-alphanumeric `url.format` values, preventing path traversal / local file inclusion via the error-template include path in `$runOnError` (#2900)
+
+----
+
 # [4.0.3](https://github.com/wheels-dev/wheels/releases/tag/v4.0.3) => 2026-06-09
 
 > **Wheels 4.0.3** — third patch on the 4.0 line. Completes the CLI argument-parsing overhaul (`ArgSpec` consumes LuCLI's structured arguments in every command — `--no-*` negations and named-only flags now reach their parsers, and user-error paths exit non-zero) and lands the fixes from a full 24-command CLI audit; write-side commands (`migrate`, `seed`, `reload`, `generate admin`) now refuse to attach to a sibling project's server instead of running against the wrong database; PostgreSQL/CockroachDB foreign-key migrations and pre-23c Oracle `DROP TABLE`/`DROP VIEW` work again; framework helpers can no longer be invoked as controller actions from a URL; auto-derived model properties preserve database column casing; and scaffolded apps keep their reload password out of source control (`WHEELS_RELOAD_PASSWORD` in `.env`). ~45 PRs since the 4.0.2 GA (2026-05-27).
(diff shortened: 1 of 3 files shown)


bpamiri:

PR #2901, opened 2026-06-10
fix(controller): make $acceptableFormats honor onlyProvides per-action restrictions

bpamiri:
## Summary

`onlyProvides()` was a silent no-op: it stores per-action format restrictions in `variables.$class.formats.actions` (`vendor/wheels/controller/provides.cfc:55`), but the reader `$acceptableFormats()` (`vendor/wheels/controller/rendering.cfc:881-887`) checked the **top-level** `variables.$class.formats` struct — whose only keys are `default` / `actions` / `existingTemplates` / `nonExistingTemplates` — so the lookup never matched and every `onlyProvides()` call was ignored. The function also read `arguments.action` without declaring the argument.

## Source

Internal multi-agent framework review, 2026-06-09, finding T6 (onlyprovides-noop).

## Fix

- `$acceptableFormats()` now reads `variables.$class.formats.actions[arguments.action]` — the exact sub-struct `onlyProvides()` writes — guarded by `Len(arguments.action) && StructKeyExists(...)`. `formats.actions` is always initialized in `Controller.cfc`, so the guard is safe.
- Declared `string action = ""` so bare calls fall back to the controller default instead of erroring on an undeclared argument.
- Both framework callers (`$callAction` at `vendor/wheels/controller/processing.cfc:156` and `renderWith` at `vendor/wheels/controller/rendering.cfc:209`) already pass `action` named; format resolution funnels through `$requestContentType()` into those two sites only, so there is no bypass.

**Behavioral change by design:** `onlyProvides()` restrictions are now enforced. `renderWith()` coerces a non-acceptable format to html and the `$callAction` auto-render block skips view rendering for non-acceptable non-html formats — both branches pre-existed (the error text at `processing.cfc:201` already documents `onlyProvides` as their trigger); they were just unreachable.

**Deliberately deferred:** `renderWith()` still coerces a non-acceptable format to html even when html itself is not in the `onlyProvides` list. That fallback semantics pre-exists this fix and is left untouched here; flagged for a follow-up doc note now that enforcement is live.

## Tests

- `vendor/wheels/tests/specs/controller/providesSpec.cfc`: three new read-path specs — restriction honored for the restricted action, fallback to the controller default for unrestricted actions, and a bare call (no `action`) returning the default.
- `vendor/wheels/tests/specs/controller/renderingSpec.cfc`: behavioral spec proving `renderWith()` falls back to the html view when `onlyProvides` excludes the requested xml format.
- All red against the pre-fix code: verified locally on Lucee 7 + SQLite via the worktree docker recipe — HTTP 417 (failures) pre-fix, 200 (pass) post-fix. Pre-fix the default was `"html,xml"` so xml rendered anyway, and the bare-call spec errored on the undeclared `action` argument.
- Specs clean up the shared `formats.actions` controller class data (cached in the application scope by reference) before asserting.

## Cross-engine notes

- Mixin helper stays `public` with `$` prefix (invariant 7).
- No catch-local assignments, no reserved-scope parameter names, no `attributeCollection` usage, no inline closures as constructor named args, no `Left(str, 0)`.
- Spec patterns mirror existing matrix-tested files in the same directory; full engine x DB matrix runs in per-PR CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ All historical references to "CFWheels" in this changelog have been preserved fo
 
 ----
 
+# [Unreleased]
+
+### Fixed
+
+- `onlyProvides()` per-action format restrictions now take effect. `$acceptableFormats()` read the top-level `variables.$class.formats` struct (whose keys are only `default` / `actions` / `existingTemplates` / `nonExistingTemplates`) instead of the `.actions` sub-struct that `onlyProvides()` writes to, so the per-action lookup never matched and every `onlyProvides()` call was a silent no-op since introduction — controllers happily rendered whatever format the request asked for. The read path now checks `formats.actions[action]` (with the `action` argument declared optional so bare calls stay safe); `renderWith()` coerces a non-acceptable requested format to html, and the `$callAction` auto-render block skips view rendering for non-acceptable non-html formats. **Behavior change by design** — apps that relied on the silent no-op will now see restrictions enforced (#2901)
+
+----
+
 # [4.0.3](https://github.com/wheels-dev/wheels/releases/tag/v4.0.3) => 2026-06-09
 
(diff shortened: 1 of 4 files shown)


bpamiri:

PR #2903, opened 2026-06-10
fix(dispatch): block /wheels/* dev UI unless environment is development

bpamiri:
## Summary

`$shouldBlockInProduction()` in `vendor/wheels/Public.cfc` (develop line 19-22) was a denylist that only matched the literal environment name `"production"`. The handlers it guards expose the dev-UI surface — routes browser, env info, a CFML REPL (`consoleeval.cfm`), test runners, the migrator UI, and the MCP endpoint — behind the `enablePublicComponent` gate in `vendor/wheels/Dispatch.cfc:308`. The RCE-grade views (`consoleeval.cfm`, `mcp.cfm`) already required `environment == "development"`, but every *other* gated handler passed for any non-production environment name (`staging`, `qa`, `uat`, `testing`, `maintenance`, `design`, `production-1`, ...) whenever a deployer force-enabled `enablePublicComponent`.

**Intentional behavior change:** setting `enablePublicComponent=true` in testing / maintenance / design / any non-development environment no longer exposes the gated dev-UI handlers — they 404. Only `environment="development"` may reach them. `index()` (the congratulations page) stays ungated per #2233.

## Source

Internal multi-agent framework review, 2026-06-09, finding T2 (devui-allowlist-gate, S2/SEC-1).

## Fix

Rewrite the predicate as a fail-closed allowlist (`vendor/wheels/Public.cfc:23`):

- block when `application.wheels` or `application.wheels.environment` is missing (previously fell open),
- otherwise block unless the environment is `development` (case-insensitive `!=`, the same operator `consoleeval.cfm` uses).

Method names `$shouldBlockInProduction()` / `$blockInProduction()` are kept — 29 handler call sites and the static-coverage spec grep for them. `$loadRegistryPackages()` inherits the stricter gate. The new gate is strictly stronger than the per-view checks in `consoleeval.cfm` / `mcp.cfm`, which fall through when `application.wheels` is missing.

Deliberately deferred (non-blocking follow-ups, not in this minimal diff):
- stale docblock at `vendor/wheels/Public.cfc:47` ("Short-circuits in production" should read "outside development")
- stale comment at `vendor/wheels/tests/specs/dispatch/InvokeMethodSpec.cfc:46` ("non-production... no-op" is now development-only)
- `web/sites/guides/.../digging-deeper/debug-panel.mdx` `enablePublicComponent` override example could note the Tools links 404 outside development

## Tests

`vendor/wheels/tests/specs/security/PublicComponentProductionSpec.cfc`:
- inverted the testing / maintenance / design predicate cases to expect blocking
- added arbitrary non-dev environment names (`staging`, `qa`, `uat`, `prod`, `production-1`)
- added a fail-closed case for a missing `environment` key

`vendor/wheels/tests/specs/packages/LoadRegistryPackagesSpec.cfc`: staging now yields empty packages (would have surfaced `should-not-appear` pre-fix).

All 9 new/inverted assertions fail against develop's predicate — verified RED empirically on Lucee 7 + SQLite: 35 pass / 9 fail pre-fix, 44 / 0 post-fix. Only the `security` + `packages` bundles were run locally (acceptable given the narrow caller surface); **gate the merge on a green per-PR compat-matrix run.**

CI safety: all fixture/demo/template apps set `environment="development"` and nothing flips the env mid-run, so the gated `/wheels/core/tests` endpoint CI hits stays reachable.

## Cross-engine notes

The predicate is `StructKeyExists` + case-insensitive `!=` only — no closures, no reserved-scope names, no `attributeCollection`, no catch-local writes. Specs reuse the file's existing IIFE loop pattern. `$`-prefixed methods stay `public` (mixin invariant #7). Safe `StructDelete`/restore ordering in the missing-key spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ All historical references to "CFWheels" in this changelog have been preserved fo
 
 ----
 
+## [Unreleased]
+
+### Security
+
+- `$shouldBlockInProduction()` is now a fail-closed allowlist: the `/wheels/*` dev-UI handlers (routes, env info, CFML REPL, test runner, migrator UI, MCP) 404 unless `application.wheels.environment == "development"` — case-insensitive, matching the gates in `consoleeval.cfm` and `mcp.cfm` — and also when `application.wheels` or its `environment` key is missing. Previously the predicate matched only the literal `"production"`, so any other environment name (`staging`, `qa`, `testing`, `maintenance`, `design`, ...) bypassed the gate when `enablePublicComponent=true` was set, exposing RCE-grade surfaces. `index()` stays ungated per #2233 and `$loadRegistryPackages()` inherits the stricter gate. **Behavioral change:** operators who set `enablePublicComponent=true` in a non-development environment for ad-hoc debugging will now receive 404s on the Tools links — set `environment="development"` (and `enablePublicComponent=true`) to reach them (#2903)
+
+----
+
 # [4.0.3](https://github.com/wheels-dev/wheels/releases/tag/v4.0.3) => 2026-06-09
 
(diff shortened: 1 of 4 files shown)


bpamiri:

Issue #2906, opened 2026-06-10
db-adapters: same-batch MSSQL identity fallback, Oracle CURRVAL, and $identitySelect six-way dedup


bpamiri:
Split out of the identity-retrieval-race fix (finding "Top 10 #10" / DA3 / DA8 / DA9 of the internal multi-agent framework review, 2026-06-09). That PR makes both adapters prefer the driver-supplied generated key, which fixes the race/trigger hazard on Lucee and Adobe CF. Three pieces of deliberately deferred scope remain:

## 1. SQL Server: same-batch retrieval for the no-generated-key fallback

`vendor/wheels/databaseAdapters/MicrosoftSQLServer/MicrosoftSQLServerModel.cfc` — when the engine surfaces no `result.generatedKey` (currently BoxLang), the adapter falls back to a standalone `SELECT @@IDENTITY`, which is session-scoped and can return a trigger-generated identity from another table. A standalone `SELECT SCOPE_IDENTITY()` is not a fix: it executes in its own batch (a batch is a scope per MS docs) and returns NULL over JDBC. Proper fix options:

- run the retrieval in the insert's own batch (e.g. `INSERT ...; SELECT SCOPE_IDENTITY() AS lastId` as one statement), or
- use `OUTPUT INSERTED.<pk>`, or
- drop the fallback entirely once BoxLang surfaces JDBC generated keys in its query result struct.

## 2. Oracle: replace the MAX(ROWID) last-resort with CURRVAL

`vendor/wheels/databaseAdapters/Oracle/OracleModel.cfc` — the same no-generated-key fallback still resolves the new PK via `WHERE ROWID = (SELECT MAX(ROWID) FROM <table>)`. ROWID encodes physical storage location, not insertion order, so this races under concurrent inserts and can assign another session's key. Replace with `CURRVAL` on the backing identity sequence (session-scoped, race-free), or remove once BoxLang surfaces generated keys.

## 3. DA3: dedupe the six divergent `$identitySelect` implementations

`$identitySelect` is copy-pasted across six adapters with divergent engine workarounds:

- `vendor/wheels/databaseAdapters/Base.cfc`
- `MicrosoftSQLServer/MicrosoftSQLServerModel.cfc`
- `PostgreSQL/PostgreSQLModel.cfc`
- `Oracle/OracleModel.cfc`
- `SQLite/SQLiteModel.cfc`
- `CockroachDB/CockroachDBModel.cfc`

[...]

PR #2908, opened 2026-06-10
fix(model): prefer driver generated keys for Oracle and SQL Server identity retrieval

bpamiri:
## Summary

Identity retrieval after INSERT could return the **wrong primary key** on two adapters, silently corrupting subsequent `update()`/`delete()` calls made through the freshly created object:

- **Oracle** (`vendor/wheels/databaseAdapters/Oracle/OracleModel.cfc:146-149` pre-fix): resolved the new PK via `WHERE ROWID = (SELECT MAX(ROWID) FROM <table>)`. ROWID encodes physical block location, not insertion order, so a concurrent committed insert (or block reuse) returns another session's row.
- **SQL Server** (`vendor/wheels/databaseAdapters/MicrosoftSQLServer/MicrosoftSQLServerModel.cfc:288-295` pre-fix): resolved it via a standalone `SELECT @@IDENTITY`, which is session-scoped across all scopes — an audit/history trigger that inserts into another identity table makes it return the trigger's key (non-empty, so the existing SCOPE_IDENTITY empty-fallback never fired).

## Source

Internal multi-agent framework review, 2026-06-09 — finding T10 (identity-retrieval-races), report "Top 10 #10" plus DA8/DA9.

## Fix

Both adapters now prefer the **driver-supplied generated key** from the insert's own result struct, mirroring the proven CockroachDB adapter:

- **MSSQL**: `ListFirst(result.generatedKey)` when present — mssql-jdbc retrieves it in the insert's own batch, so it is scope-, trigger-, and concurrency-safe. The `@@IDENTITY -> SCOPE_IDENTITY` chain survives only as a documented last resort for engines that surface no generated key (currently BoxLang).
- **Oracle**: resolves `result.generatedKey` (Lucee) / `result.rowid` (ACF). A numeric value is used directly; a ROWID-shaped value is resolved with a race-free exact-row `CHARTOROWID` lookup, gated by a strict case-sensitive `^[A-Za-z0-9/+]{18}$` pattern (quote-free alphabet — `$query` has no parameter binding, so nothing unvalidated is ever interpolated). `MAX(ROWID)` survives only as a documented BoxLang fallback.
- Both adapters also gain the BoxLang `ReplaceList` column-parse workaround already used by the PostgreSQL/CockroachDB adapters, so the fallback path cannot falsely match the PK and skip key retrieval.

**Deliberate deviation from the finding's recommendation**: the suggested "default to `SCOPE_IDENTITY()`" reorder was NOT implemented. A standalone `SELECT SCOPE_IDENTITY()` over JDBC executes in its own batch (a batch is a scope per MS docs) and returns NULL — the documented BoxLang empty-value behavior — so the reorder would add a wasted round-trip per insert and still resolve via `@@IDENTITY`. The rationale is documented in the adapter comments.

**Deferred scope** — tracked in #2906: same-batch retrieval for the MSSQL BoxLang fallback, replacing Oracle's `MAX(ROWID)` last-resort with `CURRVAL`, and the DA3 six-way `$identitySelect` dedup (`$parseInsertColumnList` template + `$lastIdLookup` hook).

Contract unchanged for the sole caller (`Base.cfc::$querySetup` -> `create.cfm`); ACF on MSSQL is byte-identical via the untouched early `identitycol` return.

## Tests

New DB-free unit specs, modeled on the matrix-proven `CockroachDBUnitSpec`:

- `vendor/wheels/tests/specs/database/MicrosoftSQLServerUnitSpec.cfc`
- `vendor/wheels/tests/specs/database/OracleUnitSpec.cfc`

They construct each adapter directly and drive `$identitySelect` with synthetic result structs (driver key present/absent, list-valued keys, PK-in-column-list, non-INSERT). Verified red against the pre-fix adapters via stash-and-rerun (4 pass / 2 errors per bundle on Lucee 7 + SQLite) and green with the fix; the database spec area passes 44 / 0 fail / 0 error locally.

## Cross-engine notes

- Lucee: `result.generatedKey` (case-insensitive `StructKeyExists` matches the lowercase `generatedkey` key). ACF: untouched early `identitycol` return on MSSQL; `result.rowid` on Oracle. BoxLang: takes the guarded fallbacks, now with the `ReplaceList` column-parse workaround copied verbatim from `PostgreSQLModel.cfc`.
- Live Lucee+MSSQL and BoxLang+MSSQL legs were not run locally — **please confirm the PR CI compat-matrix `sqlserver` legs are green before merging**. Oracle is soft-fail in CI as usual; its changes ship guarded behind the strict ROWID regex with the legacy query as fallback.
- No new closures-as-constructor-args, no `attributeCollection = arguments`, all mixin-visible helpers remain public `$`-prefixed.


bpamiri:

PR #2909, opened 2026-06-10
fix(view): redact secret-shaped settings on the /wheels/info HTML page

bpamiri:
## Summary

The HTML branch of the `/wheels/info` debug page rendered `csrfCookieEncryptionSecretKey` in plaintext: `vendor/wheels/public/views/info.cfm:366` renders `#outputSetting(csrf)#` over the CSRF settings array (which includes the key), and `outputSetting()` in `vendor/wheels/public/helpers.cfm:88-98` emitted the raw value via `formatSettingOutput(get(...))`. The JSON branch already redacted that one key via a hardcoded name compare at `info.cfm:195`, so the two branches had drifted and the HTML page leaked the live CSRF cookie encryption key. With `csrfStore="cookie"` that AES key signs the authenticity token (`vendor/wheels/controller/csrf.cfc`), so anyone who can view `/wheels/info` could forge valid CSRF cookies.

## Source

Internal multi-agent framework review, 2026-06-09, finding T3 (info-secret-redaction) — canonical writeup: Security Top 10 #3 / S1 ("CSRF cookie encryption secret rendered in plaintext on the HTML `/wheels/info` page"), also tracked as public-ui P4.

## Fix

Single source of truth on `wheels.Public`:

- `$isProtectedSetting(settingName)` — true for secret-shaped names matching `(secret|password|passphrase|privatekey|apikey|credential|token)`, with an explicit `allowcredentials$` exemption: `accessControlAllowCredentials` is the boolean CORS flag mirroring `Access-Control-Allow-Credentials`, not a credential value. It is the only other rendered setting name matching the regex, so JSON output stays byte-identical today while HTML changes exactly the one secret row.
- `$settingDisplayValue(settingName)` — returns `<em>[redacted]</em>` for protected names WITHOUT calling `get()` (an unset key cannot throw); otherwise renders through `formatSettingOutput(get(...))` as before.

`outputSetting()` now routes every HTML settings row (environment, paths, components, csrf, cors, settings tabs) through `$settingDisplayValue()`, and the JSON branch replaces the hardcoded `csrfCookieEncryptionSecretKey` compare with `!$isProtectedSetting()` on all six collection loops, so future secret-shaped settings are omitted uniformly.

Deliberately deferred (out of scope for T3):
- Production reachability of `/wheels/*` GUI pages (`$blockInProduction`) is a separate review finding; this change only fixes what leaks when the page is reachable.
- The JSON branch still serializes the full `getApplicationMetadata()` under `application.metadata` — pre-existing and orthogonal; flagged by the reviewer as a candidate for its own finding.

## Tests

Added `vendor/wheels/tests/specs/security/InfoSecretRedactionSpec.cfc` (WheelsTest BDD, modeled on `PublicComponentProductionSpec`): unit coverage of `$isProtectedSetting()` (secret-shaped names true, ordinary names false, `accessControlAllowCredentials` exempt) and `$settingDisplayValue()` (redacts without reading the value, passes ordinary settings through), plus source-inspection assertions that `helpers.cfm` renders via `$settingDisplayValue` and `info.cfm`'s JSON branch uses `!$isProtectedSetting`. Verified red on pre-fix code (0 pass / 2 fail / 7 error — the two new methods don't exist and the source pins don't match) and green post-fix (9 pass) on Lucee 7 + SQLite via the worktree docker recipe. Full engine x DB matrix runs in per-PR CI.

## Cross-engine notes

- Both new methods are `public` with a `$` prefix (mixin/internal invariant; `private` would not integrate).
- No closures, no `catch`-scoped locals, no `Left(str, 0)`, no reserved-scope parameter names, no `attributeCollection` usage.
- The regex uses a trailing-anchor exemption (`allowcredentials$`) instead of a lookbehind, which is safe on Adobe's ORO regex engine.
- The unqualified `$settingDisplayValue()` call inside the included `outputSetting()` UDF resolves through the Public component's `variables` scope — the same mechanism the pre-existing `get()` call already relied on. `helpers.cfm`'s double-include into non-Public contexts (tests/`html.cfm` via `_header.cfm`) is safe because `outputSetting()`'s only call site is `info.cfm`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,14 +18,6 @@ All historical references to "CFWheels" in this changelog have been preserved fo
 
 ----
 
-# [Unreleased]
-
-### Security
-
-- The HTML branch of `/wheels/info` no longer renders `csrfCookieEncryptionSecretKey` in plaintext. The HTML branch had been routing every setting through `outputSetting() -> formatSettingOutput(get(...))`, which read and rendered the live CSRF cookie encryption key into the page; the JSON branch already omitted it via a hardcoded one-off compare, so the two branches could drift. A shared predicate `$isProtectedSetting()` and display helper `$settingDisplayValue()` on `wheels.Public` now serve as the single source of truth for both branches — secret-shaped setting names (`secret|password|passphrase|privatekey|apikey|credential|token`) render as `<em>[redacted]</em>` without ever being read, so an unset key cannot throw and the value never reaches the output buffer. `accessControlAllowCredentials` is exempt (it mirrors the boolean `Access-Control-Allow-Credentials` CORS response header, not a credential value); without the exemption it would have been redacted in HTML and silently dropped from the JSON `cors` collection. Reachability of `/wheels/info` is gated by `$blockInProduction()` independently — this change only narrows what leaks when the page is reachable. Verified red→green on Lucee 7 + SQLite (#2909)
-
-----
-
 # [4.0.3](https://github.com/wheels-dev/wheels/releases/tag/v4.0.3) => 2026-06-09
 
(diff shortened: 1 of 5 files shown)


bpamiri:

PR #2910, opened 2026-06-10
fix(model): key association JOIN memo by soft-delete and alias context

bpamiri:
## Summary

`$expandedAssociations()` in `vendor/wheels/model/sql.cfc` (memo site at ~lines 1221-1303) memoized the built association `JOIN` string onto the shared, application-scoped association struct keyed only on existence — so the first caller's context won permanently for the lifetime of the app:

- A join built with `includeSoftDeletes=true` was later served to default calls, **dropping the soft-delete `IS NULL` predicate** (soft-deleted rows silently leak into results).
- An aliased self-join built in a nested-include context was served to top-level includes (and vice versa), producing wrong table references.
- The unlocked write also raced across concurrent first hits.

## Source

Internal multi-agent framework review, 2026-06-09 — finding **H4** (join-memo-poisoning).

## Fix

- Memoize per context variant: a `joinVariants` struct keyed by soft-delete flag + alias flag (`sd{0|1}_alias{0|1}`, `1/0` ternary so keys stay engine-stable on Adobe CF). These two flags are the only per-call inputs to the join build; all other inputs are stable per association struct per app lifecycle.
- Write the memo under a named exclusive lock (`wheelsJoinMemo#application.applicationName#`, matching the framework's established lock-name convention) taken only on memo miss, with an in-lock container-existence re-check to prevent the lost-update race between concurrent first hits for different variants. Steady-state reads stay lock-free, preserving the perf benefit.
- Return per-call shallow copies carrying the context-correct `join` (with `joinVariants` stripped), so callers can never hold a reference that another request re-poisons. Every returned element always carries a `join` key, preserving the existing `StructKeyExists` guards on the `$whereClause` UPDATE paths.
- Stop writing the legacy shared `join` key — no framework reader of `$classData().associations[x].join` exists outside the returned arrays (grepped across `vendor/wheels`, `cli`, `app`, and all test suites). Theoretical third-party-plugin readers of that key are the only behavior change.

**Deliberately deferred:** the context-independent metadata fill-ins above the memo (`foreignKey`/`joinKey`/`tableName`/etc., sql.cfc ~1183-1216) still write to the shared struct unlocked — those writes are idempotent and benign, and are left for a follow-up.

## Tests

New spec `vendor/wheels/tests/specs/model/ExpandedAssociationsJoinMemoSpec.cfc` (WheelsTest BDD) pins all four poisoning mechanisms:

1. `includeSoftDeletes=true` first, then default — default must keep the `deletedat IS NULL` predicate.
2. Default first, then `includeSoftDeletes=true` — the soft-delete variant must drop it.
3. Nested-alias context first, then top-level — top-level must not inherit the alias.
4. Memoization retained — repeat call serves the per-variant memo (perf-preservation contract).

Verified locally (Lucee 7 + SQLite): all four cases **fail against the pre-fix code** and pass post-fix; full model area at 848 pass / 0 fail / 0 error. Per-PR CI runs the full engine × DB compat matrix.

## Cross-engine notes

- Memo keys use a `1/0` ternary instead of boolean string interpolation, avoiding Adobe CF's `YES`/`NO` boolean stringification.
- cfscript `lock` with attributes matches existing framework usage (`Global.cfc`).
- The spec derives the alias fragment from the live adapter's `$tableAlias`/`$quoteIdentifier`, so assertions hold on Oracle (space vs `" AS "`); `toInclude` in the vendored TestBox is case-insensitive.
- No inline-closure constructor args, no catch-local persistence assumptions, no reserved-scope parameter names, no `attributeCollection`, no `Left(str, 0)` introduced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ All historical references to "CFWheels" in this changelog have been preserved fo
 
 ----
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


bpamiri:

PR #2921, opened 2026-06-10
fix(model): SQLite recreate rollback, DEFAULT quote escaping, pg advisory-lock timeout

bpamiri:
## Summary

Hardens five adapter-layer sites found in the 2026-06-09 internal framework review (package `adapter-misc`): SQLite table-recreate now cooperates with the migrator's wrapping transaction instead of defeating its rollback, `Abstract.quote()` escapes every embedded single quote in DDL DEFAULT values, the PostgreSQL advisory lock honors its timeout (matching the MySQL contract), `H2Model.$getColumnInfo()` no longer leaks unscoped variables into the shared adapter instance, and `$assignAdapter()` memoizes the resolved adapter per datasource instead of re-probing the database on every model class init.

## Findings addressed

- **DA7 — SQLite `changeColumnInTable()` raw BEGIN/COMMIT defeats migrator transaction rollback** @ `vendor/wheels/databaseAdapters/SQLite/SQLiteMigrator.cfc:224` — when `request.$wheelsTransactionWrapper` is set (Migrator.cfc wraps migrations in `cftransaction`, #2789), the adapter now emits `PRAGMA defer_foreign_keys = ON` (SQLite's documented in-transaction alternative, auto-resets on commit/rollback) and lets the engine transaction own atomicity; standalone execution keeps the original explicit `PRAGMA foreign_keys` + `BEGIN`/`COMMIT` sequence unchanged.
- **DA10 / SEC-24 — `Abstract.quote()` escaped only the first embedded single quote** @ `vendor/wheels/databaseAdapters/Abstract.cfc:135` — `ReplaceNoCase` defaulted to scope `one`, so `default="O'Brien's"` produced unbalanced DDL; the `binary/date/datetime/time/timestamp` branch did no escaping at all. Both branches are merged and now escape every quote via `Replace(..., "''", "all")`. `CURRENT_TIMESTAMP` passthrough is preserved and spec-asserted.
- **DA11 — PostgreSQL `$acquireAdvisoryLock()` blocked forever, ignoring its timeout** @ `vendor/wheels/databaseAdapters/PostgreSQL/PostgreSQLModel.cfc:203` — replaced blocking `pg_advisory_lock` with a `pg_try_advisory_lock` poll loop (250ms interval) that throws `Wheels.AdvisoryLockTimeout` on expiry, matching the MySQL adapter's contract verbatim; `timeout=0` gives one try then throws. The CockroachDB override is unaffected.
- **DA6 — `H2Model.$getColumnInfo()` leaked unscoped `pkList`/`row`/loop counter into the shared adapter instance** @ `vendor/wheels/databaseAdapters/H2/H2Model.cfc:197` — mechanical `local.` scoping of all three.
- **DA15 — `$assignAdapter()` re-probed the database on every model class init** @ `vendor/wheels/Model.cfc:380` — resolved adapter (namespace + name) is now memoized per datasource in `application.wheels.adapterCache`, rebuilt on framework reload. No negative caching: the probe throws before any cache write, and the cached path preserves `$set(adapterName=...)` and credential parity. Sole caller is the lock-guarded class-init path in `Model.cfc`, so the check-then-set is benign.

None — all five findings in this package were live on `origin/develop` (the `quote()` single-escape and blocking `pg_advisory_lock` were re-verified against the develop tip before patching).

## Source

Internal multi-agent framework review, 2026-06-09 — wave 2, package `adapter-misc`.

## Tests

New specs, each written first and verified red against pre-fix code, then green post-fix (Lucee 7 + SQLite locally):

- `vendor/wheels/tests/specs/migrator/adapterQuoteEscapingSpec.cfc` — multi-quote escaping in both type branches + `CURRENT_TIMESTAMP` passthrough (3 failures pre-fix).
- `vendor/wheels/tests/specs/migrator/sqliteChangeColumnTransactionSpec.cfc` — wrapped path emits no raw `BEGIN`/`COMMIT`/`PRAGMA foreign_keys` and uses `defer_foreign_keys`; standalone path byte-identical to before (1 failure + 1 error pre-fix). Skips itself on non-SQLite databases.
- `vendor/wheels/tests/specs/model/adapterMemoizationSpec.cfc` — cache populated per datasource, adapter identity stable across re-resolution (1 failure pre-fix).

The PostgreSQL polling path (`lockingSpec` on the postgres matrix leg) and Adobe/BoxLang compilation of the new specs are exercised by CI's full engine x DB matrix, per campaign rules.

## Cross-engine notes

- All 11 cross-engine invariants checked against the diff: no inline closures as constructor named args, no `obj.map()`, no `attributeCollection = arguments`, no `Left(str, 0)`, no reserved-scope parameter names, no `local.X`-in-catch assertions.
- The `skip()` guard in the SQLite spec's `beforeEach` is safe on every engine: TestBox runs `beforeEach` inside the try whose catch handles `TestBox.SkipSpec`, so non-SQLite legs mark the spec Skipped (not errored), and the guard is the first statement so no cleanup is bypassed.
- `Sleep()`, `GetTickCount()`, `Replace(..., "all")`, and struct-literal cache entries are portable across Lucee 5/6/7, Adobe 2018-2025, and BoxLang.

## Changelog

### DIFF
--- a/vendor/wheels/Model.cfc
+++ b/vendor/wheels/Model.cfc
@@ -371,6 +371,24 @@ component output="false" displayName="Model" extends="wheels.Global"{
 	 * Internal function.
 	 */
 	public any function $assignAdapter() {
+		// Memoize the resolved adapter per datasource: the engine behind a
+		// datasource doesn't change at runtime, so probing the database
+		// ($dbinfo version plus a SELECT version() roundtrip for
+		// PostgreSQL-driver datasources) on every model class init is
+		// redundant. The cache lives in the application scope and is rebuilt
+		// on framework reload.
+		if (!StructKeyExists(application.wheels, "adapterCache")) {
+			application.wheels.adapterCache = {};
+		}
+		if (StructKeyExists(application.wheels.adapterCache, variables.wheels.class.dataSource)) {
+			local.cached = application.wheels.adapterCache[variables.wheels.class.dataSource];
+			$set(adapterName = local.cached.name);
+			return CreateObject("component", "wheels.databaseAdapters.#local.cached.namespace#.#local.cached.name#").$init(
+				dataSource = variables.wheels.class.dataSource,
+				username = variables.wheels.class.username,
+				password = variables.wheels.class.password
+			);
+		}
 		if ($get("showErrorInformation")) {
 			try {
 				local.info = $dbinfo(
@@ -439,6 +457,10 @@ component output="false" displayName="Model" extends="wheels.Global"{
 				extendedInfo = "Use SQL Server, MySQL, MariaDB, PostgreSQL, CockroachDB, Oracle, SQLite or H2."
 			);
 		}
+		application.wheels.adapterCache[variables.wheels.class.dataSource] = {
+			namespace: local.adapterNamespace,
+			name: local.adapterName
+		};
 		$set(adapterName = local.adapterName);
 		return CreateObject("component", "wheels.databaseAdapters.#local.adapterNamespace#.#local.adapterName#").$init(
 			dataSource = variables.wheels.class.dataSource,
--- a/vendor/wheels/databaseAdapters/Abstract.cfc
+++ b/vendor/wheels/databaseAdapters/Abstract.cfc
@@ -126,13 +126,13 @@ component extends="wheels.migrator.Base"{
 		}
 		if (
 			StructKeyExists(arguments.options, 'type') && ListFindNoCase(
-				"binary,date,datetime,time,timestamp",
+				"binary,date,datetime,time,timestamp,text,string",
 				arguments.options.type
 			)
 		) {
-			arguments.value = "'#arguments.value#'";
-		} else if (StructKeyExists(arguments.options, 'type') && ListFindNoCase("text,string", arguments.options.type)) {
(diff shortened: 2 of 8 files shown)


bpamiri:
Re: Reviewer A ([review 4465418255](https://github.com/wheels-dev/wheels/pull/2921#pullrequestreview-4465418255)) — all three points addressed:

- **Spec semicolons (blocking):** fixed in `4eddb04c3`, merged with this PR. For the record, the suite already contains Adobe-green specs without statement semicolons (e.g. `migratorSpec.cfc`), so the compile-crash risk was overstated — but the consistency fix was right regardless.
- **`char` missing from `Abstract.quote()`:** confirmed — `char` is string-family in `addColumnOptions` (line 87) but was absent from `quote()`'s list, so a `char` default containing a quote emitted unquoted, unbalanced DDL. Fixed in follow-up #2948 (this PR merged before the fix could land on its branch), with a new `adapterQuoteEscapingSpec` assertion.
- **`adapterMemoizationSpec` teardown:** `beforeAll`/`afterAll` snapshot-and-restore of `application.wheels.adapterCache` also in #2948.

Verified on the develop tip, Lucee 7 + SQLite: migrator 290 passed, model 871 passed.

PR #2922, opened 2026-06-10
fix(controller): sendEmail multipart handling and sendFile path containment

bpamiri:
## Summary

Fixes nine review findings in the controller mailer (`sendEmail`) and file-streaming (`sendFile`) mixins in `vendor/wheels/controller/miscellaneous.cfc`. The `sendEmail` fixes remove three crash paths (3+ templates, `layout=""`, `detectMultipart=false` without `type`), make `detectMultipart=false` actually honor the documented text-first template order, emit a real CRLF-CRLF blank line in `writeToFile` output, stop email view arguments from leaking into (or permanently overwriting parts of) the controller's `variables` scope, and pass Adobe's S/MIME signing/encryption attributes through to `cfmail` instead of silently swallowing them. The `sendFile` fixes add a path-traversal guard on `file`/`directory` (null-byte strip + `..` rejection after URL-decode and backslash normalization, same shape as `$generateIncludeTemplatePath`) and sanitize the download display name so it cannot break out of the quoted `Content-Disposition` filename parameter.

## Findings addressed

- **ML1 [High] Out-of-bounds layout indexing crashes on 3+ templates or an empty `layout` string** @ `vendor/wheels/controller/miscellaneous.cfc:42-45` — zero-length `layout` is coerced to `false` before the list split, and the 3+-template sub-cases now route into the ML2 throw before the render loop.
- **ML2 [High] Hardcoded two-part typing makes any 3+-part email impossible** @ `vendor/wheels/controller/miscellaneous.cfc:47-55` (docblock at `:10`) — `ListLen(arguments.template) > 2` now throws a friendly `Wheels.IncorrectArguments` documenting the 2-template limit, before any side effects.
- **ML3 [High] `detectMultipart=false` + single template + no `type` → undefined-variable crash** @ `vendor/wheels/controller/miscellaneous.cfc:135-138` — new `else if` defaults `arguments.type = "text"` (matching `cfmail`'s own default) when no explicit type is passed.
- **ML4 [Medium] `detectMultipart=false` does not actually disable multipart detection** @ `vendor/wheels/controller/miscellaneous.cfc:91-103` — the `<`-count reordering is now gated on `arguments.detectMultipart`; when false, the given order is preserved and labeled `[1]=text` / `[2]=html` per the documented contract.
- **ML5 [Low] `writeToFile` joins text and html bodies with a single bare CR, not a blank line** @ `vendor/wheels/controller/miscellaneous.cfc:176-184` — plain concatenation with a real CRLF-CRLF separator (CFML list functions treat each delimiter character separately); separator skipped when either side is empty.
- **ML6 [Medium] Custom email-view arguments are injected into the controller's `variables` scope and never cleaned up** @ `vendor/wheels/controller/miscellaneous.cfc:62-72` (snapshot) and `:109-118` (restore) — shadowed values are restored and added keys deleted after the email templates render.
- **ML7 [Medium] Unknown mail attributes are silently swallowed by the hardcoded allowlist** @ `vendor/wheels/controller/miscellaneous.cfc:40` — allowlist gains `sign`, `keystore`, `keystorepassword`, `keyalias`, `keypassword`, `encrypt`, `recipientcert`, `encryptionalgorithm`; docblock at `:5` now documents the allowlist semantics.
- **C10 [Medium] `sendFile()` has no path-traversal guard on `file`/`directory`** @ `vendor/wheels/controller/miscellaneous.cfc:225-238` — null bytes stripped, then `..` rejected on the URL-decoded, backslash-normalized `file` AND `directory` before any path math; throws `Wheels.InvalidPath`. Blocks literal, percent-encoded, and backslash variants.
- **SEC-23 [Low] `Content-Disposition` filename built from an unsanitized display name in `sendFile()`** @ `vendor/wheels/controller/miscellaneous.cfc:322-323` — `[\r\n"\\]` stripped from the display name at the finalization point, covering both the `deliver=true` header and the `deliver=false` test struct.

None — all nine findings reproduced against `origin/develop` at implementation time (zero stale references). The reviewer independently spot-checked ML3 (develop's `if` without `else` at the type-assignment site), ML5 (develop's 4-character `ListAppend` delimiter), and C10 (no guard on develop) and confirmed all were live bugs.

## Source

Internal multi-agent framework review 2026-06-09, wave 2, package `controller-mailer-sendfile`.

## Tests

11 new BDD specs in `vendor/wheels/tests/specs/controller/miscellaneousSpec.cfc` (plus a new view asset `vendor/wheels/tests/_assets/views/test/bracketsemailtemplate.cfm`), each written to fail against the pre-fix code: path-traversal rejection (literal, encoded, backslash, and `directory` variants), display-name sanitization, `writeToFile` blank-line bytes, `layout=""`, >2 templates, `detectMultipart=false` without `type`, order preservation under `detectMultipart=false`, variables-scope restore (shadow + leak), and S/MIME attribute pass-through.

Local verification: single-bundle run of `wheels.tests.specs.controller.miscellaneousSpec` on Lucee 7 + SQLite via the prebuilt docker test image — HTTP 200, **39 pass / 0 fail / 0 error** (1 pre-existing intentional skip). Full cross-engine coverage deferred to the CI compat-matrix, which is the real gate.

## Cross-engine notes

No new closures, no `arguments`-as-`attributeCollection`, no `Left(str, 0)`, no `local` assignment inside `catch`; all touched functions remain `public`. The `ReReplace("[\r\n""\\]", ...)` sanitizer has direct in-tree precedent in `vendor/wheels/controller/sse.cfc` (runs on all engines), and the `URLDecode` + backslash-normalize traversal guard mirrors `$generateIncludeTemplatePath` (`vendor/wheels/global/rendering.cfc`). The new test asset's bare `<` sequences are not tag-parseable (Lucee tag-scanner safe). Reviewer flagged two non-blocking follow-up suggestions (try/finally around the ML6 restore; padding `layoutArray` to template count for the degenerate `layout=","` input) — deferred to keep this diff minimal.

## Changelog


bpamiri:
Addressed the Reviewer A review in 97a10c256:

**ML6 `try/finally` (blocking)** — Agreed and fixed, with one engine-driven twist: putting the restore loop directly inside the `finally` block crashes Lucee 7 at runtime with `variable [local] doesn't exist` — Lucee 7 (verified on `lucee@7.0.1+100`) miscompiles `for` loops that use `local.`- or `var`-scoped variables inside `finally` blocks (minimal probe: a bare `for (local.i = 1; ...)` in any `finally` reproduces it; plain `local.x = 1` assignments and function calls are fine). So the render loop is now wrapped in `try`, and the `finally` calls a new `$restoreEmailViewVariables()` helper (public + `$`-prefix per the mixin-integration rule) that contains the loop. A new spec (`does not leak custom view arguments into the controller when rendering throws`) covers the render-exception path.

**`Find("..")` substring note** — The `Wheels.InvalidPath` message/extendedInfo now state explicitly that two consecutive dots are rejected anywhere in the value (calling out `report..final.pdf`-style names), and the `@file`/`@directory` docblocks document the guard.

**Missing backslash traversal test** — Added `rejects backslash path traversal in the file argument` using `Chr(92)` construction, matching `PathTraversalSpec`'s precedent. The PR description's test inventory is now accurate.

**Changelog** — still intentionally deferred to the campaign-end consolidation, as noted.

Verified `wheels.tests.specs.controller.miscellaneousSpec` on Lucee 7 + SQLite **and** Adobe 2023 + SQLite (mixin surface touched): both **41 pass / 0 fail / 0 error** (1 pre-existing intentional skip).

PR #2926, opened 2026-06-10
fix(auth): JWT secret validation and token strategy hardening

bpamiri:
## Summary

Hardens the `wheels.auth.*` primitives against weak-key and credential-leak misconfiguration:

- `JwtService.init()` fails fast on bad HMAC secrets: empty secret throws `Wheels.Auth.JWT.InvalidSecretKey`; a secret under 32 UTF-8 bytes throws `Wheels.Auth.JWT.WeakSecretKey` (RFC 7518 §3.2), instead of failing obscurely inside `HMac()` or silently signing every token with a brute-forceable key.
- `JwtService.decode()` now validates the `iss` claim (present + case-sensitive match) whenever an issuer was configured, throwing `Wheels.Auth.JWT.InvalidIssuer` after signature verification. `refresh()` inherits the check; `JwtStrategy` maps the failure to a 401 (no 500s).
- `TokenStrategy` static token map entries are matched case-sensitively and in constant time (`MessageDigest.isEqual` over UTF-8 bytes, iterating all keys without early break) — CFML struct key lookups are case-insensitive, so `"AbC123"` previously authenticated as `"abc123"`.
- `TokenStrategy.queryParam` now defaults to `""` (disabled), matching `JwtStrategy`, so bearer credentials are no longer accepted from the URL by default and cannot leak into access logs, browser history, or `Referer` headers. Set `queryParam` explicitly (e.g. `"api_key"`) to opt back in — this is a deliberate, documented breaking change.
- Perf: the `java.security.MessageDigest` and `java.lang.System` class handles are cached in `init()`; `$epochSeconds()` keeps wall-clock `currentTimeMillis()` semantics with zero per-call object creation.

The authentication-patterns guide and the `JwtStrategy` doc-comment example (whose old `"my-secret"` sample would now throw `WeakSecretKey`) are updated to match.

## Findings addressed

- **MA10 [Medium]** `JwtService` accepts an empty or trivially weak HMAC secret with no fail-fast validation; `decode()` never validates `iss` @ `vendor/wheels/auth/JwtService.cfc:31-46` (secret validation) and `vendor/wheels/auth/JwtService.cfc:150-160` (issuer validation)
- **MA11 [Medium]** `TokenStrategy` static token map matches API keys case-insensitively, and tokens are accepted via query string by default @ `vendor/wheels/auth/TokenStrategy.cfc:49` (`queryParam` default) and `vendor/wheels/auth/TokenStrategy.cfc:229-244` (constant-time case-sensitive lookup)
- **MA16 [Low, perf]** `JwtService` creates Java class handles per call instead of caching them @ `vendor/wheels/auth/JwtService.cfc:54` (cached `MessageDigest` handle) and `vendor/wheels/auth/JwtService.cfc:312` (`GetTickCount` epoch idiom)
- **SEC-18 [Medium]** — security-pass duplicate of MA10; resolved by the same change
- **SEC-19 [Medium]** — security-pass duplicate of MA11; resolved by the same change

None — `staleRefs` for this package is empty (spot-checked for MA10 and MA11 against `origin/develop` during review; all five findings required code changes on this branch).

## Source

Internal multi-agent framework review 2026-06-09, wave 2, package `auth-services` (findings middleware-auth:10/11/16, security-pass:18/19).

## Tests

New/updated specs under `vendor/wheels/tests/specs/auth/` (WheelsTest BDD), each verified red against the pre-fix `origin/develop` source:

- `JwtServiceSpec.cfc` — empty-secret throws `InvalidSecretKey`; <32-byte secret throws `WeakSecretKey`; exactly-32-byte boundary accepted; issuer match decodes; mismatched / missing / case-variant `iss` throw `InvalidIssuer`; no `iss` validation when no issuer configured.
- `TokenStrategySpec.cfc` — query-param tokens ignored by default in both `supports()` and extraction; explicit `queryParam` opt-in works for `params` and `urlParams`; static token rejected when it differs only in case; exact-case token accepted.
- `JwtStrategySpec.cfc` — fixtures updated for the >=32-byte secret requirement.

Local verification: ran the full `wheels.tests.specs.auth` area via the worktree-safe docker recipe on Lucee 7 + SQLite **and** Adobe CF 2023 + SQLite — **167 pass, 0 fail, 0 error (HTTP 200)** on each. This also exercised the `Len(CharsetDecode(...))` byte-counting idiom on real engines. Full cross-engine coverage is deferred to the CI compat matrix, which constructs `JwtService` in every auth spec.

## Cross-engine notes

- Checked against the CLAUDE.md invariants list: no inline closures as constructor named args, no `local.X` assignments inside `catch`, no reserved-scope parameter names, no `Left(str, 0)`; `Throw` + `extendedInfo` has same-file prior art.
- `Compare()` is used for the case-sensitive issuer check and `MessageDigest.isEqual` for token bytes — both behave identically across Lucee 5/6/7, Adobe 2018-2025, and BoxLang.
- Adobe CF uppercases unquoted struct-literal keys, so a user-defined static token map with unquoted keys would not match under case-sensitive comparison; the guide examples use quoted keys, and the `@tokens` doc-comment now carries the quoted-keys caveat (bot review round 1).
- `$epochSeconds()` uses the cached `java.lang.System` handle's `currentTimeMillis()` — guaranteed wall-clock epoch on every engine. The earlier `GetTickCount()` form was reverted per bot review: although it empirically equals `currentTimeMillis()` on Adobe CF 2023 and Lucee 7 (probed in-container), its documented contract only promises a relative millisecond clock, which is not a safe basis for RFC 7519 claims.

## Changelog

### DIFF
--- a/vendor/wheels/auth/JwtService.cfc
+++ b/vendor/wheels/auth/JwtService.cfc
@@ -50,8 +50,9 @@ component output="false" {
 		variables.issuer = arguments.issuer;
 		variables.allowedClockSkew = arguments.allowedClockSkew;
 
-		// Cache the Java class handle used on the per-request decode path
+		// Cache the Java class handles used on the per-request encode/decode paths
 		variables.messageDigest = CreateObject("java", "java.security.MessageDigest");
+		variables.javaSystem = CreateObject("java", "java.lang.System");
 
 		return this;
 	}
@@ -307,9 +308,14 @@ component output="false" {
 
 	/**
 	 * Get current time as Unix epoch seconds (UTC).
+	 *
+	 * Uses the cached java.lang.System handle: currentTimeMillis() is guaranteed wall-clock
+	 * epoch time on every engine, unlike GetTickCount(), whose contract only promises a
+	 * relative millisecond clock (JVM uptime on some engines) — unusable for RFC 7519
+	 * iat/exp/nbf claims.
 	 */
 	private numeric function $epochSeconds() {
-		return Int(GetTickCount() / 1000);
+		return Int(variables.javaSystem.currentTimeMillis() / 1000);
 	}
 
 }
--- a/vendor/wheels/auth/JwtService.cfc
+++ b/vendor/wheels/auth/JwtService.cfc
@@ -5,7 +5,7 @@
  * Used standalone or by JwtStrategy for request authentication.
  *
  * Usage:
- *   var jwt = new wheels.auth.JwtService(secretKey="my-secret-key");
+ *   var jwt = new wheels.auth.JwtService(secretKey="a-random-secret-of-at-least-32-bytes");
  *   var token = jwt.encode({sub=42, role="admin"});
  *   var claims = jwt.decode(token);
  *   var refreshed = jwt.refresh(token);
@@ -18,9 +18,9 @@ component output="false" {
 	/**
 	 * Creates a new JwtService instance.
 	 *
-	 * @secretKey The HMAC-SHA256 signing key.
+	 * @secretKey The HMAC-SHA256 signing key. Must be at least 32 bytes (256 bits) per RFC 7518 Section 3.2; construction throws otherwise.
 	 * @defaultExpiry Default token lifetime in seconds (default 3600 = 1 hour).
-	 * @issuer Default issuer claim (iss). Empty string means no iss claim added.
+	 * @issuer Default issuer claim (iss). Empty string means no iss claim added and no iss validation on decode.
 	 * @allowedClockSkew Seconds of clock skew tolerance for expiry checks (default 0).
 	 */
 	public JwtService function init(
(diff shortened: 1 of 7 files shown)


bpamiri:
Addressing wheels-bot Reviewer A (review round 1) — fixes pushed in 31e9d4e0d.

**1. `$epochSeconds()` / `GetTickCount()` — fix implemented, with one factual correction.**
I probed all three clocks in a single request on a real Adobe CF 2023.0.11.330706 container before changing anything:

| clock | value |
|---|---|
| `System.currentTimeMillis()` | 1781077761329 |
| `GetTickCount()` | 1781077761329 (delta **0 ms**) |
| JVM uptime (`RuntimeMXBean.getUptime()`) | 322,687 ms |

So the specific claim "on Adobe CF, `GetTickCount()` returns JVM uptime" is not correct on Adobe 2023 — it returns wall-clock epoch ms there, identical to Lucee, and the described never-expiring-token scenario would not have occurred on that engine. The substantive objection stands anyway: `GetTickCount()`'s documented contract only guarantees a *relative* millisecond clock, which is not a safe foundation for RFC 7519 `iat`/`exp`/`nbf` across all current and future engines. Implemented exactly the suggested shape: `variables.javaSystem` is cached in `init()` alongside `messageDigest`, and `$epochSeconds()` returns `Int(variables.javaSystem.currentTimeMillis() / 1000)` — the MA16 zero-allocation goal is preserved with guaranteed semantics, and the doc-comment now warns against the `GetTickCount()` substitution.

Re-verified `wheels.tests.specs.auth` at the new commit: Adobe CF 2023 + SQLite **167 pass / 0 fail / 0 error** and Lucee 7 + SQLite **167 / 0 / 0** (HTTP 200, dir-only docker mounts).

**2. Unquoted struct-literal token keys on Adobe CF — accepted.** The `@tokens` doc-comment (`TokenStrategy.cfc:41`) now says: *"Always quote the keys (e.g. {"AbC-123": {id: 1}}): Adobe CF uppercases unquoted struct-literal keys, so an unquoted mixed-case key would never match its token."* The guide examples already used quoted keys.

**3. Commit scope — accepted.** PR title changed to `fix(auth): …` so the squash-merge landing commit carries the correct layer scope.

**4. Changelog** — per campaign convention, consolidated at campaign end; no change.

PR #2929, opened 2026-06-10
perf(model): findEach cache retention, quadratic serialize dedupe, double hashedKey

bpamiri:
## Summary

Fixes three performance findings in the model read/serialize hot paths (review findings M11, M12, M13 in the model-orm subsystem):

- `findAll`'s request-level query cache stored results unconditionally (the read was gated on `cacheQueriesDuringRequest`, the store was not), and the batch finders `findEach`/`findInBatches` accumulated every batch's PK-page and data query in the request cache for the rest of the request — defeating their memory-efficiency purpose. The store and key computation are now gated on the setting, batch finders opt out via an internal `$useRequestCache` flag (forwarded to the recursive pagination-ids `findAll`), and the COUNT now runs once up front via `$countForBatching` and is passed to `findAll`'s existing `count` argument so it is not re-run per batch even when request caching is disabled.
- `$serializeQueryToArray` used a growing `Chr(7)`-delimited list with `ListFind` plus a per-row `SerializeJSON`+`Hash` — an O(n²) scan run even when `include` was empty (where join duplicates cannot occur). Duplicate detection now uses a struct-as-set for O(1) lookups and is skipped entirely when no associations are included. Intentional behavior change (documented in the commit): identical rows from PK-less selects are no longer silently dropped on plain queries.
- `findAll` computed two full `$hashedKey` serializations of the arguments struct per call. The request cache key is now derived from the SQL shell key (which already encodes the model name and full arguments struct) plus the original where clause, eliminating the second serialization. Nothing mutates `arguments` between the two original hash sites, so the derived key is informationally equivalent.

## Findings addressed

- **M11 [Medium, perf] `findEach`/`findInBatches` retain every batch in the request query cache (and re-COUNT when caching is disabled)** @ `vendor/wheels/model/read.cfc:303-311` (gated store + key), `:664/:678-682` and `:743/:758-762` (batch opt-out + up-front count), `:801` (`$countForBatching`)
- **M12 [Medium, perf] O(n²) duplicate-row detection with per-row JSON hashing in query→objects serialization** @ `vendor/wheels/model/serialize.cfc:98-109, 120, 135, 151, 183-184`
- **M13 [Low, perf] `findAll` computes two full `$hashedKey` serializations of the arguments struct per call** @ `vendor/wheels/model/read.cfc:311`

None. Both staleness spot-checks confirmed the report's references were still accurate against current `origin/develop`: the store at `read.cfc:341` was still unconditional, and the double `$hashedKey` at `read.cfc:234`/`:305` was still present. All three findings required fixes.

## Source

Internal multi-agent framework review 2026-06-09, wave 2, package `model-read-serialize`.

## Tests

New specs (RED-verified against the pre-fix code, with failure shapes matching each finding):

- `vendor/wheels/tests/specs/model/requestQueryCacheSpec.cfc` — single entry per unique `findAll` when enabled; distinct entries for same-shape queries differing only by where values (guards the M13 key derivation); no store when `cacheQueriesDuringRequest` is disabled; `findEach`/`findInBatches` do not accumulate per-batch cache entries (pre-fix: 11/9 accumulated entries).
- `vendor/wheels/tests/specs/model/serializeDedupeSpec.cfc` — one item per query row when `include` is empty even for identical rows (pre-fix: duplicates dropped); join-duplicated rows still deduped when `include` is present.

Specs restore the mutated `application.wheels.cacheQueriesDuringRequest` setting in `afterEach`. Local verification: full model spec area on Lucee 7 + SQLite passes 862/0/0 post-fix. Full cross-area, cross-engine coverage deferred to CI per campaign rules.

## Cross-engine notes

- `$countForBatching` is `public` with a `$` prefix (mixin integration only copies public methods — Cross-Engine Invariant 7).
- Specs use shared-struct closure state, no inline closures as constructor named args, no `local.X` writes inside `catch`, no literal `##` in string literals.
- `StructCount()` over `returnAs="structs"` results is valid because that return shape is a numerically keyed struct.
- No reserved-scope parameter names introduced; the new `$useRequestCache` flag participates in the query-shell key hash, so batch queries get cache entries distinct from identical non-batch queries (no cross-contamination).

## Changelog

### DIFF
--- a/vendor/wheels/model/read.cfc
+++ b/vendor/wheels/model/read.cfc
@@ -232,8 +232,11 @@ component {
 			local.originalWhere = arguments.where;
 			arguments.where = ReReplace(arguments.where, variables.wheels.class.RESQLWhere, "\1?\8", "all");
 
-			// get info from cache when available, otherwise create the generic select, from, where and order by clause
-			local.queryShellKey = $hashedKey(variables.wheels.class.modelName, arguments);
+			// get info from cache when available, otherwise create the generic select, from, where and order by clause.
+			// $useRequestCache governs request-level caching only — it doesn't change the generated SQL — so strip it from the shell-key args; otherwise the application-scoped SQL cache fragments into two entries (batch vs non-batch) for every model that uses both.
+			local.shellKeyArgs = Duplicate(arguments);
+			StructDelete(local.shellKeyArgs, "$useRequestCache");
+			local.queryShellKey = $hashedKey(variables.wheels.class.modelName, local.shellKeyArgs);
 			local.sql = $getFromCache(local.queryShellKey, "sql");
 			if (!IsArray(local.sql)) {
 				local.sql = [];
@@ -663,6 +666,11 @@ component {
 		// Run the COUNT query once up front and pass the total to every page request below so each batch doesn't re-run it.
 		local.totalRecords = $countForBatching(argumentCollection = arguments);
 
+		// Nothing to iterate — return now so findAll isn't called with count=0 (which would re-run COUNT internally).
+		if (local.totalRecords == 0) {
+			return;
+		}
+
 		local.page = 1;
 		local.keepGoing = true;
 
@@ -675,9 +683,7 @@ component {
 			local.finderArgs.page = local.page;
 			local.finderArgs.perPage = arguments.batchSize;
 			local.finderArgs.includeSoftDeletes = arguments.includeSoftDeletes;
-			if (local.totalRecords > 0) {
-				local.finderArgs.count = local.totalRecords;
-			}
+			local.finderArgs.count = local.totalRecords;
 			// Opt out of the request-level query cache so the per-batch results don't accumulate in memory for the remainder of the request (which would defeat the purpose of batched processing).
 			local.finderArgs.$useRequestCache = false;
 			if (arguments.returnAs == "struct") {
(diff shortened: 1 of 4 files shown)


bpamiri:
Addressed Reviewer A's two findings (CHANGES_REQUESTED, 2026-06-10T07:13:26Z). Both verified valid against the code before changing anything.

**[C1] `$useRequestCache` in the application-scoped SQL shell key — fixed.** Confirmed: the flag has no effect on generated SQL, so hashing it into the shell key at `read.cfc:236` fragmented the "sql" cache into batch/non-batch pairs per shape. The shell key is now computed from a copy of `arguments` with the flag removed (`3d795385f`), refined in `b71dac9c5` to use a shallow `StructCopy` instead of `Duplicate` — we only remove a top-level key and `$hashedKey` never mutates, so a deep copy of the full arguments struct on every `findAll` would be unnecessary hot-path cost. One note on the review's framing: stripping the flag is safe not because the request-level key "remains unaffected", but because batch calls never read or write the request cache at all (both sides gated on the flag at `read.cfc:306/318/353`), so no cross-contamination path exists. The PR body's original "distinct entries, no cross-contamination" rationale was therefore unnecessary, and the fragmentation cost was real.

**[C2] Double COUNT for empty result sets — fixed.** Confirmed: with `$countForBatching` returning 0, the `> 0` guard skipped `count`, the loop still ran one iteration, and `findAll` re-ran the COUNT (it only honors `count` when > 0). Both batch finders now early-return on a zero count (`3d795385f`); the now-redundant per-iteration guard is gone. Observable behavior unchanged — the callback was never invoked for empty sets before (still covered by `queryBuilderSpec` empty-`whereIn` specs). `b71dac9c5` adds RED-verified regression specs: no-match `findEach`/`findInBatches` leave exactly one request-cache entry (the single up-front COUNT); pre-fix both fail with `Expected [1] but received [2]`.

Local verification (Lucee 7 + SQLite): full core suite **3935/0/0** with the fixes vs 3932/0/0 baseline at the original PR head; both new C2 specs and the C1 SQL-fragmentation spec RED-verified against pre-fix code.

PR #2939, opened 2026-06-10
fix(plugin): deprecation links and text, log routing, write-back references

bpamiri:
## Summary

Finalizes the **plugins-deprecation** package from the 2026-06-09 framework review (wave 2). Six findings across the legacy plugin subsystem and its deprecation messaging: plugin-load tracing no longer pollutes the security log, `$invokeOnPluginLoad` no longer replaces live `application` references (DI container, config) with stale deep clones, mixin classification matches dotted-path segments instead of substrings, all plugin deprecation messages point at live versioned guide URLs with accurate present-tense text and the real remediation (`wheels packages add <name>`), and a new shared `$deprecated()` helper unifies the warn-once policy with a debug-panel consumer.

Reviewer history: `CHANGES_REQUESTED` at `e4b5bf94f` (WriteLog outside the registration block + missing `cflock` on the dedup registry); both addressed in `8c97f0b54` and confirmed resolved by Reviewer A's follow-up at `3d998ff9b`.

## Findings addressed

- **DI7 [Low] Routine plugin-load log entries routed to the security log instead of `wheels.log`** @ `vendor/wheels/Plugins.cfc:143-147` — `file="wheels_security"` → `file="wheels"`, matching the symmetric `PackageLoader.cfc` entry; zero `wheels_security` refs remain in `Plugins.cfc`.
- **DI10 [Medium] `$invokeOnPluginLoad` write-back replaces `application.wheelsdi`/`application.wo`/`application.$wheels` with deep clones** @ `vendor/wheels/Plugins.cfc:724` — `Duplicate(application)` → `StructCopy(application)` (the Adobe CF context-struct workaround only needs a plain struct). Live DI/config references now survive; the full write-back loop is deliberately retained because same-ref reassignment is a no-op while Adobe CF copies arrays by value even under shallow `StructCopy`, making changed-key detection unreliable there (documented inline).
- **DI12 [Low] `$initializeMixins` classifies components by substring match on the dotted path** @ `vendor/wheels/Plugins.cfc:866-876` — unanchored `FindNoCase("controllers", fullname)` → dotted-segment `ListFindNoCase`, with controllers checked before tests so the `wheels.tests._assets` fixtures stay honest. A model named `app.models.ControllerStats` no longer receives the controller mixin set.
- **U2 [High] All three plugin deprecation warnings link to dead documentation URLs** @ `vendor/wheels/Plugins.cfc:179, 250, 617` — dead `/docs/...` paths replaced with live versioned URLs (`https://guides.wheels.dev/v4-0-0/upgrading/3x-to-4x/`, `https://guides.wheels.dev/v4-0-0/digging-deeper/packages/`); both map to real pages under `web/sites/guides/src/content/docs/v4-0-0/`.
- **U3 [Medium] Stale and self-contradicting deprecation text in `Plugins.cfc` and the upgrade-check fix text** @ `vendor/wheels/Plugins.cfc:176-180, 614-618` + `cli/lucli/Module.cfc:4102, 4121` — future-tense "will be deprecated" and the abandoned "move to packages/ and copy to vendor/" staging design replaced with "deprecated as of 4.0, removed in 5.0" and the shipped remediation: packages installed under `vendor/` via `wheels packages add <name>`.
- **U6 [Low] Deprecation-warning channels are inconsistent and invisible in-app** @ `vendor/wheels/Global.cfc:2965` (`$deprecated(feature, message, docUrl)`) — one policy: warn once per feature per application to `wheels.log`, register in `application[appKey].deprecationWarnings` (try/catch best-effort), consumed by a new guarded + HTML-encoded debug-panel **Deprecations** section (`vendor/wheels/events/onrequestend/debug.cfm:412-430`). `Plugins.cfc:176, 614` and the `paginationLinks` warning (`vendor/wheels/view/links.cfc:240`) route through it. The MCP endpoint notice (protocol payload), the CommandBox upgrade command (separate runtime), and `wheels.Test` intentionally keep their own surfaces.

None — staleRefs checks confirmed all six findings were still live on `origin/develop` before this branch:

- `wheels_security` present at develop `vendor/wheels/Plugins.cfc:146` (DI7 live).
- Stale `"Migrate to packages/ + vendor/ activation model"` / `"Migrate to packages/ + vendor/ system"` fix text present at develop `cli/lucli/Module.cfc:4102` and `:4121` — the report's `:4015` citation was a line offset, not a stale finding; both real sites updated.

## Source

Internal multi-agent framework review 2026-06-09, wave 2, package **plugins-deprecation** (refs di-packages:7, di-packages:10, di-packages:12, upgrade-docs:2, upgrade-docs:3, upgrade-docs:6).

## Tests

New/updated specs (core suite, WheelsTest BDD):

- `vendor/wheels/tests/specs/global/deprecatedHelperSpec.cfc` — `$deprecated()` logs once per feature, registers in `deprecationWarnings`, dedupes on repeat calls.
- `vendor/wheels/tests/specs/pluginsMixinClassificationSpec.cfc` — dotted-segment classification, incl. the `app.models.ControllerStats` regression (fixtures under `vendor/wheels/tests/_assets/mixins_classification/`).
- `vendor/wheels/tests/specs/pluginsDeprecationMessagingSpec.cfc` — source-scan pins: live guide URLs, no future-tense/staging text, no `wheels_security` anywhere in `Plugins.cfc` (intentionally broad regression pin).
- `vendor/wheels/tests/specs/pluginsModernSpec.cfc` — extended for the `StructCopy` context behavior.

Local verification (Lucee 7 + SQLite via the worktree-safe docker single-bundle recipe): both behavioral specs RED-verified against pre-fix code and green post-fix; `plugins*`, `global`, `view`, and `events` bundles green. The `cli/lucli/Module.cfc` change is string-only. CI runs the full engine x DB matrix as the real gate.

## Cross-engine notes

- `$deprecated` is `public` with `$` prefix (mixin invariant 7 — private mixins are not integrated on Lucee/Adobe); the bare call in `links.cfc` resolves because `Controller` extends `wheels.Global`.
- DI10 keeps the plain-struct context required by the Adobe CF "no function members on `application`" workaround; `$installPluginLoadAPI` injects only a closure, which the `IsCustomFunction` write-back guard skips.
- The retained full write-back loop is the documented safe choice for Adobe CF's array copy-by-value semantics in struct copies (Cross-Engine Invariants item 6).
- Debug-panel output is HTML-encoded and existence-guarded; no `attributeCollection`, no inline-closure constructor args, no `Left(str, 0)`, no reserved-scope parameter names in the diff.
- One spec caveat (reviewer, non-blocking): `pluginsMixinClassificationSpec` relies on `GetMetadata` omitting `displayName` for undeclared components — verified on Lucee 7; Adobe/BoxLang behavior is gated by the CI compat matrix (the production framework already depends on the same semantics).

## Changelog


bpamiri:
Addressed the review in 8c97f0b54.

**Blocking 1 — WriteLog outside the registration block (Global.cfc):** Fixed as proposed. `WriteLog` now sits immediately after `ArrayAppend` inside the registration block, wrapped in its own try/catch, so it fires if-and-only-if registration just succeeded. On Reviewer B's nuance (an `else { WriteLog }` when the app struct is absent): chose silence deliberately — with no registry to dedup against, an unconditional log would fire on *every* call, which is the same failure mode the fix removes. All framework callers run after `application.wheels` is established; the tradeoff is documented in the docblock.

**Blocking 2 — no cflock (Global.cfc):** Fixed with the app-wide lock variant the review blessed (`lock name="wheels_deprecated_registry" type="exclusive" timeout="5"`, same form as the prior art at `TenantMigrator.cfc:159`). App-wide rather than per-feature on purpose: it also serializes the lazy `deprecationWarnings = []` creation, which per-feature lock names would leave racy across two different first features. Dedup read, append, and log all run inside the lock.

**Conventions — dual registry (Plugins.cfc:172):** Intentional. The per-instance array feeds the public `getDeprecationWarnings()` accessor (`Plugins.cfc:958`), which existing specs read (`pluginsModernSpec.cfc:805-905`). Added the requested inline comment.

**Tests — pluginsModernSpec cleanup:** Wrapped the assertion block in `try/finally` so a failing expectation can no longer leak `application.$wheelstestSharedRef` or the mutated `pluginComponentPath`.

**Tests — StructCopy + Adobe arrays:** Acknowledged, left as documented-only (as the review allowed). A reference-identity assertion on an array would legitimately diverge per engine — fail on Adobe (copy-by-value), pass on Lucee — so no single cross-engine-stable spec can pin it without per-engine forks.

**Docs — debug.cfm `application.wheels` vs `$appKey()`:** Rebutting the reachability, with the read documented inline instead of changed: `application.$wheels` exists only *during* `onapplicationstart`, and `onapplicationstart.cfc:387-388` reassigns the **same struct reference** to `application.wheels` then deletes `$wheels`. Init-time registrations (plugin loading) therefore land in the struct that *is* `application.wheels` by the time any `onrequestend` runs — the panel cannot observe a `$wheels`-keyed registry. Added a comment at the panel section so this doesn't get re-flagged.

**Reviewer B's missed issue — deprecatedHelperSpec closure scoping:** Fixed with the suggested carrier struct (`var state = {...}` read by reference, fields mutated — anti-pattern 10 pattern).

**Verification** (Lucee 7 + SQLite, worktree docker single-bundle recipe): `deprecatedHelperSpec` 4 pass (incl. dedupe through the new lock), `pluginsModernSpec` 33 pass, `pluginsDeprecationMessagingSpec` 3 pass, `pluginsMixinClassificationSpec` 2 pass, full `global` directory 119 pass — 0 fail / 0 error across all runs.

bpamiri:
Confirming the three follow-up review nits are addressed as of `8c0a2d4af`:

- **Carrier-struct state** in `pluginsMixinClassificationSpec.cfc` — `var state = {originalMixins = {}}` shared by `beforeEach`/`afterEach`, matching `deprecatedHelperSpec.cfc` and CLAUDE.md anti-pattern 10, with the same explanatory comment.
- **`CreateObject` instead of `new wheels.Plugins()`** in both `it()` blocks — skips `init()` so `$checkPluginsDeprecation()` no longer appends to `application.wheels.deprecationWarnings` as a cross-spec side effect. `$initializeMixins` only reads `application.wheels.mixins` (mocked) and `GetMetadata(variablesScope.this)`, so no initialised instance is required.
- **PR description** — the false "Reviewer verdict: approve" sentence is replaced with the accurate reviewer history (`CHANGES_REQUESTED` at `e4b5bf94f`, addressed in `8c97f0b54`, confirmed at `3d998ff9b`).

Independently verified locally (Lucee 7 + SQLite, single-bundle runs in a clean worktree of this branch): `pluginsMixinClassificationSpec` 2 pass, and the sibling `global.deprecatedHelperSpec` + `pluginsDeprecationMessagingSpec` + `pluginsModernSpec` bundles 40 pass — 0 fail, 0 error. I had prepared the identical fix in parallel; my push was rejected because the round-2 commit landed first, and the diff between the two is comment wording only, so no further push is needed.

PR #2941, opened 2026-06-10
fix(cli): upgrade-check exit code, breaker-scan coverage, ArgSpec positional gaps

bpamiri:
## Summary

Finalizes wave-2 package `cli-upgrade-argspec`: six findings across the `wheels upgrade check` scanner and the CLI ArgSpec positional-binding layer. `upgrade check` can now gate CI (non-zero exit on breaking findings + `--format=json`), the breaker scan covers the canonical eleven 3.x→4.x changes, the `wheels.Testbox` shim is documented/owned with a 5.0 removal target, and ArgSpec correctly binds LuCLI's gap-numbered positionals so flags between positionals no longer drop or reject arguments. Branch is one commit behind `origin/develop` (#2917, dispatch) with zero file overlap; merges clean.

## Findings addressed

- **U1 [High] `wheels upgrade check` exits 0 even when breaking changes are found** @ `cli/lucli/Module.cfc:4576` — throws `Wheels.UpgradeCheckFailed` after the report flushes (mirrors `validate()`'s `Wheels.ValidationFailed`), so breaking findings exit non-zero and gate CI. Adds `--format=json` for a machine-readable report (`jsonMode` @ `:4069`, JSON emission @ `:4515`); advisories never affect the exit code.
- **U4 [Medium] Breaker-scan coverage gaps vs the canonical eleven** @ `cli/lucli/Module.cfc:4172-4220` — WireBox check now matches `wirebox.system.ioc` and scans the root `Application.cfc` + `config/` in addition to `app/` (`:4172-4185`); new breaking grep for removed `renderPage()`/`renderPageToString()` (`:4188-4196`); new advisories for the HSTS default flip (SecurityHeaders, #2081 @ `:4199-4208`) and the CSRF SameSite cookie attribute (protectsFromForgery, #2035 @ `:4210-4220`).
- **U5 [Medium] The `wheels.Testbox` shim is unowned — undocumented, unscanned, silent, load-bearing** @ `vendor/wheels/Testbox.cfc:1-13` — CFC docblock declares deprecation since 4.0 and a 5.0 removal target; `mapperModernSpec` migrated off the shim to `wheels.WheelsTest` (`vendor/wheels/tests/specs/mapperModernSpec.cfc:1`) so the core suite no longer depends on it; the upgrade-check "Old test base class" grep now matches `wheels.Testbox` and single-quoted `extends` forms (`cli/lucli/Module.cfc:4141-4163`); the 3x-to-4x guide documents the shim (`web/sites/guides/src/content/docs/v4-0-0/upgrading/3x-to-4x.mdx`).
- **U7 [Low] Notes (loop-var shadowing, dead plugins check, missing guide links, no adapter regression coverage)** @ `cli/lucli/Module.cfc:4430-4433` — `scanTargets` loop var renamed to `scanTarget` so it no longer shadows the target-version local; dead 2.x→3.x `app/plugins` check fixed to the real root `plugins/` path (`:4136`, deduped on 2.x→4.x jumps where the 4.x entry covers it); breaking-finding output now links the version-appropriate upgrade guide and the `wheels-legacy-adapter` soft landing (`:4539`); new in-repo regression spec for legacy-adapter registry resolution (`cli/lucli/tests/specs/packages/LegacyAdapterResolutionSpec.cfc`).
- **CLI-D5 [Low-Medium] Instance-level `__arguments` set on delegation but never cleared — stale argv replays under the long-running MCP server** @ `cli/lucli/Module.cfc:109-111` — `structuredArgs()` now consumes `__arguments` once (cleared on every call), so a delegation stash can never replay stale argv into a later zero-arg tool call. The only two delegation sites (`generate()` @ `:414-415`, `create()` @ `:1046`) re-stash immediately before `return new()`, so delegation is unaffected.
- **CLI-D1 [High] LuCLI's gap-numbered positionals break fixed-index / stop-at-gap consumers — positionals after a flag silently dropped or wrongly rejected** @ `cli/lucli/services/ArgSpec.cfc:94-99,150,179` — `parse()`/`toArgv()` now bind positionals by collecting and numerically sorting `arg<N>` keys (`$positionalIndices()`, the proven `parseTestArgs` pattern) instead of fixed-index probing / stop-at-gap loops. Fixes `wheels new --port=3000 blog` (app name was ignored, then threw `Wheels.InvalidArguments`) and `wheels g scaffold Post --force title:string body:text` (every column after the flag was silently dropped).

- **CLI-D1 sub-item 3 — the `getArgs()` `arg1`-gate dropping all args on named-only invocations for the 8 dispatcher commands** (generate/migrate/start/create/deploy/packages/db/browser): already fixed on `origin/develop` before this branch by the ArgSpec `toArgv` migration (#2872). Evidence: `git grep 'toArgv(structuredArgs' origin/develop -- cli/lucli/Module.cfc` hits the dispatcher sites (develop `Module.cfc:364, 458, 740, 1019, 2055`, …). This branch builds on that path and fixes the remaining stop-at-gap defect inside `toArgv` itself.
- Staleness spot-check on the addressed findings (confirming they were live, not stale): pre-branch `origin/develop` still had the dead `app/plugins` scan path (`cli/lucli/Module.cfc:4101` on develop) and no `Wheels.UpgradeCheckFailed` throw (`git grep 'Wheels.UpgradeCheckFailed' origin/develop` returned no hits).

## Source

Internal multi-agent framework review 2026-06-09, wave 2 package `cli-upgrade-argspec`.

## Tests

- `vendor/wheels/tests/specs/cli/UpgradeCheckCoverageSpec.cfc` — extended to pin the full canonical-eleven pattern set (WireBox incl. `wirebox.system.ioc`, renderPage grep, HSTS + SameSite advisories), the `Wheels.UpgradeCheckFailed` exit-code throw, and the `--format=json` plumbing. Red pre-fix.
- `cli/lucli/tests/specs/services/ArgSpecSpec.cfc` — gap-numbered positional binding for `parse()`/`toArgv()` (flag-between-positionals collections). Red pre-fix.
- `cli/lucli/tests/specs/commands/CommandArgParsingSpec.cfc` + `cli/lucli/tests/_fixtures/commands/ModuleArgvProbe.cfc` — end-to-end flag-between-positionals repro through the module dispatch path and `__arguments` consume-once behavior. Red pre-fix.
- `cli/lucli/tests/specs/packages/LegacyAdapterResolutionSpec.cfc` — new regression coverage for `wheels-legacy-adapter` registry resolution.
- Local verification (Lucee 7 + SQLite, worktree docker single-bundle recipe): core `cli` specs 78/78, `mapperModernSpec` 27/27, CLI suite 774 pass with only the pre-existing environment-dependent SSH/live-server failures. Full engine × DB matrix deferred to CI (runs per PR).

## Cross-engine notes

- ArgSpec changes are pure struct/array CFML using the collect-and-sort pattern already proven cross-engine in `parseTestArgs()`; behavior-identical for gap-free collections, so the ~12 existing `toArgv`/`parse` call sites are unaffected on every engine.
- `mapperModernSpec` base-class swap is alias-to-alias (`wheels.Testbox` extends `wheels.WheelsTest`) — zero behavior change, keeps the shim deletable at 5.0 without crashing the bundle.
- Specs follow the documented gotchas: WheelsTest BDD, no inline closures as constructor named args, `##` escaped in string literals, no `local.X`-in-catch persistence reliance, no reserved-scope parameter names.

## Changelog

### DIFF
--- a/cli/lucli/Module.cfc
+++ b/cli/lucli/Module.cfc
@@ -98,12 +98,20 @@ component extends="modules.BaseModule" {
 	 * `create` → `new`, and unit tests). That array is reconstructed into the
 	 * same structured shape LuCLI would have produced, so a command behaves
 	 * identically whether LuCLI dispatched it or another command delegated to it.
+	 *
+	 * `__arguments` is consume-once: it is cleared on every call so a stale
+	 * stash can never replay. One-shot CLI runs hid the leak, but the stdio
+	 * MCP server (`wheels mcp wheels`) is a persistent process — after any
+	 * delegating call (create/generate app → new) a later zero-arg tool call
+	 * (e.g. wheels_test()) would otherwise re-parse the stale argv and turn
+	 * into a completely different invocation.
 	 */
 	private struct function structuredArgs(struct callerArgs = {}) {
+		var raw = __arguments ?: [];
+		__arguments = [];
 		if (!structIsEmpty(arguments.callerArgs)) {
 			return arguments.callerArgs;
 		}
-		var raw = __arguments ?: [];
 		return argvToCollection(isArray(raw) ? raw : []);
 	}
 
@@ -2724,10 +2732,12 @@ component extends="modules.BaseModule" {
 		var parsed = new services.ArgSpec()
 			.positional(name = "subcommand", default = "")
 			.option(name = "to", default = "")
+			.option(name = "format", default = "")
 			.parse(arguments.coll);
 		return {
 			isCheck = lCase(parsed.subcommand) == "check",
 			targetVersion = parsed.to,
+			format = parsed.format,
 			sawTo = structKeyExists(arguments.coll, "to"),
 			sawDryRun = structKeyExists(arguments.coll, "dry-run")
 		};
@@ -2744,9 +2754,14 @@ component extends="modules.BaseModule" {
 	 * Despite occasional appearances in older help output, `--dry-run` is not
 	 * supported — the command is already read-only by design.
 	 *
+	 * Breaking findings throw Wheels.UpgradeCheckFailed after the report is
+	 * printed, so the command exits non-zero and can gate CI. Advisory
+	 * (opt-in recommendation) findings never affect the exit code.
+	 *
 	 * Examples:
-	 *   wheels upgrade check                 - scan against the latest stable release
-	 *   wheels upgrade check --to=4.0.0      - scan against a specific target version
+	 *   wheels upgrade check                       - scan against the latest stable release
+	 *   wheels upgrade check --to=4.0.0            - scan against a specific target version
+	 *   wheels upgrade check --format=json         - machine-readable report (CI pipelines)
 	 */
 	public string function upgrade() {
 		var opts = parseUpgradeArgs(structuredArgs(arguments));
@@ -2760,6 +2775,10 @@ component extends="modules.BaseModule" {
 				& nl
 				& "Options:" & nl
 				& "  --to=<version>    Target Wheels version (default: latest stable)" & nl
+				& "  --format=json     Emit a machine-readable JSON report" & nl
+				& nl
+				& "Exit status:" & nl
+				& "  Non-zero when breaking changes are found (advisories never fail the check)." & nl
 				& nl
 				& "Unsupported flags:" & nl
 				& "  --dry-run is not supported — the command is already read-only," & nl
@@ -2781,7 +2800,7 @@ component extends="modules.BaseModule" {
 			return help;
 		}
 
-		return runUpgradeCheck(opts.targetVersion);
+		return runUpgradeCheck(opts.targetVersion, opts.format);
 	}
 
 	// ─────────────────────────────────────────────────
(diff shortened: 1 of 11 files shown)


bpamiri:
Non-blocking follow-ups from the wave-2 internal review (reviewer approved; these are fast-follow candidates, not blockers):

- [ ] `cli/lucli/Module.cfc` ~`:4096` — in `--format=json` CI mode, a GitHub fetch failure (no `--to` pinned) still `return ""` with exit 0, so a pipeline silently passes when the scan never ran. By the module's own `validate()` rationale ("an analyzer crash must not exit 0") this path should throw. Also: the error JSON is emitted with colour `"yellow"` while the main JSON doc is uncolored — risks ANSI pollution of machine output.
- [ ] `cli/lucli/Module.cfc` ~`:4480` — matchEntry structs use unquoted keys, so the JSON `breaking`/`advisories` entries serialize with Lucee's uppercase keys (`DESCRIPTION`/`FIX`/`MATCHES`) while top-level keys are case-preserved. Quote the keys for a stable contract before pipelines depend on it.
- [ ] On a 2.x→4.x jump the "Old test base class" check is appended by both version blocks (duplicate finding/passed entry) — the plugins-dir dedup guard was added but the same dedup for the test-base check was not (pre-existing dupe).
- [ ] The renderPage breaking pattern `renderPage(ToString)?\s*\(` has no leading boundary, so a user-defined `customRenderPage(` false-positives as a breaking finding — now consequential since breaking findings fail CI.
- [ ] Theoretical/cosmetic: `$positionalIndices` accepts a non-canonical key like `arg007` via the regex but then looks up `arg7` and would throw. LuCLI never emits zero-padded indices, so no real-world impact.

PR #2945, opened 2026-06-10
fix(dispatch): gate migrator and MCP endpoints, fix payload guard, purge sessions

bpamiri:
## Summary

Gates the publicly dispatched dev-tool endpoints and fixes the migrator GUI / `/wheels/ai` / MCP session issues from the public-devtools package of the 2026-06-09 framework review. The destructive migrator command endpoint (dispatched outside the middleware pipeline and controller CSRF layer) is now loopback-gated with a fail-closed anti-CSRF header token; the deprecated HTTP MCP surface no longer shells out via `cfexecute`; the migrator GUI JSON payload and JS refresh URLs are fixed; `/wheels/ai` serves project context in-process instead of via loopback HTTP self-requests; and the MCP `SessionManager` purges expired sessions on the request path.

## Findings addressed

- **SEC-3 [High, dev-scoped] Destructive migrator endpoints have no CSRF token, no localhost gate, no password** @ `vendor/wheels/public/migrator/command.cfm:13-79` — added the `consoleeval.cfm` loopback (`InetAddress.isLoopbackAddress()`) + `X-Forwarded-For` checks, plus a fail-closed, constant-time-compared anti-CSRF token round-tripped in the `X-Wheels-Csrf-Token` header (generated app-wide when the migrator GUI renders). All four GUI XHR paths send the header. The reload password was deliberately not required: the GUI has no password UX, and the loopback + custom-header token closes the cross-site auto-submit-GET vector the finding describes.
- **SEC-6 [Medium] Deprecated `/wheels/mcp` shells out via `cfexecute` with a literal-string localhost list** @ `vendor/wheels/public/views/mcp.cfm:54-60`, `vendor/wheels/public/mcp/McpServer.cfc:1275` — `mcp.cfm` now uses `InetAddress.isLoopbackAddress()` (covers `::ffff:127.0.0.1`); `McpServer.cfc` no longer contains any `cfexecute` — CLI-backed tools return a pointer to the canonical stdio MCP server (`wheels mcp wheels`) instead.
- **P5 [Medium] Always-false `structKeyExists(variables, "local")` guard kept `remainingMigrations` out of the JSON payload** @ `vendor/wheels/public/views/migrator.cfm:19-33, 68-70` — `remainingMigrations` is now stored unscoped like its siblings (reset each request) and serialized into the payload, along with `outOfSequenceMigrations` which `updateOutOfSequenceBanner()` reads.
- **P6 [Medium] Migrator GUI JS rebuild hardcoded `/wheels/migrator/...` URLs bypassing `urlFor()`** @ `vendor/wheels/public/views/migrator.cfm:150-272` and the JS refresh code — URL templates are now emitted once via `urlFor()` into data attributes and resolved client-side, so the post-refresh buttons survive `URLRewriting=off` and non-root context paths.
- **P12 [Medium, perf] `/wheels/ai` issued serial loopback HTTP self-requests to `/wheels/routes|plugins|migrator|info`** @ `vendor/wheels/public/views/ai.cfm:503+` — `handleProjectContext()` now reads the application scope / migrator directly, and the mode handlers include the target views in-process (each already emits JSON and aborts under `format=json`). Zero `new http()` / `cgi.server_port` self-request sites remain.
- **P13 [Low, perf] MCP `SessionManager` never purged expired sessions on the request path** @ `vendor/wheels/public/mcp/SessionManager.cfc:17, 95-100` — `createSession()` now calls `cleanupExpiredSessions()`; `init()` accepts a default-preserving `sessionTimeout` override so the purge is unit-testable.

None — all six findings were live on `origin/develop` pre-fix. Review spot-checked the pre-fix state for the always-false `variables`/`local` guard, the literal loopback string list, and both `cfexecute` sites.

## Source

Internal multi-agent framework review 2026-06-09, wave 2, package `public-devtools`.

## Tests

New spec: `vendor/wheels/tests/specs/security/PublicDevtoolsGatingSpec.cfc` — source-inspection assertions for the `.cfm` gates (mirroring `PublicComponentProductionSpec`: loopback gate, XFF check, CSRF header requirement, constant-time compare, no hardcoded JS URLs, no always-false guard, no `cfexecute`, no loopback self-requests) plus behavioral specs for `SessionManager` purge-on-create (negative-timeout expiry and live-session retention). The spec is red against pre-fix develop (6 hardcoded URLs, guard present, cfexecute present) and green on this branch.

Local docker single-bundle verification was deferred due to docker contention; CI compat-matrix is the gate. The spec avoids known bundle-compile risk patterns (no `##` in string literals needing escape, no literal CFML tag text in strings, no inline closures as constructor named args).

## Cross-engine notes

- The command.cfm gate is a verbatim copy of `consoleeval.cfm`'s loopback + XFF blocks, which already pass the full engine matrix.
- `isLocalhost` is initialized `false` before the `try`, so the BoxLang catch-scope `local.X` invariant does not apply (fail-closed regardless).
- Tag balance verified on all four touched `.cfm` files.
- `$isSafeArgument` in `McpServer.cfc` is still used by three other call sites and was intentionally retained.

## Changelog

### DIFF
--- a/vendor/wheels/public/mcp/McpServer.cfc
+++ b/vendor/wheels/public/mcp/McpServer.cfc
@@ -1271,108 +1271,23 @@ Provide migration code following Wheels conventions."
 	}
 
 	private string function executeCommand(required string command) {
-		// Structural allowlist: only permit known wheels subcommands with safe arguments
-		var allowedSubcommands = "g,generate,test,server,dbmigrate,db:seed,jobs,reload,info,new,init,deps";
-
-		if (!len(trim(arguments.command))) {
-			return "Error: Empty command.";
-		}
-
-		if (!reFind("^wheels\s", arguments.command)) {
-			return "Error: Only 'wheels' commands are allowed.";
-		}
-
-		// Strip "wheels " prefix once — used by both primary and fallback paths.
-		// 3-arg mid() is required for Adobe CF compatibility; Lucee/BoxLang accept 2 args.
-		local.strippedArgs = trim(mid(arguments.command, 7, len(arguments.command)));
-
-		if (!len(local.strippedArgs)) {
-			return "Error: Missing subcommand. Allowed: #allowedSubcommands#";
-		}
-
-		// Parse into parts and validate subcommand against allowlist
-		local.parts = ListToArray(local.strippedArgs, " ");
-		local.subcommand = local.parts[1];
-
-		if (!ListFindNoCase(allowedSubcommands, local.subcommand)) {
-			return "Error: Unknown subcommand '#EncodeForHTML(local.subcommand)#'. Allowed: #allowedSubcommands#";
-		}
-
-		// Validate each subsequent argument for shell metacharacters
-		for (var i = 2; i <= ArrayLen(local.parts); i++) {
-			if (!$isSafeArgument(local.parts[i])) {
-				return "Error: Argument contains disallowed characters.";
-			}
-		}
-
-		try {
-			// Get the application root directory using Application.cfc mappings
-			// The /app mapping points to the application's app directory (e.g., /project/app/)
-			// So /app/../ gives us the project root directory
-			local.appPath = expandPath("/app/../");
-
-			// Fallback: If /app mapping doesn't exist or doesn't point to a valid location,
-			// use the traditional detection method
-			// Wheels project marker: wheels.json (4.0+) or legacy box.json (pre-4.0).
-			// Either is a sufficient signal that we're in a Wheels project root.
-			if (!directoryExists(local.appPath) || (!fileExists(local.appPath & "wheels.json") && !fileExists(local.appPath & "box.json") && !fileExists(local.appPath & "public/Application.cfc"))) {
-				// Fallback to manual path detection from webroot
-				local.appPath = expandPath("/");
-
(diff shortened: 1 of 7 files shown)


bpamiri:
Both review points are addressed:

- **CSRF token init race (blocking)** — fixed in 1aad4a96c: the one-time `$migratorCsrfToken` generation is now wrapped in the suggested double-checked `cflock(scope="application", type="exclusive", timeout=5)`, so a second concurrent first-load waits and re-reads the first writer's token instead of clobbering it.
- **`JSStringFormat()` nit** — applied in 878acf25c to both JS embeds of the token (`command.cfm:152` and `views/migrator.cfm`), making the safe-embedding intent self-documenting. Output is byte-identical for the hex token.

Verified on Lucee 7 + SQLite (prebuilt docker image, dir-only mount): `PublicDevtoolsGatingSpec` 14/14, full `security` area 211 pass / 0 fail / 0 error, `/wheels/migrator` renders (confirms the `cflock(){}` script syntax compiles — no spec executes that page), and the command-endpoint round-trip behaves: non-loopback POST → 403, loopback POST with missing/bogus token → 403, loopback POST with the page's real token → 200 confirmation flow. No gating was weakened.

PR #2951, opened 2026-06-10
fix(cli): SecretResolver Left(str,0) crash and secret parsing corruption

bpamiri:
## Summary

Fixes the two `deploy-secretresolver` findings from the 2026-06-09 internal framework review: a `Left(str, 0)` crash path (Cross-Engine Invariant 8) in the deploy secrets CLI/resolver, and a `SecretResolver` parsing design that silently dropped or corrupted secrets. The resolver no longer subtracts a baseline environment snapshot; it scans candidate keys from the secrets file itself, reads each declared key back from the sourced bash as `KEY<US>VALUE<RS>` records, and throws typed errors (`SecretResolver.BashUnavailable`, `SecretResolver.ResolutionFailed`) instead of silently yielding zero secrets.

## Findings addressed

- **DEP-8 — `Left(str, 0)` crash on keyless `=` lines** @ `cli/lucli/services/deploy/cli/DeploySecretsCli.cfc:68` — `extract()` now requires `eq > 1` so a line starting with `=` cannot reach `left(line, 0)`, which crashes Lucee 7. The two equivalent sites in `cli/lucli/services/deploy/lib/SecretResolver.cfc` are removed entirely by the parser rewrite; the new `$parseRecords()` guards with `sep > 1`.
- **DEP-9 — dropped and corrupted secrets** @ `cli/lucli/services/deploy/lib/SecretResolver.cfc:71-160` (`$resolveFile` / `$candidateKeys` / `$parseRecords` / `$runBash`):
  - Keys that also exist in the parent environment (user export, CI var) now resolve to the file's value instead of being dropped — registry login no longer proceeds with an empty password.
  - Multi-line secrets (TLS certs, SSH keys) survive intact via US/RS record delimiters instead of newline-split `env` diffing; base64 continuation lines are filtered by a bash `${!k+x}` set-check so they are no longer misparsed as keys. (NUL records were the first choice, but Lucee's `chr(0)` yields an empty string, so RS `chr(30)` is used.)
  - Bash failures surface as `SecretResolver.BashUnavailable` (unlaunchable bash) or `SecretResolver.ResolutionFailed` (non-zero exit, with stderr) instead of returning an empty struct.
  - `arrayContainsNoCase()` (avoided elsewhere in the deploy tree for engine availability) removed; candidates are deduped case-sensitively, matching bash's case-sensitive resolution.

Candidate keys are regex-restricted to `[A-Za-z_][A-Za-z0-9_]*` before interpolation into the bash script (no command injection); the file path remains single-quote-escaped.

None — `staleRefs` is empty for this package. Both develop baselines (the `eq < 1` / `eq > 0` sites in `SecretResolver.cfc` and the `DeploySecretsCli.cfc:68` site) were confirmed present pre-fix.

## Source

Internal multi-agent framework review 2026-06-09, wave 2, package `deploy-secretresolver`.

## Tests

- `cli/lucli/tests/specs/deploy/lib/SecretResolverSpec.cfc` — four new cases: parent-env-overridden key resolves to file value; multi-line cert value survives intact; base64 `=`-padded continuation line not misparsed as a key; export-prefixed declarations resolve.
- `cli/lucli/tests/specs/deploy/cli/DeploySecretsCliSpec.cfc` — one new case: `extract()` ignores a keyless line starting with `=` (the `Left(str, 0)` guard).
- Four of the five specs fail against the pre-fix code; the export-form spec was already green pre-fix (the old `set -a`/`source`/`env` flow handled export form) and is kept as regression coverage for the new candidate scanner.
- Local verification: the record protocol was exercised on macOS bash 3.2 and the candidate-scan/record-parse logic run red→green on the bundled Lucee (LuCLI 0.3.17) in script mode — the pre-fix code crashes with `parameter 2 of the function left can not be 0`. These are CLI specs (not core-framework specs), so the worktree docker single-bundle recipe doesn't apply; the full CLI suite runs in CI (`test-cli`), which is the real gate.

## Cross-engine notes

- Invariant 8 (`Left(str, 0)` crashes Lucee 7) is the subject of DEP-8; all new code guards length before `left()` (`eq > 1`, `sep > 1`). No new unguarded `Left(str, N)` sites.
- RS `chr(30)` / US `chr(31)` record delimiters chosen over NUL because Lucee's `chr(0)` yields an empty string.
- No reserved-scope parameter names, no inline closures as constructor named args, no `attributeCollection` usage, no `arrayContainsNoCase` (engine availability).
- Specs follow sibling style in `cli/lucli/tests/specs/deploy/`.

## Changelog

### DIFF
--- a/cli/lucli/services/deploy/cli/DeploySecretsCli.cfc
+++ b/cli/lucli/services/deploy/cli/DeploySecretsCli.cfc
@@ -65,7 +65,9 @@ component {
         if (!len(key)) return "";
         for (var line in listToArray(text, chr(10), false)) {
             var eq = find("=", line);
-            if (eq > 0 && left(line, eq - 1) == key) {
+            // eq > 1, not > 0: a line starting with '=' has no key, and
+            // left(line, 0) crashes Lucee 7 (Cross-Engine Invariant 8).
+            if (eq > 1 && left(line, eq - 1) == key) {
                 return mid(line, eq + 1, 99999);
             }
         }
--- a/cli/lucli/services/deploy/lib/SecretResolver.cfc
+++ b/cli/lucli/services/deploy/lib/SecretResolver.cfc
@@ -25,12 +25,19 @@
  * A present file on a machine where bash can't run throws
  * SecretResolver.BashUnavailable instead of silently resolving zero
  * secrets (which would let callers proceed with empty credentials).
+ * A failing command inside the file (e.g. `$(op read …)` when not signed
+ * in) throws SecretResolver.ResolutionFailed instead of silently
+ * exporting the key with an empty value.
  */
 component {
 
     public SecretResolver function init(struct opts = {}) {
         variables.projectRoot = arguments.opts.projectRoot ?: expandPath("./");
         variables.destination = arguments.opts.destination ?: "";
+        // Override when bash isn't reachable as plain `bash` on PATH
+        // (e.g. a non-standard Git Bash install on Windows). Also the
+        // seam the BashUnavailable spec uses.
+        variables.bashCmd = arguments.opts.bashCmd ?: "bash";
         variables.resolved = $loadAll();
         return this;
     }
@@ -76,6 +83,13 @@ component {
 
         // `set -a` exports every assignment made while sourcing; the file's
         // own stdout is discarded so it can't corrupt the record stream.
+        // `set -e` makes a failing command inside the file — most importantly
+        // an assignment whose $(cmd) substitution fails, like `$(op read …)`
+        // when not signed in — abort sourcing with a non-zero exit, so it
+        // surfaces as ResolutionFailed below instead of silently exporting
+        // the key with an empty value. (Without -e, bash only reports the
+        // status of the file's LAST statement, and even that is ignored
+        // because the script continues into the for-loop.)
         // `${!k+x}` (set-check on the indirected name) filters out candidate
         // keys bash never actually set — e.g. base64 continuation lines of a
         // quoted multi-line value that merely look like assignments.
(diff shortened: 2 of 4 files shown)


bpamiri:
Addressed the review in f765455c9:

**Deadlock in ``** — fixed with the recommended `pb.redirectError(File)` temp-file sink, read after `waitFor()`; cleanup is in a `finally` block per Reviewer B's note so the file is deleted even when `waitFor()` or the throw paths interrupt. Verified red/green locally on the bundled Lucee: the old sequential drain blocks past 30s on a 70 KB-stderr subprocess; the new shape completes in milliseconds with stdout intact and no `wheels-secret-err*` files left in the temp dir.

**Error-path tests** — three new specs in `SecretResolverSpec.cfc`:
- `ResolutionFailed` on a failing subshell (`BROKEN=$(exit 1)`, the suggested vector)
- `ResolutionFailed` when the failing command is mid-file
- `BashUnavailable` via a new `opts.bashCmd` seam pointing at a nonexistent binary, so the real `ProcessBuilder.start()` throw site is exercised rather than a stubbed override

One finding surfaced while implementing the suggested `ResolutionFailed` vector: it didn't fire against the script as merged-candidate — without errexit, bash continues past a failing `$(cmd)` substitution and exports the key with an empty value (script exit 0), i.e. the silent-empty-credential mode this PR set out to eliminate. The resolution script is now `set -ae`, so any failing command while sourcing (a not-signed-in `op read`, a locked `bw`, …) aborts with a non-zero exit and surfaces as `ResolutionFailed` with stderr in the detail. Verified on macOS bash 3.2 (single-line, mid-file, and success paths), and all ten pre-existing behaviors re-verified green under errexit.

bpamiri:
Noted on Reviewer B's round-2 follow-up: `proc.waitFor()` with no deadline means an interactively-blocking secrets command (e.g. `op read` prompting for a master password) would hang the resolution thread rather than fail. Agreed it's out of scope here — `waitFor(long, TimeUnit)` + `destroyForcibly()` surfacing as `ResolutionFailed` after a deadline is the obvious shape — and it applies to the whole deploy subprocess surface, so it deserves its own issue rather than a rider on this PR. Human sign-off before merge per B's note is with @bpamiri.

PR #3024, opened 2026-06-11
docs(web/blog): Anatomy of a Wheels Package

bpamiri:
Publishes "Anatomy of a Wheels Package".

- slug: `anatomy-of-a-wheels-package`
- live after merge + Cloudflare build: https://blog.wheels.dev/posts/anatomy-of-a-wheels-package

Opened automatically by wheels-publishing-admin on initial publish.

### DIFF
--- a/web/content/blog/posts/anatomy-of-a-wheels-package.md
+++ b/web/content/blog/posts/anatomy-of-a-wheels-package.md
@@ -2,15 +2,14 @@
 title: Anatomy of a Wheels Package
 slug: anatomy-of-a-wheels-package
 publishedAt: '2026-06-11T18:34:00.000Z'
-updatedAt: '2026-06-11T18:34:45.000Z'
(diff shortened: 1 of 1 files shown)


bpamiri:
Both findings addressed in d310db130:

1. **categories frontmatter** — fixed here and at the source: the publishing admin's legacy import had stored the literal string `[]` in the categories column for the three imported drafts; the db rows are cleaned (the exporter emits a proper `categories: []` for empty input — verified in this commit), so the two scheduled follow-up posts in this series won't ship the same defect.
2. **fence titles** — all 8 `title="..."` annotations moved into the preceding prose per the live-post convention (the three WheelsGreeter.cfc variants each now name file + iteration in the prose). The two scheduled posts got the same treatment (4 + 2 occurrences).

Also took the DCO note: this follow-up commit is signed off; adding `-s` to the publishing pipeline's commits is on the Phase 2 backlog.

Issue #2963, opened 2026-06-10
roadmap: high-impact CLI/MCP tooling-honesty gaps — failure exit codes (incl. upgrade check) + MCP tool input schemas


bpamiri:
## Problem

The wave-2 framework review's feature-set assessment (§5.2) flags two **High-impact** tooling-honesty/discoverability gaps. Both undermine the AI/CI story Wheels otherwise leads on. Verified against `origin/develop` @ `8971094`.

### 1. CLI failure exit codes (High / Medium effort)
`test`/`migrate`/`validate` (and the upgrade-check scanner) print red but return exit 0 — false-green CI. The throw-for-nonzero pattern already exists in-repo. A broad fix (#2890, "CLI exit non-zero on user-error paths") landed for the CLI generally, but coverage is incomplete:
- **U1 [High]** `wheels upgrade check` still exits 0 even when breaking changes are found — `cli/lucli/Module.cfc:4315-4355` ends the loop with `return "";`, no `throw` (contrast the module's `throw(message=...)` convention at `:2003`). The comment at `:4131-4134` implies breaking findings fail CI; they don't. The primary upgrade-safety tool cannot gate CI, and there's no `--format=json`.

Audit `test`, `migrate`, `validate`, and `upgrade check` for nonzero exit on failure and confirm #2890 actually reaches each; add `--strict`/`--ci` and JSON output where missing.
**Prior art:** pytest/RSpec; artisan `Command::FAILURE`; `db:migrate` abort.

### 2. MCP tool input schemas (High / Medium effort)
All ~20 auto-discovered MCP tools advertise empty `properties` with `additionalProperties:false`, so parameters are undiscoverable to MCP clients; two `$`-prefixed helpers also leak as tools. Tools are auto-discovered from `cli/lucli/Module.cfc` public functions but no per-tool input schema is emitted (no `inputSchema`/`properties` population found in `Module.cfc`).
Derive typed input schemas from each tool function's signature (matching the now-landed ArgSpec migration), and exclude `$`-prefixed helpers from the tool surface.
**Prior art:** FastMCP typed-signature schemas; GitHub MCP per-tool schemas.

## Impact

- CI cannot gate on `wheels upgrade check` (the primary 3.x→4.x safety tool) or any command #2890 missed — red output with a green exit is silently ignored by CI.
- MCP clients (including AI agents, the audience Wheels targets) cannot discover tool parameters, and two internal helpers pollute the tool list.

## Suggested approach

[...]

Issue #2215, opened 2026-04-22
wheels new: explicit WHEELS_FRAMEWORK_PATH should hard-fail when it doesn't exist


bpamiri:
Follow-up noticed while working on #2211.

## Current behavior

In [cli/lucli/Module.cfc:3454-3469](cli/lucli/Module.cfc) (`resolveFrameworkSource`), when the user explicitly sets \`WHEELS_FRAMEWORK_PATH\`:

\`\`\`cfm
var override = javaSystem.getenv(\"WHEELS_FRAMEWORK_PATH\");
if (!isNull(override) && len(trim(override))) {
    arrayAppend(variables.frameworkSearchPaths, override & \"  (from \$WHEELS_FRAMEWORK_PATH)\");
    if (directoryExists(override)) {
        return override;
    }
}
\`\`\`

If the path exists, it's used. **If it doesn't exist, the code silently falls through to the next search path** (project root → module root walk). The user's explicit instruction is ignored with no warning.

## Why that's wrong

If I explicitly set \`WHEELS_FRAMEWORK_PATH=/some/path\`, I'm telling the CLI \"use this, not something else.\" A typo or a stale path should be flagged — not silently overridden by whatever happens to be discoverable elsewhere. The current fallback can produce surprising results where one developer gets a different framework version than they expected.

## Proposed fix

When \`WHEELS_FRAMEWORK_PATH\` is set but \`directoryExists()\` is false, throw early:

\`\`\`
Error: WHEELS_FRAMEWORK_PATH is set to '/some/path' but that directory does not exist.
       Unset the variable to fall back to auto-discovery, or point it at a valid vendor/wheels/ source.
\`\`\`

Exit non-zero (same mechanism as #2211).

## Acceptance

- [ ] \`WHEELS_FRAMEWORK_PATH=/does/not/exist wheels new foo\` exits non-zero with a clear message
- [ ] Unsetting the variable restores the auto-discovery fallback behavior
- [ ] \`WHEELS_FRAMEWORK_PATH=/valid/path wheels new foo\` still works

## Related

- #2211 — the exit-code fix that made this discoverable

PR #3039, opened 2026-06-11
feat(cli): wheels upgrade applies the framework swap from the CLI bundle

### LINKED ISSUE #3035: wheels upgrade: apply mode — swap the app's vendored framework from the CLI bundle
## Problem

Today `wheels upgrade` is check-only. The bare verb prints usage, `wheels upgrade check` scans the app for breaking changes (read-only, #2941/#2963/#2967 added exit-code gating, `--strict`, and `--format=json`), and the actual framework swap is left to the user: the upgrade guides document downloading a zip and manually replacing `vendor/wheels/`, and the CLI's own hint — `brew upgrade wheels` — only upgrades the CLI binary, never the app's vendored framework copy. There is no supported one-command path from "CLI updated" to "app framework updated".

## Proposal (PR1 — this issue)

Give `wheels upgrade` an apply mode that swaps the app's `vendor/wheels/` with the framework bundled inside the installed CLI:

- **Bare `wheels upgrade` applies the swap.** Source = the CLI's bundled `vendor/wheels/` (resolved via `WHEELS_FRAMEWORK_PATH` override first, else walking up from the module's own install location — mirroring `resolveFrameworkSource()`, minus the project-root candidate, which is the swap *target*).
- **Backup first.** The existing `vendor/wheels/` is renamed to `vendor/wheels.bak-<yyyymmdd>-<HHmmss>` (collision counter on rapid re-runs) via an atomic Java `renameTo`, with a hard error if the rename fails — recovery is a single `mv`. `--nobackup` opts out.
- **Safety rails before any destructive step:** both source and target must sniff as a Wheels framework dir (`wheels.json`/`box.json` present); refuse when run outside a Wheels app (no `vendor/wheels/`); refuse when source and target resolve to the same directory (e.g. running inside the wheels repo checkout itself); `--to=<version>` must exactly match the bundled framework version or the command hard-errors.
- **Report old → new version** and the backup path on success.
- **`wheels upgrade check` keeps its current read-only behavior** unchanged, including `--strict`, `--format=json`, and the `Wheels.UpgradeCheckFailed` exit-code contract.

This unlocks the bleeding-edge flow: `brew uninstall wheels && brew install wheels-be && wheels upgrade` puts the BE framework into an existing app, with the stable copy parked in `vendor/wheels.bak-*`.

## Follow-up (PR2 — not in scope here)

Network download of arbitrary `--to=<version>` targets (resolved via `ReleaseChannel`), so apply mode can install versions the CLI doesn't bundle.

bpamiri:
Closes #3035. PR1 of the two-PR apply-mode plan (PR2 = network download of arbitrary `--to=` targets via `ReleaseChannel`; not in scope here).

## The story

Today `wheels upgrade` is check-only: the bare verb prints usage, and the help points at `brew upgrade wheels` — which upgrades the **CLI binary** but never the app's vendored framework copy. The actual swap is a documented manual zip dance. This PR closes that gap with the smallest trustworthy source: the framework already bundled inside the installed CLI.

- **`wheels upgrade apply`** — replaces the app's `vendor/wheels/` with the CLI's bundled framework. The old copy is parked at `vendor/wheels.bak-<yyyymmdd>-<HHmmss>/` (atomic Java `renameTo`, collision counter, hard error if the rename fails) so recovery is a single `mv`. `--nobackup` opts out. Reports `old -> new` versions, the backup path, and the exact (quoted) recovery command.
- **`wheels upgrade`** (bare) — prints concise usage steering at the two verbs and exits 0, same as before this PR. **Contract decision from review:** destructive commands deserve an explicit verb, and MCP clients calling `wheels_upgrade` with `{}` must never mutate — requiring `apply` fixes that transport-independently, and exit-0 bare matches the pre-PR behavior so existing CI invocations see usage text, not a surprise swap or a new failure.
- **`wheels upgrade check`** — unchanged read-only scan, including `--strict`, `--format=json`, and the `Wheels.UpgradeCheckFailed` exit-code contract. Only its closing hint changes: `Apply with: wheels upgrade apply` instead of the misleading `brew upgrade wheels`. Note: `wheels upgrade check --help` now prints usage instead of running the scan (help wins over any verb).

This unlocks the bleeding-edge flow:

```bash
brew uninstall wheels && brew install wheels-be
wheels upgrade apply      # BE framework into the app, stable parked in vendor/wheels.bak-*
```

## Failure-mode UX (review fix)

The plan is announced **before any mutation**, with the exact reserved backup destination and the recovery one-liner:

```
Backing up vendor/wheels -> vendor/wheels.bak-20260611-141502
If this is interrupted, restore with:
  rm -rf "/path/to/app/vendor/wheels" && mv "/path/to/app/vendor/wheels.bak-20260611-141502" "/path/to/app/vendor/wheels"
```

The backup path is reserved up front (`FrameworkUpgrader.reserveBackupPath()`, now public) and passed back into `applyUpgrade()`, so the announcement and the actual backup can never disagree. The create+copy step is wrapped: a mid-copy failure throws `Wheels.FrameworkUpgrader.CopyFailed` whose message names the partial-state target and the backup to restore from (quoted paths) — or, with `--nobackup`, says the old tree is gone and points at re-running `wheels upgrade apply` after fixing the cause. Module-level dispatch catches it and rethrows `Wheels.UpgradeApplyFailed` after printing, mirroring `validate()`'s print-then-throw convention (#2941).

## Safety rails (all fire before any mutation)

- Source **and** target must sniff as a Wheels framework dir — and the sniff is hardened (review fix): a bare `box.json` is no longer evidence (every CommandBox-era project has one). Required: `wheels.json` with a non-empty `version`, or `box.json` whose `version` is non-empty and whose `name`/`slug` (when present) identifies a wheels artifact. A generic app `box.json` (`{"name":"myapp", ...}`) is refused on both sides.
- Identity/containment guard: running inside the wheels repo checkout itself resolves the bundled source to the very directory being replaced — the backup rename (or `--nobackup` delete) would destroy the source mid-swap. Refused, along with either direction of nesting.
- Refuses outside a Wheels app (no `vendor/wheels/`).
- `--to=<version>` is an assertion: must equal the bundled framework version or the command errors with the brew/scoop pointer (PR2 lifts this).
- Typo'd subcommands (`wheels upgrade chekc`) and unknown flags hard-stop (`Wheels.InvalidArguments`); check-only flags on the apply verb (`--strict`, `--format`, `--dry-run`) nudge toward `wheels upgrade check` and refuse.

## Design notes

- The swap lives in `cli/lucli/services/FrameworkUpgrader.cfc`, isolated from `Module.cfc` (mirrors `FrameworkInstaller.cfc`) so file-level behavior is spec'd without the LuCLI runtime.
- Bundled-source resolution: `WHEELS_FRAMEWORK_PATH` override first (invalid path hard-fails, same semantics as `resolveFrameworkSource()` / #2215), else walk up from the module's own install location. The project-root candidate `resolveFrameworkSource()` prefers is deliberately skipped — it's the swap *target*.
- `parseUpgradeArgs` accepts `subcommand` as a **named** key as well as positionally: the MCP inputSchema (#2963) advertises it as a named property; `{subcommand: "apply"}` is the only MCP shape that may mutate, `{}` prints usage.
- `upgradeArgSpec()` documents the explicit-verb contract, so CLI parsing, MCP `tools/list` schema, and help can't drift.

## Tests

RED→GREEN against the local CLI harness (`tools/test-cli-local.sh`, Lucee 7 + SQLite):

- GREEN (this branch): **902 pass / 0 fail / 0 error**.
- RED (review-fix specs against the pre-fix implementation): 881 pass / 14 fail / 7 error — exactly the 21 new/flipped specs (hardened sniff, `reserveBackupPath`, `CopyFailed` contract incl. a real mid-copy failure simulated via an unreadable source file, bare-verb usage steer, apply-verb dispatch, pre-swap announcement).
- Core cli area (`tools/test-local.sh wheels.tests.specs.cli`): **79 pass** (was 78; +1 pin that the help advertises `wheels upgrade apply`).

New/updated specs:

- `FrameworkUpgraderSpec` (34) — hardened sniff matrix (generic `box.json` refused, version-only legacy `box.json` accepted, malformed/empty-version refused), version reads, swap, backup naming/collision, caller-reserved backup path honored, `--nobackup`, identity/containment refusals, fresh-install path, missing-parent refusal, and the two `CopyFailed` contracts (with-backup names the backup to restore; `--nobackup` says the tree is gone). Fixture cleanup now lives in `afterEach` hooks.
- `UpgradeApplyCommandSpec` (19) — help paths, bare-verb usage steer (with and without stray apply flags, vendor untouched), every pre-mutation refusal on the `apply` verb, the real swap + backup, the pre-swap announcement (exact backup dir + quoted recovery command, ordered before the swap summary in the output), `--to` assert-match, `--nobackup`, MCP named-`subcommand` dispatch for both verbs.
- `InfoCommandSpec` — bare verb pins the usage steer; `apply` over the empty stub pins the sniff refusal.
- Core `UpgradeCommandHelpSpec` — pins the explicit-verb surface (summary, hint mentions both verbs, `wheels upgrade apply` advertised, `--dry-run` gap named, backup convention referenced).

## Follow-ups

- PR2 (#3035): download arbitrary `--to=` targets via `ReleaseChannel`.
- Guides refresh: `command-line-tools/wheels-commands/upgrade/` and the upgrading guides still describe the check-only command and the manual zip swap.
- Cap the bundled-source walk-up depth and surface `@build.version@` more nicely in repo checkouts (noted on #3035).


bpamiri:
7f82d2af7 (pushed before the review posted) addresses notes 1–3:

1. **Copy failures now caught with pre-swap recovery guidance** — `applyUpgrade` wraps the file-copy in a try/catch; on failure it prints the reserved backup path (which was announced before any mutation started) alongside the exact restore command, then exits non-zero via `Wheels.UpgradeApplyFailed`. Users are never left with a half-swapped `vendor/wheels/` without a recovery path.

2. **Source sniff hardened** — the Wheels-identity check now requires `wheels.json` to be present *and* to contain a `"name": "wheels"` (or `"cfwheels"`) field. A bare directory with no manifest, or a manifest without the identity field, is rejected before the swap begins.

3. **Resolved MORE conservatively than suggested** — bare `wheels upgrade` never mutates on *any* transport, not just MCP. The suggestion was to guard MCP callers; instead the `apply` verb is required everywhere. `wheels_upgrade({})` and `wheels upgrade` both print usage steering and exit 0 — destructive commands deserve an explicit opt-in.

Note 4 is pinned in 0b4b6680d (`test(cli): pin the --no-backup normalization spelling on upgrade apply`).

Note 5 is now tracked in #3045.

Issue #3159, opened 2026-06-12
deploy: redact interpolated secret values in $raiseRemoteFailure command summaries


bpamiri:
`lib/SshClient.cfc:187-206` (`$raiseRemoteFailure`, mirrored byte-identical in `lib/FakeSshPool.cfc`) embeds the first 200 chars of the failed command in `Wheels.Deploy.RemoteExecutionFailed`. Registry login no longer carries the password in the command (#3008), but `env.clear` values — including values interpolated from `${SECRET}` tokens in `.kamal/secrets` via `ConfigLoader.$interpolate` — appear in `docker run ... -e 'KEY=value'` and therefore in exception messages that land in CI logs.

**Direction (deferred from #3008 by design):** track the set of resolved secret values per config load (SecretResolver already knows them) and replace each occurrence with `[REDACTED]` in the command summary before throwing; keep the FakeSshPool mirror in lockstep (tests assert the exact throw shape — see the MIRROR comment at `SshClient.cfc:182-185`).

**Acceptance:** a deploy failure whose command contains a `${SECRET}`-interpolated env value produces an exception message without the secret value; spec covers it via FakeSshPool.

Residue of #2956 (DEP-3 deferred slice).

PR #3206, opened 2026-06-13
fix(cli): redact interpolated secret values on the deploy accessory verb (#3159 leftover)

bpamiri:
## Summary

Follow-up to #3205 (merged). That PR wired `$registerSecretsForRedaction()` into `DeployAppCli` and `DeployMainCli`, but the standalone `wheels deploy accessory boot|reboot <name>` verbs route through `DeployAccessoryCli`, which shares no base class with the other two and never registered the resolver's secret values on its `SshPool`.

`AccessoryCommands.$envArgs` embeds `env.clear` values — including `${SECRET}`-interpolated ones (e.g. `MYSQL_ROOT_PASSWORD: '${DB_ROOT_PW}'`) — as `docker run ... -e 'KEY=value'` in argv. So a nonzero accessory exit surfaced the **raw** secret in the `Wheels.Deploy.RemoteExecutionFailed` message and CI logs: the exact leak #3159 closes, just on the accessory verb instead of the app verb.

(`wheels deploy setup`'s accessory phase was already safe — it dispatches via `DeployMainCli.$dispatch`, which does register. Only the standalone `accessory boot/reboot` verbs leaked.)

## Fix

`DeployAccessoryCli.$forEach` now calls `$registerSecretsForRedaction()` right after `load()` (ahead of any `$dispatch`), reading `loader.secretResolver().all()` the same way `DeployAppCli`/`DeployMainCli` do. The helper is a byte-for-byte mirror of the other two CLIs (guarded for pools that predate `$setSecretValues`).

## Test

A regression spec drives the leak path end-to-end: a postgres accessory with an `env.clear` value interpolated from a `${DB_ROOT_PW}` token, a failed `docker run`, asserting the value is scrubbed to `[REDACTED]` in the thrown message.

CLI suite (lucee7 / sqlite): **1080 pass, 0 fail**, 2 pre-existing/environmental errors (`SshClientSpec`/`SshPoolSpec` — `docker: command not found` from the docker-in-docker sshd fixture).

Refs #3159

### DIFF
--- a/cli/lucli/services/deploy/cli/DeployAccessoryCli.cfc
+++ b/cli/lucli/services/deploy/cli/DeployAccessoryCli.cfc
@@ -55,6 +55,13 @@ component {
             arguments.opts.configPath,
             {destination: arguments.opts.destination ?: ""}
         );
+        // Register resolved secret values so a nonzero `docker run ... -e
+        // KEY=value` exit can't leak an env.clear value interpolated from a
+        // ${SECRET} token into the RemoteExecutionFailed message / CI logs
+        // (#3159). The env.secret path already avoids -e via --env-file, but
+        // env.clear values still ride argv — AccessoryCommands.$envArgs emits
+        // them as `-e 'KEY=value'`.
+        $registerSecretsForRedaction();
         var accCmds = new modules.wheels.services.deploy.commands.AccessoryCommands(cfg);
         var dryRun = arguments.opts.dryRun ?: false;
         var targets = (arguments.opts.name == "all")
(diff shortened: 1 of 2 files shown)


bpamiri:
Superseded by #3205 (merged): the deferred-from-#3008 redaction of interpolated secret values in $raiseRemoteFailure summaries shipped there (redaction before the 200-char trim, SshClient/FakeSshPool mirror in lockstep), closing #3159. This propose-fix draft raced that campaign PR; #3205 landed first. Closing the duplicate.

bpamiri:
Correction: reopened — this is **not** superseded by #3205. #3205 wired secret-redaction into `DeployAppCli` + `DeployMainCli`; this PR closes the **third dispatch site, `DeployAccessoryCli`** (`wheels deploy accessory boot`), which #3205 left uncovered — env.clear values interpolated from `${SECRET}` tokens still ride `docker run -e KEY=value` on that path and would leak into a RemoteExecutionFailed summary / CI logs. Apologies for the premature close; it's a legitimate leftover-gap fix for #3159.

Issue #3213, opened 2026-06-16
Wheels 2.x > 4.x Performance Comparison


chapmandu:
Hi Team,

I have done a preliminary port of my application from 2.5 > 4.0.3. I am just wanting to know if there has been any performance comparisons or load testing done between the two versions. My initial test run is showing a considerably slower test suite execution.

Wheels 2.5 (RocketUnit)
Finished Tue Jun 16 2026 10:05:50 (Took 274 Seconds)

Wheels 4.0.3 (RocketUnit)
The full run took 1599 seconds (~27 minutes) — 1574 pass / 5 skip / 0 fail, finished at 08:03:19 yesterday.

Thanks for all your hard work on wheels framework. 💪 

Regards,
Adam

bpamiri:
@chapmandu Adam, would you be open to sharing a codebase to demonstrate the issue so I can run perf mon on it.

chapmandu:
Hi Peter,

Sorry but I can't share my employer's IP but would happily run some performance analysis on the code and share any findings. 

Are there any specific tools you'd like to run or metrics to gather? I'm running latest stable Lucee 5 on Tomcat. 

Regards,
Adam

bpamiri:
Ah, understood. I just need a rocketunit test suite. I'll create that on this side. But just out of curiosity, can you run your code on Lucee 7 to see if the results are any different.

bpamiri:
Hi @chapmandu — we found and fixed the main culprit. 🎯

**Root cause:** it wasn't per-*test* overhead, it was per-**object** overhead. Every `model().new()` and every row returned by a finder re-ran the framework's mixin integration from scratch — a directory scan of `vendor/wheels/model/`, a `createObject` + `getMetaData` for each of its ~18 files, and a re-resolve of all ~231 mixed-in methods — on *every* instance. A RocketUnit suite materializes huge numbers of objects, so it compounded badly. (`Controller`/`Mapper` had the same pattern.)

**Fix ([#3236](https://github.com/wheels-dev/wheels/pull/3236), merged to `develop`):** that work is now done once per application and reused. Measured on Lucee 7 + SQLite: model-instance creation dropped ~1.8× (2772 ms → 1513 ms for 2,000 instances) and the full framework suite went 33.3 s → ~22 s, with no behavior change. The win should be larger for an instance-heavy app suite like yours.

[...]

PR #3236, opened 2026-06-20
perf(model): cache mixin-integration plan to fix per-instance overhead (#3213)

bpamiri:
## What & why

Fixes the per-instance overhead behind #3213 (a 2.x RocketUnit suite running ~6× slower under 4.0.x).

### Root cause

A RocketUnit suite runs the whole suite in **one request** (`Test.cfc::$wheelsRunner`), so application-scoped caches (models, routes, schema) are built once — they are *not* the per-test cost. The cost is **per model-instance creation**, which a test suite does constantly:

`model("X").new()` and **every row** returned by `findAll`/`findOne`/`findByKey` → `model/create.cfc::$createInstance` → `$createObjectFromRoot(method="$initModelObject")` → `Injector.getInstance()`. Model components aren't singletons, so the injector takes the **transient** path: `createObject` + **`init()`** + `onDIcomplete()` *every time*. `Model.cfc::init()` runs **`$integrateComponents("wheels.model")`**, which on **every instance** did:

- a `directoryList` of `vendor/wheels/model/`
- a `createObject` **per file** (18)
- a `getMetaData()` **per file** (18)
- iterated all **231** public methods, and called `$willBeOverriddenByMixin` (which did a slow `IsDefined`) **per method**

`Controller.cfc` and `Mapper.cfc` carried the identical pattern.

### The fix

Build the integration plan **once per application** and replay it cheaply:

- `Global.cfc::$componentIntegrationPlan(path)` memoizes the directory scan + per-file `createObject` + `getMetaData` in `application.wheels.integrationPlans` (a sibling of the schema cache — rebuilt on reload, so framework edits are still picked up).
- Public method **references are pre-resolved** into the plan, removing the per-method `.access` check and `instance[name]` scope lookup from the hot path.
- The plugin-mixin **override set is precomputed** (`$mixinOverrideSet`), so the per-method `$willBeOverriddenByMixin` function call is gone from the loop.

**Semantics are unchanged**: the same public methods (and `super<name>` aliases) are mixed into `variables`/`this`, in the same order. Model keeps its `else → super<name>` branch, Controller keeps its no-`else` form, and Mapper keeps the `get`/`controller` exclude-list on its `wheels.Global` integration (via a fallback path). Function references are late-bound to the object they're invoked on, so sharing the cached references across instances is safe.

### Measurements (Lucee 7 + SQLite, this branch)

| | before | after |
|---|---|---|
| 2,000 × `model().new()` | 2772 ms | **1513 ms** (~1.8×) |
| full core suite duration | 33.3 s | **~22–24 s** |
| full core suite result | 4545 pass / 0 fail | **4549 pass / 0 fail / 0 error** |

(The full suite is only partly instance-creation, so its delta understates the win for an instance-heavy suite like the reporter's.)

### Tests

Adds `vendor/wheels/tests/specs/model/integrationPlanCacheSpec.cfc` pinning that the cache is populated and reused, and that materialized instances keep the full, working mixed-in method surface (and stay independent).

### Follow-up (not in this PR)

Profiling showed a larger ceiling: cloning a lean pre-mixed prototype with `Duplicate()` materializes an instance in ~0.13 ms vs ~0.76 ms — ~5.8×. I did **not** take that here because it's a deep change to the hottest path with real cross-engine hazards (e.g. BoxLang's `onDIcomplete` does `variables.this = this`, which `Duplicate` would leave pointing at the prototype), and I could only validate Lucee 7 in this environment. Worth its own PR with the full engine matrix.

Verified on Lucee 7 + SQLite; the cross-engine matrix (Adobe/BoxLang) should run in CI.

https://claude.ai/code/session_0167uSbSN4vZqQqL5QZfdiQm

---
_Generated by [Claude Code](https://claude.ai/code/session_0167uSbSN4vZqQqL5QZfdiQm)_


bpamiri:
Addressed the non-blocking test note in `92e8c87`: dropped the array `.equals()` idiom from `integrationPlanCacheSpec.cfc`.

I went with a slightly different mechanism than the suggested `ArrayAppend(plan, …)`, because assigning the cached array to a local first (`var plan = …integrationPlans["wheels.model"]`) can hit Adobe CF's array-copy-by-value behavior — the append would mutate the copy, leaving the cache untouched and the assertion failing on Adobe. Instead I tag the cached entry through the full application-scope path so the write lands in place:

```cfm
application.wheels.integrationPlans["wheels.model"][1]["cacheReuseSentinel"] = true;
model("author").new();   // must reuse the cached plan, not rebuild it
expect(
    StructKeyExists(application.wheels.integrationPlans["wheels.model"][1], "cacheReuseSentinel")
).toBeTrue();
```

Same property proven (entry reused, not rebuilt), only core struct functions, so it's portable across Lucee/Adobe/BoxLang. Verified green on Lucee 7 (model suite 917 pass / 0 fail).

---

Issue #2962, opened 2026-06-10
roadmap: high-impact application feature gaps — auth scaffold/generator, authorization policy layer, file attachments (Active Storage), channels docs + WebSocket parity


bpamiri:
## Problem

The wave-2 framework review's feature-set assessment (FRAMEWORK-REVIEW-2026-06-09.md §5.2) identifies four **High-impact** application-layer parity gaps. Grouped into one roadmap issue to bound issue count; each section is an independent, separately-shippable feature. The framework already ships the lower-level primitives for several of these — the gap is the developer-facing scaffold/abstraction on top.

### 1. Authentication scaffold/generator (High / Medium effort)
No `wheels generate auth` producing a User model + secure password + registration/login/reset/verification flows on top of the existing `vendor/wheels/auth/` strategies (`Authenticator` + Session/Token/JWT + `JwtService`). The primitives exist; the one-command scaffold does not.
**Prior art:** Rails 8 `generate authentication`; Laravel Fortify.

### 2. Authorization / policy layer (High / Medium effort)
No gates/policies/`can()` abstraction. `AuthMiddleware` answers "who are you" but not "may you do this". Apps must hand-roll authorization in controller filters.
**Prior art:** Laravel Gates + Policies; Pundit.

### 3. File attachments / Active Storage equivalent (High / High effort)
No `hasOneAttached`/`hasManyAttached`, no storage-service abstraction (local/S3/etc.), no variants, no signed URLs. Uploads are raw `cffile`.
**Prior art:** Rails Active Storage.

### 4. Channels documentation + WebSocket parity (High docs / Medium WS, Medium-High effort)
Working SSE channels ship with **zero user-facing docs** (`.ai/wheels/channels/channels.md` is internal only). SSE is server→client only; there is no client→server WebSocket path.
**Prior art:** Rails ActionCable; Phoenix Channels; Laravel Reverb.

## Impact

These are the most-requested parity gaps versus Rails 8 / Laravel 12 for teams adopting Wheels for new apps. Auth scaffold and authorization in particular are table-stakes for typical CRUD apps; the absence forces every app to hand-roll them. Channels has working code with no docs — the lowest-effort, highest-leverage of the four (docs first, WebSocket later).

## Suggested approach

Treat each section as its own design + implementation effort, sequenced by effort/leverage:
1. **Channels docs** first (code exists; write user-facing guide pages under `web/sites/guides/`), then scope client→server WebSocket support.
[...]

PR #3248, opened 2026-06-24
feat(storage): add storage-disk abstraction with local and S3 drivers (#3157)

### LINKED ISSUE #3157: feature: storage-disk abstraction (local + S3 adapters, signed URLs) — attachments phase 1
Child of #2962, sequenced LAST. Phase 1 only: vendor/wheels/storage/ disk abstraction — local + S3 adapters, signed URLs; S3 via plain cfhttp request signing (no JARs). Phase 2 (Rails-style hasOneAttached model macro + polymorphic blob tables) gets its own issue later ONLY if demand shows; storage adapters alone cover most demand at a fraction of the effort.

bpamiri:
Closes #3157. Phase 1 of the attachments line (child of #2962, sequenced last and independently shippable).

## What this adds

A pluggable storage layer under `vendor/wheels/storage/` — local filesystem + S3 behind one uniform disk interface — modeled on Laravel Filesystem / Rails ActiveStorage / AdonisJS Drive. Fully self-contained: **no edits to existing framework files**, purely additive.

- `vendor/wheels/interfaces/StorageDiskInterface.cfc` — 6-method contract: `put / get / exists / delete / url / signedUrl`.
- `vendor/wheels/storage/StorageManager.cfc` — resolves named disks from config and caches them (mirrors `service()` / `model()` resolution).
- `vendor/wheels/storage/drivers/LocalDisk.cfc` — filesystem disk; HMAC-signed local URLs + a `verifySignature()` for a future serving route; rejects `..` traversal; creates dirs via `java.io.File.mkdirs()` (not Lucee-only `DirectoryCreate`).
- `vendor/wheels/storage/drivers/S3Disk.cfc` — S3 over plain `cfhttp`, no AWS SDK / no JARs.
- `vendor/wheels/storage/S3Signer.cfc` — **from-scratch AWS SigV4** (presigned GET URLs + Authorization-header signing).
- `vendor/wheels/tests/specs/storage/StorageSpec.cfc` — 20 specs.
- `changelog.d/3157-storage-disk-abstraction.added.md`.

### Explicitly deferred to follow-ups (per the issue body + parent #2962)
- The `hasOneAttached` model macro / polymorphic blob tables (Phase 2).
- The built-in local-file-serving route + handler (the one genuinely-new infra piece).
- The `storage()` controller/model mixin helper.

So this slice is **drivers only** — no unmade design call blocks it. For now, resolve a disk programmatically (e.g. register `StorageManager` as a singleton in `config/services.cfm`).

## Test status — GREEN on two engines

Ran the storage bundle against the worktree on prebuilt engine images (dir-only Docker mount):

| Engine | Result |
|---|---|
| Lucee 7.0.1 | **20 pass / 0 fail / 0 error** (HTTP 200) |
| Adobe CF 2023 | **20 pass / 0 fail / 0 error** (HTTP 200) |

The SigV4 presign spec pins the **AWS-documented test vector** (`examplebucket/test.txt`, `20130524T000000Z` → signature `aeeed9bb…f604d404`). That value was also reproduced independently from the canonical algorithm in Python before trusting it, so the signer is provably byte-correct against AWS's own reference — on both Lucee and Adobe.

### Bug found and fixed during validation
The original spec used `var url = signer.presignGetUrl(...)` then `expect(url)`. `url` is a **CFML reserved scope** (Anti-Pattern #11): the reference resolved to the URL scope struct, so the matcher compared against `{db=…, format=…, directory=…, method=runRemote}` instead of the return value — 3 specs failed on Lucee. Renamed the locals to `presigned`. (The production code never names anything `url`, so only the test was affected.) This is exactly the silent-ship trap that running the suite catches.

## Cross-engine notes (verified)
- **Invariant 10**: the S3 `cfhttp` wrapper builds a **plain header struct** and attaches each header via `cfhttpparam` in a loop — never `attributeCollection=arguments`. Same pattern as `vendor/wheels/services/packages/HttpClient.cfc`.
- **SigV4 crypto**: SHA-256 hex + chained HMAC-SHA256 key derivation via `javax.crypto.Mac` (the built-in `HMac()` only takes string keys; SigV4 chains binary-keyed rounds). Same primitive family as `wheels.auth.JwtService`.
- **Canonical query** sorts keys with `ArraySort(…, "text")` — does **not** rely on struct insertion-order iteration.
- No closures-as-constructor-args (Invariant 5); no `obj.map()`; no reserved-scope params in production code; driver private helpers are fine (standalone services, not `$integrateComponents`-mixed).

## Sensitive-area review flags (kept as DRAFT)
This touches **DI/config resolution + security-sensitive crypto + outbound HTTP**, so extra scrutiny is warranted on:
- The **SigV4 signer** — request-signing correctness and credential handling. (Mitigated: AWS test-vector pinned + green on Lucee 7 and Adobe 2023.)
- **Path-traversal** rejection in `LocalDisk.$resolve` (`..` blocked).
- **Local signed-URL HMAC** verification (`verifySignature`) — constant-time-ish compare; note the future serving route that consumes it is deferred, so nothing serves local files yet.

### DIFF
--- a/changelog.d/3157-storage-disk-abstraction.added.md
+++ b/changelog.d/3157-storage-disk-abstraction.added.md
@@ -1,0 +1,1 @@
+- Added a pluggable storage-disk abstraction under `wheels.storage` with `LocalDisk` and `S3Disk` drivers behind a uniform `put/get/exists/delete/url/signedUrl` interface, resolved by name through `StorageManager`. S3 access — including presigned, expiring URLs — uses a from-scratch SigV4 signer over plain `cfhttp` (no AWS SDK, no JARs) (#3157).
--- a/vendor/wheels/interfaces/StorageDiskInterface.cfc
+++ b/vendor/wheels/interfaces/StorageDiskInterface.cfc
@@ -1,0 +1,67 @@
+/**
+ * Contract every storage disk driver must satisfy.
+ *
+ * A "disk" is a named, configured storage backend (local filesystem, S3, …).
+ * Drivers expose one small, uniform surface so application code can swap
+ * backends with a config change and no call-site edits — mirroring Laravel's
+ * Filesystem and Rails' ActiveStorage::Service abstractions.
+ *
+ * Implementations live under `wheels.storage.drivers.*`. Resolve a configured
+ * disk through `wheels.storage.StorageManager`.
+ *
+ * [section: Storage]
+ * [category: Interface]
+ */
+interface {
+
+	/**
+	 * Store content at the given key, creating intermediate paths as needed.
+	 *
+	 * @key The opaque storage key (path-like, forward-slash separated).
+	 * @content Binary or string content to write.
+	 * @contentType MIME type hint (used by cloud backends; ignored by local).
+	 * @visibility "public" or "private"; backends that support ACLs honour it.
+	 * @return The stored key.
+	 */
+	public any function put(required string key, required any content, string contentType, string visibility);
+
+	/**
+	 * Read the content stored at the given key as binary.
+	 *
+	 * @key The storage key.
+	 * @return Binary content. Throws Wheels.Storage.NotFound when absent.
+	 */
+	public any function get(required string key);
+
+	/**
+	 * Whether an object exists at the given key.
+	 *
+	 * @key The storage key.
+	 */
+	public boolean function exists(required string key);
+
+	/**
+	 * Delete the object at the given key.
+	 *
+	 * @key The storage key.
+	 * @return true when an object was deleted, false when nothing was there.
+	 */
+	public boolean function delete(required string key);
+
+	/**
+	 * A non-expiring URL for the object (public objects / served route).
+	 *
+	 * @key The storage key.
+	 */
+	public string function url(required string key);
+
+	/**
+	 * A signed, time-limited URL granting temporary access to a private object.
+	 *
+	 * @key The storage key.
+	 * @expiresIn Seconds until the URL expires (default 300).
+	 * @contentDisposition Optional Content-Disposition the download should carry.
+	 */
+	public string function signedUrl(required string key, numeric expiresIn, string contentDisposition);
+
+}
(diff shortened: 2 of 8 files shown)


bpamiri:
## Adversarial review follow-ups (workflow `triage-and-fix-open-issues`)

This PR was implemented and reviewed by an automated triage+fix workflow. The review verdict was **looks-good / cross-engine clean / fixes the issue**, with the SigV4 signature math independently re-derived in Python and matched byte-for-byte against AWS's reference vector. The items below are **non-blocking nits** captured so they aren't lost before this graduates from draft.

### Code nits
- [ ] **SigV4 canonical-query sort is case-sensitive.** `S3Signer.$buildCanonicalQuery` sorts with `ArraySort(keys, "text")`. SigV4 mandates byte/ASCII order of the *percent-encoded* keys. This is correct *only* for the current fixed key set (`X-Amz-*`, optional `response-content-disposition`) — harden to true byte-order (or document the invariant) before the API ever accepts caller-supplied / mixed-case query params.
- [ ] **`LocalDisk.verifySignature()` comment overstates the guarantee.** It labels the compare "Constant-time-ish", but `CompareNoCase(expected, signature) == 0` short-circuits on the first differing char and is **not** constant-time. Drop the wording, or implement a real fixed-time compare, before the local-serve route lands.
- [ ] **`S3Signer.$amzNow()` is the one untested production path.** Every spec pins `amzDate` for determinism, so the live `DateConvert("local2utc", Now())` + format path is never exercised. Add a test (or inject/pin the clock) so timestamp generation + UTC conversion is covered.

### Design decisions to confirm before merge
- [ ] This is **Active Storage Phase-1 under roadmap #2962** — decide whether to accept a bot-authored implementation of a roadmap feature, or treat this PR as a design proposal.
- [ ] The slice is **drivers-only**: no `storage()` controller/model mixin, no local-file serving route, no DI auto-registration — callers must register `StorageManager` themselves (e.g. in `config/services.cfm`). Confirm that matches the intended Phase-1 boundary.
- [ ] Confirm the **6-method disk interface shape** (`put/get/exists/delete/url/signedUrl`) is the contract you want to commit to publicly before drivers depend on it.

Issue #2960, opened 2026-06-10
CSRF default-on decision (5.0): flip the default after the detect-and-warn deprecation window


bpamiri:
## Problem

CSRF protection is opt-in and silently disabled by the documented `config()` override footgun. `vendor/wheels/controller/csrf.cfc:18-20, 37-51` no-ops unless `StructKeyExists(variables.$class, "csrf")`; CSRF is wired only via the generated base controller (`cli/lucli/templates/app/app/controllers/Controller.cfc:23-25`), which itself doesn't call `super.config()`. Any child controller overriding `config()` without `super.config()` silently disables CSRF for that controller, with no error or startup warning. Verified against `origin/develop` @ `8971094`.

This is a recurring footgun (tracked around #2845; see also project memory on the `super.config()` chain fragility). Prior remediation chose docs + generator templates that call `super.config()`.

## Impact

A developer who overrides `config()` in a controller — a common, documented pattern — without calling `super.config()` silently ships that controller with CSRF disabled and no warning. The protection's presence is invisible and easy to drop.

## Suggested approach

This is a deliberate behavior change for every existing app, so it needs an upgrade-path design rather than an autonomous flip:
- Either enforce CSRF at the dispatch/middleware layer (independent of per-controller `config()`), or
- Detect-and-warn at startup when a controller overrides `config()` without calling `super.config()`.
- If making CSRF default-on, design a deprecation window / upgrade note so existing apps are not broken silently.

## Acceptance criteria

- [ ] A controller overriding `config()` without `super.config()` either keeps CSRF active or emits a clear startup/dev warning.
- [ ] If the default flips on, an upgrade path and `wheels upgrade check` advisory exist.
- [ ] Specs cover the override-without-super.config() case.

## Source

Internal multi-agent framework review 2026-06-09, wave 2 (issues phase). Findings: security-pass SEC-10 (S5), followups SEC-10.

bpamiri:
**Maintainer call (2026-09-08):** Remains parked for **5.0**. No work this week (including no 4.x doctor/startup warn). Revisit when 5.0 packaging/shape is clearer.

Issue #3063, opened 2026-06-12
Documented config/services.cfm snippet (`var di = injector();`) crashes the entire app at boot on Adobe CF


bpamiri:
## Summary

The DI registration snippet shown in the dependency-injection guide and CLAUDE.md:

```cfm title="config/services.cfm"
var di = injector();
di.map("emailService").to("app.lib.EmailService").asSingleton();
```

is a **compile-time error on Adobe CF** (top-level `var` in an included template). Because the file is included during `onApplicationStart`, the whole application fails to start: every request returns HTTP 500, and the real error is masked by a secondary onError failure (`Element WHEELS.ENGINEADAPTER is undefined ...`), making this brutal to debug. Lucee accepts the same file — so the documented pattern works on Lucee and takes down Adobe.

## Repro (wheels-test-adobe2023:v1.0.1, develop @ f668c5010)

```bash
cat > config/services.cfm <<'EOF'
<cfscript>
var di = injector();
di.map("probeService").to("app.lib.ProbeService").asSingleton();
</cfscript>
EOF
# boot the adobe2023 harness, then:
curl -s -o /dev/null -w '%{http_code}\n' 'http://localhost:61753/'   # -> 500 (every request)
```

- Instrumenting the file with `writeLog` lines shows the FIRST line never executes → compile-time failure, not runtime.
- Removing only the `var` keyword (`di = injector();`) boots clean; all registrations work (`di.map(...)` on an Injector instance is fine on Adobe — the struct `.map()` collision does not apply here).
- The same file with `var` boots fine on lucee7.

## Root cause

Adobe CF rejects `var` declarations at the top level of an included template; `config/services.cfm` is included via `$includeConfig()` during `onApplicationStart`, so the parse failure aborts application start, and the onError fallback then fails secondarily, masking the original message.

## Proposed direction

1. **Docs:** change the snippets to `local.di = injector();` (the file is included inside `$includeConfig()`, a function context, so `local.` is valid on both engines) or unscoped assignment. Affected: `web/sites/guides/src/content/docs/v4-0-0/core-concepts/dependency-injection.mdx` (~line 31), CLAUDE.md "DI Container Quick Reference", any other services.cfm examples using top-level `var`.
2. **Framework hardening (recommended):** wrap the `$includeConfig()` include so a template compile/parse failure is logged with the offending file name ("config/services.cfm failed to compile: \<msg\>") instead of cascading into the masked onError 500.
[...]

PR #3291, opened 2026-07-06
feat(cli): wheels generate auth — session/token/jwt scaffold on the auth primitives

### LINKED ISSUE #3155: feature: wheels generate auth — session auth scaffold on the existing wheels.auth primitives

bpamiri:
Closes #3155. Part of #2962.

Implements `wheels generate auth` — a one-command authentication scaffold built on the existing `wheels.auth` primitives (`PasswordHasher`, `Authenticator`, `SessionStrategy`/`TokenStrategy`/`JwtStrategy`), following the Phoenix/Rails-8 shape recommended by the wheels-bot cross-framework research on the issue.

> **Stacked PR:** this branch is based on #3288 (`peter/issue-3155-password-hasher`), which is still open. The `PasswordHasher` commit will fall out of this diff automatically once #3288 merges. Merge #3288 first.

## Approved design decisions

The maintainer signed off on 2026-07-06 on the design calls flagged in the issue and the bot's research comment:

- **Session flow as the default strategy** (`--strategy=session`).
- **`--strategy=token` and `--strategy=jwt` API variants included in v1** — both emit `app/controllers/api/Sessions.cfc` (no views).
- **Registration ON by default**, opt out with `--no-registration` (session strategy only; noted as not-applicable for token/jwt).
- **CFML session-scope storage** via `SessionStrategy` — no DB sessions table.
- **Code-you-own upgrade policy**: stamped header on every generated file; upgrade by re-running with `--force` and reviewing `git diff`. Marker-delimited config blocks are replaced in place, never duplicated. Migrations are never overwritten.
- **PBKDF2 hashing via the `PasswordHasher` service** (from #3288) — resolves the cross-engine bcrypt gap the bot's fix-hold flagged.

## What it generates

**Session (default):**
- `app/models/User.cfc` — transient `password` property validated (presence on create, unique+format email, confirmation, 12-char minimum), hashed into `passwordDigest` in `beforeSave` and scrubbed; `authenticate()` with transparent rehash-on-login; single-use SHA-256-digested reset tokens (2h expiry, burned on use); `protectedProperties()` guards mass assignment.
- `app/controllers/Sessions.cfc`, `Passwords.cfc`, `Registrations.cfc` — every `config()` calls `super.config()` first (the #2960 CSRF footgun), all-named `verifies(...)`, injection-safe query-builder finders.
- `startFormTag`-based views (CSRF token automatic), `cfparam`'d, plain semantic HTML.
- `app/migrator/migrations/<ts>_create_users_table.cfc` — `t.string(columnNames=...)` style, `timestamps()`, unique email index.
- Marked blocks (`// wheels:generate-auth:*`) injected into `config/routes.cfm` (before root/wildcard), `config/services.cfm` (created if absent — `passwordHasher`/`authenticator`/`sessionStrategy` singletons), and `app/events/onapplicationstart.cfm` (strategy registration, per the documented wiring).
- App specs: `tests/specs/models/UserAuthSpec.cfc`, `tests/specs/controllers/SessionsControllerSpec.cfc`.

**Token:** same model + migration plus `apiTokenDigest`; `api/Sessions.cfc` mints an opaque token (plaintext returned once, only the SHA-256 digest stored), `DELETE` revokes; `TokenStrategy` wired with a **hoisted** validator closure (Cross-Engine Invariant 5).

**JWT:** `api/Sessions.cfc` mints JWTs via `JwtService`; `WHEELS_JWT_SECRET` read from the environment and **startup fails loudly** when missing/shorter than 32 bytes; no-server-side-revocation documented in the generated header.

## Tests

- New `cli/lucli/tests/specs/services/GenerateAuthSpec.cfc` (31 specs): dispatch reaches `generateAuth` via the `generate()` switch, per-strategy file sets, `--no-registration` omissions, force-overwrites/refuses-without-force, marker idempotency across re-runs, comment-stripped `super.config()`-first scans of every emitted controller, and a no-inline-closure-as-constructor-arg guard.
- Full CLI suite against this branch (worktree server, `/wheels/cli/tests?format=json`): **1122 pass, 0 fail, 0 error**.
- Hand-verified end-to-end by running the generator (all three strategies, plus `--no-registration`) into temp project dirs through the dev server and eyeballing the emitted files and injected blocks (routes land before `.wildcard()`/`.root()`, services/bootstrap blocks correctly wrapped). A full `wheels new` boot test was **not** run.

## Docs

- `wheels generate auth` section + type-table row in the CLI code-generation reference.
- "Scaffold it in one command" lead-in (with flags table) at the top of `digging-deeper/authentication-patterns.mdx`.
- Changelog fragment `changelog.d/3155-generate-auth.added.md`.


bpamiri:
## Shepherd round 1 — review hardening pushed (2313810)

Addressed the wheels-bot review plus an internal adversarial pass. CLI suite locally: **1134 pass / 0 fail / 0 error** (GenerateAuthSpec 31 → 43 specs).

**Bot Correctness item (token revoke Authorization reachability) — verified, it was real.** `Global.cfc::$cgiScope()` builds `request.cgi` from a fixed key allowlist that does not include `http_authorization`, and Wheels never sets `request.headers`, so `authenticate(request)` could never see the bearer token and `DELETE /api/session` always 401'd. Fixed template-side: the token controller now reads `GetHttpRequestData(false).headers` and hands the strategy an explicit `{cgi: {http_authorization: ...}}` context (shape verified against `TokenStrategy.$extractFromHeader`). The same gap in the pre-existing guide filter pattern is a framework/docs follow-up outside this PR's scope.

**Bot Conventions item (`.end()` routes fallback) — taken.** The last-`.end()` fallback could park routes after `.wildcard()` where they never match; the generator now skips with a manual-insert note instead, and the three routes-anchor edge paths are covered by new specs.

**Internal review fixes in the same push:**
- **Blocker (cross-engine, #3063 class):** all three bootstrap templates emitted template-level `var` into `app/events/onapplicationstart.cfm` — an Adobe CF compile error that 500'd every request of every generated app on Adobe 2018–2025. Bootstrap variables are now `local.`-scoped, with regression specs per strategy; the guide's four init-hook snippets taught the same pattern and were fixed with a caution aside.
- **High:** blank-password reset burned the token, reported success, and kept the old password valid. `Passwords##update` now rejects blanks with a field error before clearing the token (spec pins the guard ordering).
- Login timing oracle: dummy PBKDF2 derivation on unknown emails in all three login controllers; headers + CLI next-steps now recommend `wheels.middleware.RateLimiter` on credential endpoints.
- Generated Sessions spec called `processAction("create", params)` — the only parameter is `includeFilters`, so it silently disabled before-filters; now `processAction()`.
- Scaffold rollback now runs on any failure (not just typed ScaffoldErrors), so an IO error mid-run can't leave a half-generated scaffold.
- Passwords header documents that reset does not invalidate live sessions; next-steps call out wiring reset-email delivery (the `create()` TODO).

bpamiri:
## Round 2 — runtime verification done (the bot's residual ask), two real defects found and fixed

Commit `2be96af05`. Per the reviewer's residual item ("boot the token/jwt API paths once against a live app"), I scaffolded three fresh apps with the worktree CLI (`wheels new` → `wheels generate auth` → `server run`, Lucee 7 + SQLite) and drove every strategy end-to-end. The exercise caught **two runtime defects** that the string-level CLI specs structurally cannot see — exactly why the reviewer asked for it.

### Defect 1 — every new-account creation failed validation (all three strategies)

`passwordDigest` is an `allowNull=false` column, so Wheels auto-registers `validatesPresenceOf(passwordDigest)` (`Model.cfc:224-257`). The generated model populates the digest in the `beforeSave` callback — which runs **after** validation — so every `save()` of a new record failed with `Password Digest can't be empty`. Registration (`POST /register`), programmatic creation, `wheels seed`, and the generated `UserAuthSpec` round-trip spec were all broken out of the box. This is the exact trap the authentication-patterns guide documents (its stopgap hashes in `beforeValidation`; the generator hashes in `beforeSave` so plaintext length/confirmation validate first — hence the different fix).

**Fix:** `property(name="passwordDigest", automaticValidations=false)` in the model template — the documented per-property opt-out. Presence stays guaranteed transitively: `validatesPresenceOf(password)` on create feeds the hashing callback, and the DB NOT NULL constraint is the backstop. Validate-plaintext-then-hash ordering is preserved.

### Defect 2 — token/jwt controllers 500'd on every request

`app/controllers/api/Sessions.cfc` extended the bare `"Controller"`, which cannot resolve from a subfolder: every request failed with `invalid component definition, can't find component [Controller]`. **Fix:** `extends="app.controllers.Controller"`, matching the admin generator's namespaced-controller convention (pinned by `AdminGeneratorTest`).

### Verified live (after fixes, from regenerated templates — no hand patching)

| Strategy | Checks |
|---|---|
| **token** | `POST /api/session` valid → 201 + opaque token; wrong password → 401; unknown email → 401 (timing-dummy path, comparable latency); `DELETE /api/session` with `Authorization: Bearer` → `{revoked: true}` (**confirms the Authorization-header handoff works at runtime**); same token again → 401 `Invalid or expired token`; no header → 401 |
| **jwt** | valid creds → 201 + 3-segment HS256 JWT with `sub`/`email`/`iat`/`exp` claims (`exp − iat = 3600`); wrong password → 401; `DELETE` → stateless-revocation doc message |
| **session** | `GET /register` → CSRF-tokened form; `POST /register` → 303 + `Welcome!` flash (account created + logged in); wrong-password login → re-render with uniform error; correct login → 303 + `Welcome back.` flash |

Also verified along the way: `wheels migrate latest` applies the generated migration cleanly, the routes block lands at the `// CLI-Appends-Here` anchor before `.wildcard()`, and the bootstrap registers the strategy on first boot.

### Tests

Three new specs pin the fixes (digest auto-validation opt-out asserted for all three strategy fixtures; `app.controllers.Controller` extends asserted for token and jwt). Full CLI suite against the worktree: **1137 pass / 0 fail / 0 error** (+3 vs. the previous round's 1134).

PR #3327, opened 2026-07-23
fix(events): detect RustCFML under reportAsLucee impersonation (v0.507+)

bpamiri:
## Summary

RustCFML v0.507.0 turned on `reportAsLucee` by default: `server.coldfusion.productName` now reports `"Lucee"` (with Lucee's classic `productVersion`), and the real engine version moves to `server.lucee.version` behind a Lucee-major prefix (`7.0.519.0` = RustCFML 0.519.0). The upstream commit explicitly names Wheels' engine gate as the motivation and documents `server.lucee.versionName == "RustCFML"` as the stable identity marker (their own `isRustCFML()` BIF keys on it).

`$detectEngine()` identified RustCFML solely by `productName == "RustCFML"`, so on v0.507+ it fell through to the Lucee branch and selected `LuceeAdapter` — making `RustCFMLAdapter` dead code. Concrete misroutes observed:

- `isRustCFML()` → false: `EngineCapabilitiesSpec` asserted JVM class-loading on a JVM-less engine (the ladder failure that surfaced this)
- `supportsImageInfo()` → true (Base default): `$imageTag()` proceeds into `cfimage action="info"` paths that error on RustCFML
- `databaseAdapters/Base.cfc` `serverName == "Lucee"` branch → the Lucee-only `psq=false` cfquery attribute applied to every query
- `$checkMinimumVersion` satisfied by the impersonated version string instead of the accept-any-version RustCFML branch

## Fix

Add a detection branch matching `server.lucee.versionName == "RustCFML"` ordered before the Lucee branch, deriving `serverVersion` by dropping the impersonated major segment (`ListRest`, so a future `8.`-prefix keeps working). Pre-v0.507 scopes still hit the original `productName` branch first; real Lucee carries a different `versionName` (e.g. `Gelert`) and falls through unchanged.

## Tests

- New detection-ladder spec with the exact v0.519.0 impersonating scope shape; existing Lucee fixture extended with a realistic `versionName` to pin the negative path.
- Full core suite green locally on Lucee 7 + SQLite (4694 passed).
- Live-verified against the actual RustCFML v0.519.0 binary: detection selects RustCFML, the previously-failing `EngineCapabilitiesSpec` guard fires and passes, and a bad-spec set diff vs the unfixed run shows zero collateral regressions. (It also exposed that `migrationSpec`'s multi-column-index "fix" on the ladder was an artifact of the `psq=false` misroute — it reverts to its known-failing baseline state with correct detection.)

Upstream context: RustCFML/RustCFML#283, RustCFML/RustCFML#284, RustCFML/RustCFML#285 filed today for the three genuine v0.482–v0.511 engine regressions found in the same investigation.


alexskinner:
server.lucee.versionname tells you its RustCFML

The purpose of the change was to make sure the various ways that frameworks check for Lucee could find it in order to go down the lucee compatibility path but if they looks at the actual lucee versionname they could see it was RustCFML

Issue #3302, opened 2026-07-07
compat-matrix: engine-job failures are invisible (continue-on-error) — harden after burning down pre-existing leg debt


bpamiri:
## Problem

The compat-matrix engine jobs run with `continue-on-error`, so the workflow run concludes **"success" even when every engine job fails**. Two consequences observed on 2026-07-06:

1. A guard spec (`ConfigRoutesStaleDocUrlSpec`, from #3281) failed **every engine × database leg uniformly** for a full day before anyone noticed — the run banner was green (fixed in #3293).
2. `UpdateAllIncludeJoinSpec` has been red on `lucee7+mysql` **since at least 2026-06-28** (a real data-affecting bug, now #3294) — invisible for the same reason. The June-28 baseline additionally carries per-engine debt (Adobe: `typedColumnDefaultsSpec`, `contentSpec`, `RouteTesterHardeningSpec`, transaction specs; Lucee 6: `RouteTesterHardeningSpec`; BoxLang: `miscellaneousSpec`, `AdapterIdentityTemplateSpec`, `testEvaluateExpressionSpec`, `DatabaseAdapterSpec`) that nobody triaged because nothing ever turned red at the run level.

Same lesson as the box-install check (#3281): a check that can't fail visibly is noise, and real regressions hide behind it.

## Proposal

- Keep per-DB soft-fail for the known-flaky infrastructure (`SOFT_FAIL_DBS=oracle` already exists) but make a **hard-DB leg failure fail the engine job**, and an engine-job failure fail the run.
- To get there without a permanently red banner, first burn down (or explicitly soft-fail with a linked issue) the pre-existing per-engine failures listed above — each should be a tracked issue, not silent debt.
- Optional: have the Test Matrix Summary job post a compact per-engine × per-DB verdict table into the run summary so a glance shows exactly which leg regressed.

## Evidence

- Run [28809821005](https://github.com/wheels-dev/wheels/actions/runs/28809821005) (2026-07-06 dispatch): conclusion `success`, all 5 engine jobs `failure`.
- Run [28815315339](https://github.com/wheels-dev/wheels/actions/runs/28815315339) (post-#3293): same shape; per-leg artifacts show the real state (lucee7 clean except mysql #3294).

bpamiri:
## A third failure mode: `continue-on-error` also *misattributes* the debt

Adding this here rather than as a separate issue, per maintainer call.

[...]

Issue #3294, opened 2026-07-06
updateAll with multiple includes silently updates 0 rows on MySQL


bpamiri:
## Symptom

`UpdateAllIncludeJoinSpec` → "updateAll with multiple includes updates matching rows" fails on **MySQL only**: `Expected [1] but received [0]`. Every other engine × database leg passes it (SQLite, Postgres, SQL Server, H2, CockroachDB — verified in matrix run [28809821005](https://github.com/wheels-dev/wheels/actions/runs/28809821005), lucee7 leg artifacts).

In plain terms: `updateAll()` with an `include=` spanning multiple associations appears to update **zero rows on MySQL** where it updates the matching row everywhere else — a silent bulk-update no-op on a mainstream database.

## Dating

- The spec landed in #3011 (`efffbc360`, 2026-06-11, the per-WHERE Migration + ON-split fixes).
- The failure is present in the 2026-06-28 scheduled matrix (lucee7+mysql: 4619 pass / this 1 fail), 2026-07-05, and 2026-07-06 runs — red on every weekly run since the spec existed.
- `efffbc360` is contained in v4.0.4 and v4.0.5 — **not a new regression**; shipped releases already behave this way.

## Why nobody saw it

Two blind spots stacked: the compat-matrix engine jobs are `continue-on-error` (the run banner says "success" while legs fail), and required PR CI runs Lucee 7 + SQLite only. Worth fixing the visibility problem alongside the bug — a matrix leg failing a non-soft-fail database should probably fail the job.

## Where to look

MySQL's multi-table `UPDATE ... JOIN` syntax differs from the ANSI/other-adapter forms (no `FROM` clause; joins live in the UPDATE clause itself; alias handling differs). The multi-include path from #3011's ON-split work is the likely divergence point — the single-include case passes on MySQL, so it's specifically the second join that breaks row matching.

## Repro

```
tools/test-matrix.sh lucee7 mysql
# or directly: /wheels/core/tests?db=mysql&format=json&testBundles=wheels.tests.specs.model.UpdateAllIncludeJoinSpec
```

bpamiri:
## Correcting the diagnosis: `updateAll` was never broken

Resolved by #3342, but the fix is in the **spec**, not the framework — so the description above should not be cited as a known data bug. Three of its claims do not hold.

[...]

PR #3338, opened 2026-08-02
fix(model): namespace per-request query cache under request.wheels.$queryCache

### LINKED ISSUE #3336: Per-request finder cache key collides with request.wheels.tenant for a model named Tenant (silently drops tenant datasource routing)
## Summary

The per-request finder cache keys itself on the **bare model name** in the `request.wheels` struct:

```cfm
// vendor/wheels/model/read.cfc:310-311 (4.0.5)
if (!StructKeyExists(request.wheels, variables.wheels.class.modelName)) {
    request.wheels[variables.wheels.class.modelName] = {};
}
```

The multi-tenancy subsystem stores the resolved tenant in the **same struct** under the key `tenant`:

```cfm
// vendor/wheels/middleware/TenantResolver.cfc
request.wheels["tenant"] = local.tenant;   // {id, dataSource, config, $locked}
```

CFML struct keys are case-insensitive, so an application with a model named `Tenant` — the natural name for the control-plane model in a database-per-tenant app, and the name used in the framework's own docs (`public/docs/reference/model/sharedmodel.txt` shows `app/models/Tenant.cfc` calling `sharedModel()`) — has its **query cache and its resolved-tenant context aliased onto one key**.

This affects any Wheels 4.0 app that combines a `Tenant` model with `wheels.middleware.TenantResolver`, i.e. the documented way to build multi-tenancy on 4.0.

Confirmed in **4.0.5** (`read.cfc:311, 322, 324, 356`) and still present at current dev HEAD (same block, `read.cfc:315, 326, 328, 360`).

### 1. Write path — resolved tenant is erased mid-request (more severe)

```cfm
// vendor/wheels/model/miscellaneous.cfc:5-7
public void function $clearRequestCache() {
    request.wheels[variables.wheels.class.modelName] = {};
}
```

`$clearRequestCache()` is called after every create / update / delete / bulk operation (`create.cfc:313`, `update.cfc:295,326`, `delete.cfc:199,222`, `bulk.cfc:63,153`). On a `Tenant` model instance it sets `request.wheels.tenant = {}` — **wiping the resolved tenant context for the remainder of the request.**

Because `databaseAdapters/Base.cfc` only swaps in the tenant datasource when `request.wheels.tenant.dataSource` exists:

```cfm
// databaseAdapters/Base.cfc:807-826
if (!variables.$sharedModel
    && arguments.dataSource == variables.dataSource
    && IsDefined("request.wheels.tenant.dataSource")
    && Len(request.wheels.tenant.dataSource)) {
    arguments.dataSource = request.wheels.tenant.dataSource;
}
```

…every tenant-scoped query issued after that point **silently falls back to the application default datasource** — the control-plane database — with no error. Any tenant-scoped write in the same request lands in the wrong database.

Realistic trigger: a settings or onboarding screen served on a tenant's subdomain that updates a field on the tenant's own record (`tenantRecord.save()`), then writes anything tenant-scoped (an audit row, a related record) later in the same request.

### 2. Read path — an unresolved request appears resolved

If a `Tenant` finder runs **before** `TenantResolver` sets the context — which happens when the resolver closure itself looks tenants up via `model("Tenant").findAll()`, the obvious implementation of a subdomain→tenant directory — then `read.cfc:311` creates `request.wheels.tenant = {}` and populates it with a query-cache entry keyed by a query hash.

`TenantResolver` skips assignment when the resolver returns no match, and its cleanup runs only in a `finally` **after** `next()`, so that junk struct survives for the whole downstream request. Application code that treats "tenant resolved" as "`tenant()` is non-empty" then misbehaves: in our case `requiresApex()`-style guards began rejecting legitimate apex-host requests (login, password reset, invitation redemption) with 403s, intermittently — once per directory-cache rebuild window.

The struct is truthy-but-malformed: non-empty, but with no `id`, `dataSource`, or `config`. `tenant()` returns it as though it were a real resolved tenant.

## Reproduction (failure mode 1, the data-correctness one)

1. App with `app/models/Tenant.cfc` (`sharedModel()`) and `set(middleware=[new wheels.middleware.TenantResolver(resolver=fn, strategy="subdomain")])`.
2. A tenant-scoped model, e.g. `app/models/AuditLog.cfc` (no `sharedModel()`), whose table exists in both the control-plane and tenant databases.
3. Request a route on a resolved tenant subdomain whose action does:

```cfm
var t = model("Tenant").findByKey(tenant().id);
t.name = "Renamed";
t.save();                              // $clearRequestCache() → request.wheels.tenant = {}
model("AuditLog").create(action="X");  // lands in the CONTROL-PLANE database, not the tenant DB
```

Expected: the `audit_logs` row is written to the tenant database.
Actual: `$tenantDataSource()` no longer sees a tenant, so it resolves to `application.wheels.dataSourceName` and the row is written to the control-plane database. No exception.

`cacheQueriesDuringRequest` defaults to `true` (`events/init/caching.cfm`), so no opt-in is required to reach this.

## Suggested fix

Namespace the per-request query cache so it can never collide with framework-owned `request.wheels` keys:

- `model/read.cfc` — replace `request.wheels[modelName]` with `request.wheels.$queryCache[modelName]` (creating `$queryCache` if absent) at all four sites.
- `model/miscellaneous.cfc::$clearRequestCache()` — clear `request.wheels.$queryCache[modelName]` instead.

That confines the blast radius to one reserved key and leaves `request.wheels.tenant` owned solely by the tenancy subsystem. A `$`-prefixed sub-struct also matches the convention already used elsewhere in the request scope.

Defensive hardening worth considering alongside it:

- Have `TenantResolver` **delete** any pre-existing `request.wheels.tenant` when its resolver returns no match, so a stale/foreign value can't outlive resolution.
- Consider whether `tenant()` should validate shape (return `{}` unless `dataSource` is present) rather than returning whatever occupies the key — this would have downgraded failure mode 2 from a wrong-behavior bug to a no-op.

## Impact assessment

Failure mode 1 is a silent cross-database write under a supported configuration, which makes this more than a naming nuisance. The application-level workarounds available today — avoiding the `Tenant` model name, keeping model finders out of the resolver path, and treating only a present `dataSource` as a resolved tenant — all require app authors to know about the aliasing, and none of them prevent failure mode 1 for an app that legitimately updates a tenant record from within that tenant's request.

Found while building a database-per-tenant application on 4.0.5 (Lucee 7 / YugabyteDB). Happy to open a PR for the `$queryCache` namespacing plus a regression spec if the approach looks right.

/cc @claude

bpamiri:
Fixes #3336.

## The collision

The per-request finder cache keyed itself on the **bare model name** directly in `request.wheels`. CFML struct keys are case-insensitive, so a model named `Tenant` — the natural name for the control-plane model in a database-per-tenant app, and the one used in `TenantResolver`'s own docblock — shared a single key between its query cache and the framework-owned `request.wheels.tenant`.

The fix moves the cache under a reserved `request.wheels.$queryCache` sub-struct. Caching behaviour is unchanged: same per-model slots, same query keys, same lifecycle.

```cfm
// before — aliases onto request.wheels.tenant for a model named Tenant
request.wheels[variables.wheels.class.modelName][local.queryKey] = local.findAll;

// after
request.wheels["$queryCache"][variables.wheels.class.modelName][local.queryKey] = local.findAll;
```

`model/miscellaneous.cfc` gains `$ensureRequestQueryCache()`, which owns namespace creation for the four call sites (three in `read.cfc`, one in `$clearRequestCache()`).

## Scope

The sweep found three independent families of dynamic keys in `request.wheels`. Only the first is fixed here:

| Site | Key | Auto-collides with `tenant`? |
|---|---|---|
| `model/read.cfc` ×3, `model/miscellaneous.cfc` ×1 | bare model name | **Yes** — this issue |
| `Global.cfc` `pagination()` / `setPagination()` | user-supplied handle (default `"query"`) | Only if an app names a handle `tenant` — filed as #3339 |
| `model/callbacks.cfc:331` | query hash | No |

The pagination handles are a real-but-latent collision surface; they need an app author to actively pick the name, so they're tracked separately in #3339 rather than bundled in.

## Tenant-context hardening (second commit)

Folded in at maintainer request. The first commit stops the framework from ever writing a query cache onto `request.wheels.tenant`; these two stop anything *else* that lands there from being read as a resolved tenant.

- **`tenant()` requires a non-empty `dataSource`** before treating the key as an active tenant — the same test `$tenantDataSource()` already applies before routing a query. It previously returned whatever occupied the key, so a malformed value satisfied every `IsDefined("request.wheels.tenant")` or truthiness guard while carrying no usable routing information. The two now agree by construction and can't disagree about whether a request is tenanted.
- **`TenantResolver` deletes any pre-existing value on the key when its resolver returns no match.** Previously an unresolved request only looked unresolved *after* the `finally` block, so a stale or foreign value could read as resolved for the whole downstream request.

All four framework producers — `switchTenant()` (throws without a `dataSource`), `TenantResolver`, `Job.$restoreTenantContext()` and `TenantMigrator` (both guard) — already guarantee a non-empty `dataSource`, so correctly-resolved tenants are unaffected. Only foreign values are filtered.

Net effect: a malformed tenant context degrades from *wrong behaviour* to a *no-op*.

## Tests

`vendor/wheels/tests/specs/model/requestQueryCacheTenantCollisionSpec.cfc` — 6 specs covering both failure modes plus the namespacing invariant. Backed by a new `Tenant` fixture model mapped onto the existing `c_o_r_e_authors` table, so `populate.cfm` is untouched.

Written red first. The failures before the fix were exactly the two reported behaviours:

- `key [ID] doesn't exist` reading `request.wheels.tenant.id` after a `Tenant` write → **failure mode 1**, tenant context wiped to `{}`
- `IsDefined("request.wheels.tenant")` returning true after a `Tenant` finder → **failure mode 2**, tenant context fabricated

`requestQueryCacheSpec.cfc`'s 8 existing assertions are repointed at the namespaced key — they're white-box assertions on the exact structure being moved.

## Verification

Lucee 7 + SQLite via `bash tools/test-local.sh` (full suite, not scoped):

| | Result |
|---|---|
| `develop` @ `fc59492` | **4707 passed**, 0 failed |
| after commit 1 (namespacing) | **4713 passed**, 0 failed |
| after commit 2 (hardening) | **4717 passed**, 0 failed |

Delta is exactly the 10 new specs (6 + 4).

Both sets were written red first. Reverting just the two hardening files while keeping their specs yields exactly 4 failures, confirming those specs exercise the change rather than restating existing behaviour.

⚠️ **The cross-engine matrix did not run.** `tools/test-matrix.sh lucee7 mysql` fails on this machine before any test executes — Docker Desktop can't bind-mount `tools/docker/lucee7/box.json` onto `/wheels-test-suite/box.json` under virtiofs:

```
error mounting ".../tools/docker/lucee7/box.json" to rootfs at "/wheels-test-suite/box.json":
mountpoint "/run/host_virtiofs/.../box.json" is outside of rootfs
```

It fails identically on clean `develop`, so it's environmental, not this change. Because I couldn't verify Adobe/BoxLang locally, the implementation was deliberately rewritten to avoid anything engine-sensitive: an earlier draft returned the cache struct by reference and used `StructClear()`, which would have depended on struct-reference write semantics holding on Adobe (cf. cross-engine invariant #6, where Adobe copies arrays by value in struct literals). What's here instead is a literal key-path swap of the original code plus an ensure-helper — no new semantics on any path.

### What PR CI does and does not cover

To be precise about the engine coverage this PR actually carries:

| Engine | Coverage on this PR |
|---|---|
| Lucee 7 | **Full core suite** — `Lucee 7 + SQLite (LuCLI)`, plus my local run |
| Adobe 2023 | **Boot + HTTP probes only** — `smoke-env.yml` cold-boots the demo app under `production`/`testing` and asserts clean 404s / no stack traces / reload behaviour. Per its own comments the probes never query the database, so no model spec runs. It does prove the new `$`-prefixed public mixin compiles and integrates via `$integrateComponents()` on Adobe. |
| Adobe 2025, Lucee 6, BoxLang | **Not exercised by PR CI** |

`compat-matrix.yml` — the workflow that runs the full suite across all five engines — is **weekly cron + `workflow_dispatch` only, and `continue-on-error: true`**. It does not trigger on pull requests. I've manually dispatched it against this branch: https://github.com/wheels-dev/wheels/actions/runs/30772173443

Constructs used are all matrix-safe by inspection: `StructKeyExists`, bracket-notation struct keys, no closures, no `attributeCollection`, no reserved scope names, no `finally` loops. The new helper is `public` + `$`-prefixed per cross-engine invariant #7 so it mixes in correctly, and lives in `wheels.model.*`, which is outside the `protectedControllerMethods` surface — so it can't shadow a controller action name.

## Docs

`web/content/blog/posts/caching-in-wheels-4.md` documents the old `request.wheels[ModelName]` path in two places; both are updated. Drop that hunk if you'd rather not touch a published post.


bpamiri:
## Cross-engine verification (manual `workflow_dispatch`)

`compat-matrix.yml` doesn't trigger on PRs, so I dispatched it against both this branch and clean `develop` and diffed the JUnit artifacts.

- Branch: https://github.com/wheels-dev/wheels/actions/runs/30772173443
- `develop` baseline: https://github.com/wheels-dev/wheels/actions/runs/30773074179

### This PR's specs

The 6 new specs ran **168 times across 28 engine×DB legs** (Lucee 6, Lucee 7, Adobe 2023, Adobe 2025, BoxLang × sqlite / h2 / mysql / postgres / sqlserver / cockroachdb) and **passed every execution**.

Testcase totals are exactly **+6 on every leg** vs `develop`, confirming they compiled and executed everywhere rather than being silently skipped.

### No regressions

The set of failing spec names is **identical** between `develop` and this branch on all 23 legs that produced results:

| | |
|---|---|
| Failures on branch but not `develop` | **none** |
| Failures on `develop` but not branch | none |

Per-leg failure counts are unchanged (adobe2023 3–5, adobe2025 4–6, boxlang 3–4, lucee6 1–2, lucee7 0–1; adobe2023-oracle 11). Those are pre-existing — the leg debt tracked by #3302. The most frequent ones are `HTML-encodes the path in the verb-mismatch message` (17×), `float() applies its default='' / allowNull=true outlier defaults` (11×), `is including partial with query grouped by a column` (11×), and `updateAll with multiple includes updates matching rows` (5×, which is #3294).

### Unrelated observation for #3302

Both runs report **`conclusion: success` while all six engine jobs report `failure`** — `continue-on-error: true` means six red engines roll up green. Additionally the oracle legs on adobe2025 / boxlang / lucee6 emitted **0 testcases** (engine never came up), which `SOFT_FAIL_DBS="oracle"` then absorbs. Two stacked layers of invisibility, exactly what #3302 describes.

bpamiri:
## Adobe 2025 regression, caught and fixed

The `StructKeyExists(request, "wheels")` guard added in `e689e19` errored on all five Adobe 2025 database legs — 8 TenantResolver specs each — with `Element WHEELS is undefined in REQUEST`. Every other engine was green, so local Lucee 7 and PR CI both reported clean.

Cause: `handle()` declares a parameter named `request` (the `MiddlewareInterface` signature mandates it), and on Adobe 2025 the bare `request` token doesn't resolve consistently across expression positions — `StructKeyExists(request, "wheels")` returned true while the `request.wheels` member-access in the guarded `StructDelete` resolved somewhere without the key.

Fixed in `bc0b7b3` by switching to `IsDefined("request.wheels.tenant")`, the same guard the `finally` block in this function already uses — so it's proven on Adobe 2025, not merely plausible. Behaviour is unchanged: deleting an absent key was already a no-op.

### Matrix re-run on `bc0b7b3` vs the `develop` baseline

| | |
|---|---|
| The PR's 10 specs | **280/280 passing** across 28 engine×DB legs |
| Testcase delta vs `develop` | **exactly +10 on every leg** |
| Failures on branch but not `develop` | **none** |

All five Adobe 2025 legs now match `develop` exactly (6/6, 5/5, 4/4, 4/4, 4/4). Remaining per-leg failures are unchanged pre-existing #3302 debt.

- Branch: https://github.com/wheels-dev/wheels/actions/runs/30780625771
- `develop` baseline: https://github.com/wheels-dev/wheels/actions/runs/30773074179

### Documented as invariant 15 (`8d9e20a`)

Added to CLAUDE.md and `.ai/wheels/cross-engine-compatibility.md`, since every middleware component declares a `request` parameter and is exposed to this. Anti-pattern 11 ("never name a parameter after a reserved scope") can't apply there — the interface mandates the name — so the two safe forms are the only options:

```cfm
// safe (a) — one evaluation of the whole dotted path
if (IsDefined("request.wheels.tenant")) { StructDelete(request.wheels, "tenant"); }

// safe (b) — assign before use
if (!StructKeyExists(request, "wheels")) { request.wheels = {}; }
request.wheels.tenant = local.tenant;
```

Also rewrote the note at the top of `TenantResolver.handle()`, which asserted bare `request` "always refers to the built-in request scope, even when a parameter is named `request`" — true on Lucee and Adobe 2023, false on Adobe 2025, and the assumption that produced this bug.

Audited the rest of the framework: no other function declaring a `request` parameter pairs a `StructKeyExists(request, …)` guard with a `request.x` access. The two existing sites (`TenantResolver.handle`, `RequestId.handle`) both use safe form (b).

PR #3635, opened 2026-09-18
fix(model): evaluate compound validation conditions instead of throwing (#3634)

### LINKED ISSUE #3634: 4.1.0 upgrade regression: compound validation `condition` expressions (`&&`/`||`, function calls with arguments) throw `Unsupported operator` and 500 every save
## Summary

In **v4.1.0**, a validation `condition`/`unless` expression that is a **compound boolean** or a **function call with arguments** throws `Wheels.InvalidValidationCondition` ("Unsupported operator in condition: …"). In **v4.0.2** the same expression silently evaluated to `false` and the validation was skipped.

Because the v4.1.0 change intentionally made this path *fail closed*, an app that upgrades from 4.0.2 to 4.1.0 and has any such legacy condition does not merely get a skipped validation back — it gets a **hard 500 on every save** that reaches it. This is an upgrade-safety/breaking-change problem, and 4.1.0 is freshly released.

Reported from a real application (CFWheels 4.1.0 on Lucee 6.2), where `POST /support-issues/create` began returning HTTP 500 on every submission immediately after the 4.0.2 → 4.1.0 upgrade:

```
lucee.runtime.listener.ModernAppListenerException: The `condition` expression
`StructKeyExists(this, 'requestFor') && this.requestFor == 'Engine Part'` could not be evaluated:
Unsupported operator in condition: 'requestfor')
    at model.validations_cfm$cf.udfCall2(/wheels/model/validations.cfm:589)
    at model.validations_cfm$cf.udfCall2(/wheels/model/validations.cfm:484)
    at model.create_cfm$cf.udfCall(/wheels/model/create.cfm:157)
```

## Affects

* Framework: v4.1.0 (verified at `4e7485786`, whose `vendor/wheels/model/validations.cfm` is byte-identical to tag `v4.1.0`).
* Engines: any (the evaluator is plain CFML; reproduced here on Lucee 7.0.x + SQLite via the framework's own specs).
* Not a regression in v4.0.2 behaviour: 4.0.2 had the *same* evaluator limitation, but the `catch` returned `false` (skip validation), so it was invisible. see "What changed" below.

## Minimal reproduction

```cfscript
issue = model("supportIssue").new(summary = "a very long summary that exceeds the limit")
issue.validatesLengthOf(
    property  = "summary",
    maximum   = 5,
    condition = "StructKeyExists(this, 'requestFor') && this.requestFor == 'Engine Part'"
)
issue.save()   // v4.0.2: validation skipped, save succeeds
               // v4.1.0: throws Wheels.InvalidValidationCondition / HTTP 500
```

Verified directly against `$evaluateConditionString` in `vendor/wheels/tests/specs/model/validationsSpec.cfc`:

```cfscript
user.requestFor = "Engine Part"
user.$evaluateConditionString("StructKeyExists(this, 'requestFor') && this.requestFor == 'Engine Part'")
// Error: Unsupported operator in condition: 'requestfor')
```

Other affected shapes (all throw, or silently mis-evaluate, in 4.1.0):

| Expression | 4.1.0 behaviour |
|---|---|
| `1 eq 1 && 2 eq 2` | throws `Unsupported operator in condition: eq` |
| `1 eq 0 \|\| 2 eq 2` | throws `Unsupported operator in condition: eq` |
| `StructKeyExists(this, 'requestFor') && this.requestFor eq 'Engine Part'` | throws `Unsupported operator in condition: 'requestfor')` |
| `StructKeyExists(this, 'username')` (bare) | silently `false` (skips validation) — same class of bug, no throw |

Already-working forms that must not regress: `this.property`, `this.property eq 'v'`, `this.method('arg')`, `1 eq 1`, `5 >= 3`, `isnew()`, `!isnew()`.

## Root cause chain

All references are `vendor/wheels/model/validations.cfm` at `4e7485786`.

1. `$validate()` (line 484) calls `$evaluateCondition(argumentCollection = ...)`.
2. `$evaluateCondition()` (line 581) calls `$evaluateConditionString(...)` inside a `try`; the `catch` (lines 582–590) rethrows as `Wheels.InvalidValidationCondition`. **In v4.0.2 this `catch` did `return false`.** The new comment says the intent: "Fail closed regardless of showErrorInformation. Returning false here means 'skip this validation' … which is fail-open."
3. `$evaluateConditionString()` (line 839):
   * `$normalizeConditionOperators()` (line 864) rewrites `==` → ` eq ` etc. and collapses whitespace. This part is fine for the failing string: it becomes `StructKeyExists(this, 'requestFor') && this.requestFor eq 'Engine Part'`.
   * `$splitConditionOnOperator()` (line 880) then happily finds ` eq ` and returns `{expression: "StructKeyExists(this, 'requestFor') && this.requestFor", operator: "eq", rightOperand: "'Engine Part'"}`.
   * **But that split is only used when `expression` begins with `this.`** (line 844). Here it does not, so the computed split is discarded and control falls through to `$evaluateLogicalExpression(local.normalized)` (line 858) — passing the **whole** compound string, not the left clause.
4. `$evaluateLogicalExpression()` (line 1008) does `ListToArray(condition, " ")` and assumes exactly three meaningful tokens — `tokens[1]` = left operand, `tokens[2]` = operator, `tokens[3]` = right operand (line 1019 passes `LCase(tokens[2])` to `$resolveOperator`). For the production string the split is:
   ```
   [1] StructKeyExists(this,
   [2] 'requestFor')
   [3] &&
   [4] this.requestFor
   [5] eq
   [6] 'Engine
   [7] Part'
   ```
   so `tokens[2]` is `'requestFor')` → lowercased `'requestfor')`.
5. `$resolveOperator()` (line 1039) switches on that token. There is no `&&`/`||` case (and `'requestfor')` is not an operator at all), so it hits `default:` and throws `"Unsupported operator in condition: " & arguments.operator` — **producing exactly the production message**, `Unsupported operator in condition: 'requestfor')`.

So there are two compounding limitations: (i) `$evaluateConditionString` never splits a compound boolean on `&&`/`||`, and (ii) `$evaluateLogicalExpression` is a fixed 3-token parser that reads `tokens[2]` as the operator, so *any* multi-clause or function-call-with-arguments left side hands it garbage. A secondary consequence: a bare function call with arguments (e.g. `StructKeyExists(this, 'x')`) is not recognised as a call at all (only `this.method(...)` and zero-arg `method()` are), so it is parsed by `$evaluateLogicalExpression` and silently returns `false`.

## What changed between 4.0.2 and 4.1.0 (why this is an upgrade-safety issue)

The evaluator never supported compound `&&`/`||` expressions or bare function calls with arguments. What changed is only the failure mode: 4.0.2 swallowed the failure and skipped the validation; 4.1.0 throws. Making a broken condition loud was a deliberate and reasonable goal — silently skipping validations is bad, and nobody is asking to go back to swallowing real errors. The problem is the *blast radius*: an app can carry a legacy compound condition that was harmless-but-ineffective for years, and the upgrade converts every write on that model into a 500 (a total outage of the affected write path) rather than surfacing a diagnosable warning. That deserves at least a documented migration note, and ideally the evaluator should just handle the common forms.

## Candidate directions (not mutually exclusive)

**(a) Teach the evaluator the common forms.** Split the top-level boolean on `||` / `&&` (with `&&` binding tighter), evaluate each side with the existing single-clause machinery, and combine. Also recognise a small, side-effect-free allowlist of function calls with arguments — at minimum `StructKeyExists(this, 'prop')`, which is the idiomatic property-presence check and the exact production case — without introducing `Evaluate()` (the evaluator's doc comment explicitly avoids it). Splitting must be quote- and paren-aware so `this.x eq 'a && b'` is not torn apart, and genuinely unsupported input should still throw so the fail-closed intent survives. This is the most useful outcome, but it is a real change to expression parsing and needs a decent spec table.

**(b) Keep the fail-closed intent but stop 500-ing legacy apps.** Options: log a warning via the `wheels` log (plus a `$wheels`-level deprecation) and preserve the 4.0.2 skip behaviour for conditions the evaluator cannot parse, or keep throwing but ship a documented migration note in the 4.1.x release notes/upgrade guide so affected apps can find and fix their conditions before deploying. The tradeoff is explicit: (b) reintroduces silent skip unless the logging is loud enough to be actionable.

## Non-goal

This is **not** a request to make every unsupported condition silently pass. The reporter agrees that silently skipping validations is wrong. The ask is that a framework upgrade not turn an app's write path into a total outage: either evaluate the common compound/function-call forms correctly, or fail in a way that is diagnosable and documented rather than a 500 on every save.

bpamiri:
## Summary

Fixes #3634.

`$evaluateConditionString()` handed a compound `condition`/`unless` expression (e.g. `StructKeyExists(this, 'requestFor') && this.requestFor eq 'Engine Part'`) to `$evaluateLogicalExpression()`, whose three-token parser reads `tokens[2]` as the operator. For the production string `tokens[2]` is `'requestFor')`, so the evaluator threw `Unsupported operator in condition: 'requestfor')`. Because 4.1.0 made that path fail closed (in 4.0.2 it returned `false`), affected apps moved from a silently skipped validation to an HTTP 500 on every save.

## What changed

`vendor/wheels/model/validations.cfm`:

- `$evaluateConditionString()` now splits a compound boolean at the top level and evaluates each side recursively: `||` first, then `&&`, so `&&` binds tighter than `||`.
- New `$splitTopLevelCondition()` does quote- and paren-aware splitting (reused for argument lists), so `this.status eq 'a && b'` and `StructKeyExists(this, 'a,b')` are not torn apart.
- New `$isConditionFunctionCall()` / `$evaluateFunctionCall()` / `$resolveConditionArgument()` recognise a small whitelist of read-only function calls with arguments — `StructKeyExists`, `IsNull`, `IsNumeric`, `IsSimpleValue`, `IsStruct`, `IsArray`, `IsBoolean`, `IsDate`, `Len` — including the idiomatic `StructKeyExists(this, 'property')`. Arguments are limited to `this`, `this.x` / `this.method()`, quoted literals, booleans and numbers.
- No `Evaluate()` fallback is introduced. An unrecognised function, argument shape, or clause still fails closed and throws `Wheels.InvalidValidationCondition`; the `extendedInfo` hint now lists the widened grammar.

## Tradeoff / what this does not do

This widens the supported grammar; it does **not** re-introduce silent skipping. A legacy condition outside the grammar (an unsupported operator, or a BIF not on the whitelist) still throws after upgrading. That is the deliberate fail-closed behaviour introduced in 4.1.0. If maintainers also want legacy-but-unparseable conditions to stop 500-ing on upgrade (direction (b) in #3634 — a loud `wheels` log plus a documented deprecation/migration note), that is a separate, explicit product decision and is deliberately not made in this PR.

Pre-existing and left out of scope: `$normalizeConditionOperators()` is not quote-aware, so `==`/`<`/`>` inside a quoted literal are still rewritten.

## Tests

- `vendor/wheels/tests/specs/model/validationsSpec.cfc`: 10 new specs — the exact production string, `&&`/`||` combinations, `&&` precedence over `||`, quoted values containing `&&`/`||`, bare and compared whitelisted function calls, the full `validatesLengthOf(condition=...)` → `valid()` path for both true and false conditions, and two fail-closed guards (unsupported compound clause, unrecognised function call).
- Before the fix the new specs reproduced the production failure: `Unsupported operator in condition: 'requestfor')` (992 passed, 3 failed, 3 errors).
- After the fix: model suite **1000 passed**; full core suite **5675 passed, 0 failed** (Lucee 7 + SQLite, `PORT=9090 bash tools/test-local.sh`).
- Verified on Lucee 7 only — no Adobe CF or BoxLang runtime was available in this environment, so the cross-engine behaviour of the new parsing is unverified.

## Changelog

`changelog.d/3634-compound-validation-conditions.fixed.md`.


bpamiri:
Superseded by #3636. Renaming the head branch to the `CONTRIBUTING.md` convention (`fix/3634-compound-validation-conditions`) caused GitHub to close this PR, because GitHub only auto-retargets pull requests whose *base* branch is renamed — a PR's head branch cannot be changed after creation (the REST `PATCH /pulls/{n}` ignores `head`, and GraphQL `UpdatePullRequestInput` has no `headRefName` field). Same branch content, same two commits; please use #3636.

PR #3636, opened 2026-09-19
fix(model): evaluate compound validation conditions instead of throwing (#3634)

### LINKED ISSUE #3634: 4.1.0 upgrade regression: compound validation `condition` expressions (`&&`/`||`, function calls with arguments) throw `Unsupported operator` and 500 every save

bpamiri:
> Supersedes #3635, which GitHub closed automatically when its head branch was renamed from `fix/condition-compound-expressions` to `fix/3634-compound-validation-conditions` (the `CONTRIBUTING.md` convention). Same branch, same two commits — only the head branch name and PR number changed.

## Summary

Fixes #3634.

`$evaluateConditionString()` handed a compound `condition`/`unless` expression (e.g. `StructKeyExists(this, 'requestFor') && this.requestFor eq 'Engine Part'`) to `$evaluateLogicalExpression()`, whose three-token parser reads `tokens[2]` as the operator. For the production string `tokens[2]` is `'requestFor')`, so the evaluator threw `Unsupported operator in condition: 'requestfor')`. Because 4.1.0 made that path fail closed (in 4.0.2 it returned `false`), affected apps moved from a silently skipped validation to an HTTP 500 on every save.

## ⚠️ Remaining upgrade hazard — please read

**This PR changes the failure mode for any function-call `condition`/`unless` that is not on the whitelist: 4.1.0 silently returned `false` (the validation was skipped); this PR throws `Wheels.InvalidValidationCondition`, which is an HTTP 500 on every save.**

That is the same class of upgrade footgun as #3634, and it is a deliberate, narrow consequence of keeping the path fail-closed. It is not hidden: the throw site in `$evaluateFunctionCall()` carries an `UPGRADE HAZARD` comment, and the changelog fragment states it too.

In practice:

- Conditions using the whitelisted read-only predicates/length checks — `StructKeyExists`, `IsNull`, `IsNumeric`, `IsSimpleValue`, `IsStruct`, `IsArray`, `IsBoolean`, `IsDate`, `Len`, `ListFind`, `ListFindNoCase`, `ListContains`, `ListContainsNoCase` — keep working.
- A legacy condition that calls anything else, or uses an operator outside the grammar, still fails closed and 500s after upgrading.

> **Decision (settled):** fail-closed stays. Direction (b) in #3634 — log-and-skip instead of throwing — was considered and **rejected by the maintainer**: silently skipping a validation is precisely how the two conditions that broke this app's 4.1 upgrade survived for years unnoticed (`StructKeyExists(this, 'requestFor') && ...` made three validations permanently inert; `ListFind('24,...', this.countryId)` did the same for two more). A loud failure is the point. This PR therefore widens the grammar and keeps the throw, and the upgrade hazard is called out in three places: this section, an `UPGRADE HAZARD` comment at the throw site, and a dedicated `changed` changelog fragment so it lands in `CHANGELOG.md` as an explicit 4.0.x → 4.1.x upgrade note.

## What changed

`vendor/wheels/model/validations.cfm`:

- `$evaluateConditionString()` now splits a compound boolean at the top level and evaluates each side recursively: `||` first, then `&&`, so `&&` binds tighter than `||`.
- New `$splitTopLevelCondition()` does quote- and paren-aware splitting (reused for argument lists), so `this.status eq 'a && b'` and `StructKeyExists(this, 'a,b')` are not torn apart.
- New `$isConditionFunctionCall()` / `$evaluateFunctionCall()` / `$resolveConditionArgument()` recognise a small whitelist of read-only function calls with arguments — `StructKeyExists`, `IsNull`, `IsNumeric`, `IsSimpleValue`, `IsStruct`, `IsArray`, `IsBoolean`, `IsDate`, `Len`, and the list predicates `ListFind`, `ListFindNoCase`, `ListContains`, `ListContainsNoCase` (two arguments each, result returned as-is, like `Len`) — including the idiomatic `StructKeyExists(this, 'property')`. Arguments are limited to `this`, `this.x` / `this.method()`, quoted literals, booleans and numbers.
- The list predicates are whitelisted because they match `$isConditionFunctionCall()` but were missing from the switch, so a legacy condition of the shape `ListFind('24,32,63,67,117,167,191', this.countryId)` would have gone from 4.1.0's silent `false` to a new 500 — the very regression this PR exists to fix.
- No `Evaluate()` fallback is introduced. An unrecognised function, argument shape, or clause still fails closed and throws `Wheels.InvalidValidationCondition`; the `extendedInfo` hint now lists the widened grammar. The throw site carries the `UPGRADE HAZARD` comment described above.

## Tradeoff / what this does not do

This widens the supported grammar; it does **not** re-introduce silent skipping. A legacy condition outside the grammar (an unsupported operator, or a BIF not on the whitelist) still throws after upgrading. That is the deliberate fail-closed behaviour introduced in 4.1.0. The hazard section above states the tradeoff explicitly; direction (b) in #3634 (log-and-skip) remains an open product decision and is deliberately not made in this PR.

Pre-existing and left out of scope: `$normalizeConditionOperators()` is not quote-aware, so `==`/`<`/`>` inside a quoted literal are still rewritten.

## Tests

- `vendor/wheels/tests/specs/model/validationsSpec.cfc`: 14 specs across the two commits — the exact production string, `&&`/`||` combinations, `&&` precedence over `||`, quoted values containing `&&`/`||`, bare and compared whitelisted function calls, the full `validatesLengthOf(condition=...)` → `valid()` path for both true and false conditions, list-predicate coverage (`ListFind('a,b,c', 'b')` → true, `ListFind('a,b,c', 'z')` → false, plus the `NoCase`/`Contains` variants and a property-valued `ListFind`), and three fail-closed guards (unsupported compound clause, unrecognised function call, and the non-whitelisted `ListSort`).
- The pre-existing "still fails closed for an unrecognised function call" spec still passes.
- Before the list-predicate commit: model suite **1000 passed**; full core suite **5675 passed, 0 failed**.
- After the list-predicate commit: model suite **1004 passed**; full core suite **5679 passed, 0 failed** (Lucee 7 + SQLite, `PORT=9090 bash tools/test-local.sh`).
- Verified on Lucee 7 only — no Adobe CF or BoxLang runtime was available in this environment, so the cross-engine behaviour of the new parsing is unverified.

## Changelog

`changelog.d/3634-compound-validation-conditions.fixed.md`.


bpamiri:
Reviewed the diff (`validations.cfm`, `validationsSpec.cfc`, 2 changelog fragments). The core approach is right and the fail-closed decision is well argued. I did **not** execute the suite — everything below is static analysis against the branch.

## What's good

- Splitting `||` before `&&` gives correct precedence, and `$splitTopLevelCondition` being quote- and paren-aware is the right way to avoid the old three-token parser entirely. The `this.requestFor eq 'a && b || c'` test covers the case that would otherwise silently mis-split.
- `$isConditionFunctionCall` intentionally excluding zero-arg calls correctly defers `isnew()` to `$evaluateBareCall()`.
- Never falling back to `Evaluate()` is preserved; the whitelist switch is the only execution path. Arguments are positional-only, which is a good narrowing.
- `$resolveConditionArgument` throwing on unrecognised tokens rather than guessing is the right default.
- The upgrade hazard is documented in three places — the PR body, the throw-site comment, and a dedicated `changed` fragment. That's exactly how a fail-closed behaviour change should land.

## Blocking

**1. `!` is not accepted on the added grammar, so negation of a whitelisted call still 500s.**

`$isConditionFunctionCall` anchors on `^[A-Za-z_][A-Za-z0-9_]*\s*\(`, so a leading `!` can't match. `!StructKeyExists(this, 'x')` misses the new branch, isn't caught by the bare-call check (which requires the string to end in `()`), and falls to `$evaluateLogicalExpression()`.

That's not hypothetical for this PR's own target: `app/models/CreditRequest.cfc` in the Titan app (the sibling case cited in the PR body) parses to a 2-token expression, and `$evaluateBareCall` already strips every `!` from the string, so bare calls with negation in them are fragile too. The pre-existing `extendedInfo` at `validations.cfm:589` advertises "bare `method()` (optionally negated with `!`)" — the new forms inherit that promise without honouring it.

Suggested fix: strip a single leading `!` in `$evaluateConditionString()` before the function-call check and negate the result, and add coverage for `!StructKeyExists(...)` and `!ListFind(...)`. This is the one I'd want resolved before merge.

## Should fix

**2. A `this.*` argument that doesn't resolve throws, which can flip a compound condition that should short-circuit.**

`$resolveConditionArgument` calls `$resolveThisReference()`, which returns an **empty struct** when the property doesn't exist on the instance, then explicitly throws "Could not resolve". So:

```
StructKeyExists(this, 'optA') || IsNumeric(this.optA)     → throws if optA is absent
StructKeyExists(this, 'a')    || StructKeyExists(this, 'b') → throws if a is absent
```

The second is the idiom the PR exists to support. Property absence is normal in this codebase — the existing tests note Titan properties can legitimately be absent (`_propVal` / TITAN-47). Returning an empty string (or `false`) for an unresolved property would let `||` short-circuit as intended, while still failing closed for genuinely unknown tokens.

## Minor

**3. Operator aliases inside compound clauses.** `$splitConditionOnOperator`'s regex only matches the word forms (`eq|neq|lt|lte|gt|gte`), so `a == b && c != d` depends entirely on `$normalizeConditionOperators` rewriting before the split. Worth a spec asserting the symbolic aliases work inside a compound expression — that's the exact class of gap this PR is cleaning up.

**4. Unrecognised bare identifiers in argument position.** `IsNumeric(myVar)` throws "Unsupported argument". Condition strings aren't an expression language, so failing closed is defensible — but the message could say the identifier form isn't supported, rather than leaving it to be read as a typo.

**5. Arity mismatches report "Unsupported function".** `Len()` with two args breaks out of the `switch` and hits the whitelist throw, implying the function is unsupported when the real problem is arity. A distinct "expects N argument(s)" message would save debugging time.

**6. `Left(local.argument, 5) == "this."` is case-sensitive.** `THIS.x` falls through to the throw. CFC member access is case-insensitive, so this is a small inconsistency — the `$resolveThisReference` path in `$evaluateConditionString` has the same pre-existing behaviour, so it's not a regression.

## Coverage gaps

The new specs cover the happy paths and the two fail-closed paths well. Missing all-negative combinations:

- `!` on a whitelisted call (finding 1)
- an unresolvable `this.*` argument (finding 2)
- either/or shape: `StructKeyExists(this,'a') || StructKeyExists(this,'b')` — `&&` is covered twice, `||` only with literals
- nested/parenthesised grouping, e.g. `(a || b) && c`
- operator aliases inside compound clauses (finding 3)
- arity mismatch (finding 5)

## Net

Fixes the reported `#3634` shape correctly, and the fail-closed rationale is sound — I'd keep it. Findings 3–6 are polish. **Finding 1 should block**: it leaves a 500 in the grammar this PR adds, which is the failure class it set out to eliminate. Finding 2 is close behind, because it breaks the either/or idiom the PR names as its justification.

bpamiri:
## Review follow-up committed

All three findings from review are addressed in `8ac52942e`, with a changelog note in `29de77c0a`. The complexity gate caught one regression from my first cut (see below).

**1. `!` was not accepted on the new grammar** — this was the real 500-in-the-new-grammar gap. `$isConditionFunctionCall()` anchors on `^[A-Za-z_]`, so `!StructKeyExists(this, 'x')` missed the whitelisted branch, failed the bare-call check (needs a trailing `()`), and fell through to the old three-token parser. `$evaluateConditionString()` now strips a leading `!`, evaluates the remainder as one clause, and inverts the finished result — inverting the first operand would be wrong for `!this.x eq 1`. `!=` is already rewritten to ` neq ` by the normaliser, so any `!` reaching that point is negation. Verified: `!StructKeyExists(...)`, `!ListFind(...)`, `!isnew()`, and negated calls inside a compound clause (`1 eq 0 || !ListFind(...)`).

**2. An unresolvable `this.*` argument threw, defeating `||` short-circuit.** `$resolveThisReference()` returns an empty struct for a property not on the instance, and `$resolveConditionArgument()` turned that into a throw — so `StructKeyExists(this, 'a') || StructKeyExists(this, 'b')`, the either/or idiom this PR exists to support, raised a 500. An unresolved property argument now resolves to an empty value (`IsNumeric(this.optIn)` is `false` rather than an exception). Unknown *tokens* still fail closed.

**3. Parenthesised groups were not unwrapped.** `(a || b) && c` left the inner `||` invisible to the top-level splitter. Added `$isWrappedInParens()`, which unwraps only when the parens genuinely wrap the whole clause — `isActive()` and `(this.isActive()` both end in `)` without being groups.

**Minor:** an arity mistake now reports as an arity mistake ("`len()` expects 1 argument(s) but received 2") instead of "unsupported function"; `this`/`this.` matching is case-insensitive; the unsupported-argument message names the supported forms so a bare identifier isn't taken for a typo.

**Complexity gate:** my first cut pushed `$evaluateFunctionCall` to 31 (limit 30) by adding the arity check inside the switch. Extracted to `$assertConditionArity()`, which also puts the arity table in one place. Gate passes.

**Testing note, for transparency:** the test datasource in my local worktree is broken (`c_o_r_e_users` missing from the test DB), so `validationsSpec` cannot run end-to-end locally and I could not execute the new specs there. I verified the grammar against a local Lucee 7 server with a temporary probe spec that calls `$evaluateConditionString()` directly on a `new` model instance (avoiding the DB): 13 cases covering negation — plain, list predicate, compound — absent-property short-circuit, paren groups, `&&`/`||` precedence, symbolic operators in compounds, quoted literals, the bare-call-vs-group guard, and negation of a non-boolean result all pass. `Lucee 7 + SQLite (LuCLI)` is green here, which executes the new specs in a working environment.

**Not changed, deliberately:** the fail-closed decision. I agree with the reasoning in the description — silently skipping is how both inert conditions survived. Findings are fixed *within* that decision, not by relaxing it.

